PR #161, opened 2017-12-15
Don't try to port map to 80 as it is < 1024 in docker docs

binduwavell:
In the docker example we were proposing that folks map port 80 from the docker container to port 80 on localhost. Unless the user is running as root, they likely won't be able to bind to any port < 1024. As such, I am proposing we bind to the local port 8888. Could change it to 3000 to be in-line with the `docker compose` example; but to me the use cases are different so I like having different ports...

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -35,10 +35,10 @@ docker image build -t content-app .
 To run the image locally, you can use the following command:
 
 ```sh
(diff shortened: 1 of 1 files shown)


DenisVuyka:
Thanks, @binduwavell.

PR #167, opened 2017-12-18
[ACA-1065] Add support for enabling multi-select checkboxes

pionnegru:


binduwavell:
I would expect a feature like this to include some sort of documentation?

PR #184, opened 2018-01-30
[ACA-1097] Selecting node on document picker fails

suzanadirla:
fixed bug


DenisVuyka:
@suzanadirla Travis seems to be failing

suzanadirla:
the build is fixed now because I did a rebase, but the order of commits is now wrong

PR #223, opened 2018-03-06
Gbroadbent

gbroadbent:
Uploaded localised files based on EN 2018-03-06


DenisVuyka:
Thanks, we have some merge conflicts, so will update files manually.

PR #231, opened 2018-03-14
[ACA-1213] About - release version

pionnegru:


src/app/components/about/about.component.ts
```diff
@@ -79,6 +79,13 @@ export class AboutComponent implements OnInit {
             ]);
         });
 
+
+        this.http.get('https://api.github.com/repos/Alfresco/alfresco-content-app/releases/latest')
```
DenisVuyka: This won't work in the offline docker container, and will be the same for master/dev branches

PR #232, opened 2018-03-14
[ACA-1211] upgrade to latest ADF alpha

suzanadirla:
To be able to test/move to 'Done' some of the issues dependent on recent ADF changes.


DenisVuyka:
@suzanadirla do we need a new `alfresco-js-api` instance or it remains the same?

PR #241, opened 2018-03-19
[ACA-1226] Node info panel

pionnegru:


src/app/common/directives/node-info.directive.ts
```diff
@@ -0,0 +1,76 @@
...
+    loading: boolean = null;
+
+    @HostListener('document:click', ['$event'])
+    onClick(event) {
+        this.nodeInfo();
+    }
+
+    constructor(private apiService: AlfrescoApiService) {}
+
+    nodeInfo() {
```
DenisVuyka: Can it be renamed to show an "action", like "showNodeInfo"?

PR #253, opened 2018-03-23
[ADF-2559] Sidenav layout extraction

popovicsandras:
Sidenav-layout has been created for reusage in other application like App Dashboard for Process Services (ex-DW) or in the new Business Process Modeller.

This will be moved as it is now into ADF, after the next release of ADF, the adf-layout folder could be removed, and the sidenav-layout component can be used from @alfresco/adf-core.


src/app/components/adf-layout/layout-container.component.ts
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        if (this.sidenavAnimationState === this.SIDENAV_STATES.EXPANDED) {
+            return this.CONTENT_STATES.COMPACT;
+        } else {
+            return this.CONTENT_STATES.EXPANDED;
+        }
+    }
+
+    private get isMobileScreenSize(): boolean {
```
DenisVuyka: this is a private property and HTML template binds to it, will cause error for AoT builds. Needs to be changed to public.

PR #259, opened 2018-03-26
[ACA-1230] Manage versions from right panel tab.

suzanadirla:


src/app/ui/overrides/_adf-info-drawer.scss
```diff
@@ -1,3 +1,16 @@
...
\ No newline at end of file
+}
+
+.adf-manage-versions-empty,
+.adf-manage-versions-no-permission {
+    margin: 24px;
+    color: rgba(0, 0, 0, 0.54);
+    text-align: justify;
+
+    & &-icon {
```
pionnegru: double `& &`

src/app/ui/overrides/_adf-info-drawer.scss
```diff
@@ -1,3 +1,16 @@
...
+        display: block;
```
pionnegru: why do we need to specify `display` property?

src/app/ui/overrides/_adf-info-drawer.scss
```diff
@@ -1,3 +1,16 @@
...
+        font-size: 48px;
+        margin: 0 auto 32px auto;
```
pionnegru: I believe for mat-icon size, these value should be the same
```scss
&-icon {
     width: $icon-size;
     height: $icon-size;
     font-size: $icon-size;
}
```

src/app/ui/overrides/_adf-info-drawer.scss
suzanadirla: It looks strange, I know, but that was intentional. For more clarity, this is now changed :)

PR #297, opened 2018-04-11
[ACA-1275] full view for version comments

suzanadirla:


src/app/ui/overrides/_adf-info-drawer.scss
```diff
@@ -21,3 +21,30 @@ $icon-size: 48px;
...
+    .adf-version-list {
+        height: auto;
+    }
+
+    .mat-list .mat-list-item {
+        &.mat-3-line {
+            display: flex;
+            align-items: center;
+            height: 100%;
+            min-height: 88px;
```
pionnegru: magic number? :)

suzanadirla: yep :) it is what Material uses by default for Three-line list - as I found on this page https://material.io/guidelines/components/lists.html#lists-specs

PR #304, opened 2018-04-13
Disable toolbar buttons instead of hiding

magemello:
I'm proposing to change the toolbar button from hidden to disable when they are not available....this will help the discoverability of functionality. And in general less surprising for users.

**Before**:
No selection:
screen shot 2018-04-13 at 16 18 21
Selected:
screen shot 2018-04-13 at 16 18 28

**Now**:
No selection:
screen shot 2018-04-16 at 09 46 31
Selected:
screen shot 2018-04-13 at 16 17 35

### DIFF
--- a/src/app/components/favorites/favorites.component.html
+++ b/src/app/components/favorites/favorites.component.html
@@ -8,7 +8,7 @@
             <button
                 mat-icon-button
                 color="primary"
-                *ngIf="canPreviewFile(documentList.selection)"
+                [disabled]="!canPreviewFile(documentList.selection)"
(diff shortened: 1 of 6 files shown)


john-knowles:
With the next release coming next week please can we put this on ice, and review the preferred UX with our design team.  Thank you for the submission.

magemello:
Yes no problem, this is just a proposal....let's discuss with UX and then close.

john-knowles:
We have reviewed and after careful consideration have decided not to implement the suggested changes.  The current behaviour conforms to the Material Design standards adopted for this application.

PR #323, opened 2018-04-19
Move all previews out of the LayoutComponent

magemello:
This PR is to fix an issue I had for a demo on IOS tablets. The LayoutComponent, in this case, was causing display the file preview under the LayoutComponent.

screen shot 2018-04-11 at 22 47 40

### DIFF
--- a/src/app/app.routes.ts
+++ b/src/app/app.routes.ts
@@ -50,6 +50,60 @@ export const APP_ROUTES: Routes = [
         }
     },
     {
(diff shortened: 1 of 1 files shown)


DenisVuyka:
Needs to be carefully tested. May break the "go back to folder" functionality when closing the Viewer, especially for shared links and opening file URL without previous navigation history. Those multiple routes were implemented to handle these scenarios.

DenisVuyka:
closing for now as separate routes needed to determine the source of the node to navigate back

PR #847, opened 2018-12-01
[ACA-2062] toolbar updates

john-knowles:
## PR Checklist
Please check if your PR fulfills the following requirements:

```
- [x] The commit message follows our guidelines: https://github.com/Alfresco/alfresco-content-app/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
```

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Application / Infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?
Document List toolbar and context menu actions are unorganised.

Issue Number: ACA-2062, subtask ACA-2062

## What is the new behavior?
Grouped related actions in both the Document List toolbar and context menu, also added separators for each grouping.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

### DIFF
--- a/src/assets/app.extensions.json
+++ b/src/assets/app.extensions.json
@@ -836,7 +836,7 @@
     "viewer": {
       "toolbarActions": [
         {
-          "id": "app.toolbar.download",
+          "id": "app.viewer.download",
           "order": 200,
           "title": "APP.ACTIONS.DOWNLOAD",
           "icon": "get_app",
@@ -872,7 +872,7 @@
           }
         },
         {
-          "id": "app.toolbar.share.edit",
+          "id": "app.viewer.share.edit",
           "order": 401,
           "title": "APP.ACTIONS.SHARE_EDIT",
           "icon": "share",


src/assets/app.extensions.json
```diff
@@ -806,7 +861,7 @@
         },
         {
           "id": "app.viewer.share",
```
adinapitul: should this be app.toolbar.share ?

adinapitul: SHARE_EDIT action has the id: app.toolbar.share.edit

pionnegru: @DenysVuika we need  'id' naming conventions 
@adinapitul what would be easy for testing in terms of cheap locators? having app.[page].toolbar.[action] or app.[page].[action]

DenysVuika: @adinapitul @pionnegru don't forget that `id` values can also be used in testing, we set "element.id" properties to the corresponding id values at runtime

pionnegru: precisely, thats why I've asked whats more convenient for id and testing

adinapitul: from automation perspective, doesn't matter what the id is exactly, but it would be nice for an action from the toolbar to have the same id on all the toolbars

DenysVuika: they have different IDs so that developers can address them separately, i.e. removing buttons from the Viewer toolbar, but leaving in the main page, and vice versa

DenysVuika: In the custom extension, on could disable button like:

```
{
  "id": "app.viewer.share",
  "disabled": true
}
```

adinapitul: then the SHARE_EDIT action from the viewer should follow the same logic and be called app.viewer.share.edit maybe

PR #1179, opened 2019-08-05
Feature ACA-2646 local ids

mmuller88:
Issue Number: ACA-2646

## Changes
Use ./docker-compose-keycloak.yml for IDS


.travis.yml
```diff
@@ -32,16 +32,18 @@ jobs:
...
-    - name: Test Suite sharingContent&markFavorite
-      script: npm run build.e2e && SUITE="--suite sharingContent" npm run e2e:docker
-    - name: Test Suite viewContent&metadata&extensions
-      script: npm run build.e2e && SUITE="--suite viewer,infoDrawer,extensions" npm run e2e:docker
+    # - stage: e2e
+    #   name: Test Suite appNavigation&search
+    #   script: npm run build.e2e && SUITE="--suite authentication,listViews,navigation,application,pagination,search" npm run e2e:docker
+    - name: IDS Test Suite appNavigation&search
+      script: npm run build.e2e && ./start-sso.sh && SUITE="--suite authentication" npm run e2e
+    # - name: Test Suite actionsAvailable
```
DenysVuika: remove commented out code if it is no longer required

docker-compose-keycloak.yml
```diff
@@ -96,21 +155,39 @@ services:
       - 61616:61616 # OpenWire
       - 61613:61613 # STOMP
 
-  content-app:
+  content-app2:
```
adinapitul: don't think this second aca is needed.

.travis.yml
mmuller88: Was only for faster testing. Reverted that

.travis.yml
```diff
@@ -35,6 +35,8 @@ jobs:
     - stage: e2e
       name: Test Suite appNavigation&search
       script: npm run build.e2e && SUITE="--suite authentication,listViews,navigation,application,pagination,search" npm run e2e:docker
+    # - name: IDS Test Suite appNavigation&search
```
DenysVuika: we typically don't commit commented out code, remove it if it is no longer used

docker/proxy/nginx.conf
```diff
@@ -23,6 +23,10 @@ server {
       proxy_pass http://alfresco:8080;
     }
 
+    location /api-explorer/ {
```
DenysVuika: why do we need api explorer for this compose file?

mmuller88: You could use it as an easy way to see how to communicate directly with ACS REST API or even use it for direct communication with using the try functionality from api-explorer. As well api-explorer included in the alfresco/alfresco-content-repository-community image. It cam in handy for AGS as we need to figure out the API call for creating the Share RM side.

DenysVuika:
@mmuller88 the acceptance criteria in the Jira ticket says `Must use Identity Services`. This PR is not using ID but a vanilla keycloak (as was originally introduced). It that expected?

mmuller88:
@DenysVuika great question. It seems we only have AIS in quay.io which would make qauy credentials necessary :-/

PR #1180, opened 2019-08-05
fix chrome version issue

adinapitul:
Issue Number: N/A


package.json
```diff
@@ -15,6 +15,7 @@
     "test": "ng test app --code-coverage",
     "test:ci": "npm run build.shared && npm run build.extensions && ng test adf-office-services-ext --watch=false && ng test app --code-coverage --watch=false",
     "lint": "ng lint && npm run spellcheck && npm run format:check && npm run e2e.typecheck",
+    "prewd:update": "cd ./node_modules/protractor && npm i webdriver-manager@latest",
```
DenysVuika: why Installing extra library inside node modules folder?

adinapitul: the problem with chrome 76 was fixed in webdriver-manager 12.1.6 (https://github.com/angular/webdriver-manager/issues/403), but protractor 5.4.2 does not come with that version by default.
this is a workaround that upgrades the webdriver-manager from inside protractor

DenysVuika: You can add it to the package.json of the main application, the NPM takes those files as top priority. Like `npm i webdriver-manager@latest` but inside the app root folder.

PR #1188, opened 2019-08-19
[ACA-2499] DocumentList - custom name column link does not have a role

pionnegru:
Issue Number: [ACA-2499](https://issues.alfresco.com/jira/browse/ACA-2499)

## Changes
added aria-role and aria-label for link purpose

### DIFF
--- a/src/assets/i18n/en.json
+++ b/src/assets/i18n/en.json
@@ -502,7 +502,8 @@
     },
     "CUSTOM_NAME_COLUMN": {
         "ACCESSIBILITY": {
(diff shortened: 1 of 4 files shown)


src/assets/i18n/en.json
```diff
@@ -499,5 +499,10 @@
                 "TOOLTIP": "Upload new version"
             }
         }
+    },
+    "CUSTOM_NAME_COLUMN": {
+        "ACCESSIBILITY": {
+            "LINK_ARIA_LABEL": "View file {{ fileName }}"
```
suzanadirla: what about folders?

PR #1201, opened 2019-09-11
use Xenial on Travis

adinapitul:
Latest Chrome update generates error when trying to download chrome binaries.

Using Xenial to avoid this.

Issue Number: N/A


DenysVuika:
Please revert this back and use custom chrome instead of custom travis

adinapitul:
Why? We need the latest stable chrome, not a pinned down version that we would need to manually change from time to time.

adinapitul:
In any case, I'll create a separate PR with Pupeeteer to see how that goes.

PR #1207, opened 2019-09-17
[ADF-4824] Add AuthGuard

mmuller88:
Issue Number: ADF-4824

### DIFF
--- a/src/app/app.routes.ts
+++ b/src/app/app.routes.ts
@@ -35,7 +35,7 @@ import {
   AppSharedRuleGuard,
   GenericErrorComponent
 } from '@alfresco/aca-shared';
-import { AuthGuardEcm } from '@alfresco/adf-core';
+import { AuthGuardEcm, AuthGuard } from '@alfresco/adf-core';
 
 export const APP_ROUTES: Routes = [
   {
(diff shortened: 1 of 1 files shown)


DenysVuika:
@mmuller88 I have updated the PR header to match the convention we use

src/app/app.routes.ts
```diff
@@ -73,6 +73,7 @@ export const APP_ROUTES: Routes = [
   {
     path: '',
     component: AppLayoutComponent,
+    canActivate: [AuthGuard],
```
pionnegru: can we keep both guards in the same place?

suzanadirla:
hi.. I see that the AuthGuard does a call to ` this.redirectToUrl('ALL', redirectUrl);` at some point, and ACA has 'ECM' set as the provider inside app.copnfig.json file. I wonder if that causes issues.

suzanadirla:
After more investigating the code: It seems if 'ALL' covers either 'ECM' or 'BPM' or both - so all should be ok.

src/app/app.routes.ts
mauriziovitale: I don't think is needed, the AuthGuard should be smart enough to check if a user is logged in SSO/BPM/ECM

PR #1216, opened 2019-10-02
ACA-2749 server path & ACS on root

mmuller88:
Issue Number: ACA-2749

## Changes
* Allows the definition of a server path like .../content/ for aca. /content/ is the default. For backward compability root / is as well serving aca
* As well using docker compose ACS will be served on root / which is a valid configuration and aca on /content/


README.md
```diff
@@ -93,14 +93,33 @@ Read up on our guidelines for [contributing] and then check out one of our issue
...
+- APP_CONFIG_OAUTH2_HOST: ...
+- APP_CONFIG_OAUTH2_CLIENTID: ...
+- APP_CONFIG_OAUTH2_REDIRECT_LOGIN: ...
+- APP_CONFIG_OAUTH2_REDIRECT_LOGOUT: ...
+- APP_CONFIG_OAUTH2_REDIRECT_SILENT_IFRAME_URI: ...
+
+APP Misc:
+
+- BASE_PATH: ...
+- APP_SERVER_PATH: Server path after host url e.g. /workspace. Defaults to /workspace
```
adinapitul: is this correct? shouldn't this be "content" ?

mmuller88: Yup thanks

e2e/suites/actions/share-file.test.ts
```diff
@@ -48,7 +48,11 @@ describe('Share a file', () => {
   const viewer = new Viewer();
   const page = new BrowsingPage();
   const { dataTable, toolbar } = page;
-  const shareLinkPreUrl = browser.baseUrl + browser.baseUrl.endsWith('/') ? '' : '/' + "#/preview/s/";
+  const shareLinkPreUrl = [
+    browser.baseUrl,
+    browser.baseUrl.endsWith('/') ? '' : '/',
+    '#/preview/s/'
+  ].join('');
```
pionnegru: you can use template string if you don't like concatenation 
```typescript
const baseUrl = `${browser.baseUrl}${browser.baseUrl}${browser.baseUrl.endsWith('/') ? '' : '/'}#/preview/s/`
```

Dockerfile
```diff
@@ -19,7 +19,7 @@ ARG GROUPID=1000
 ARG USERNAME=aca
 ARG USERID=33009
 
-COPY ./docker/nginx.conf /etc/nginx/nginx.conf
+COPY ./docker/nginx.conf /
```
DenysVuika: the path of the config is wrong, it should be placed to the /etc/nginx folder inside the image

docker/entrypoint.sh
```diff
@@ -81,4 +89,4 @@ if [ -n "${APP_BASE_SHARE_URL}" ];then
   cat /tmp/app.config.json > ./app.config.json
 fi
 
-nginx -g "daemon off;"
+nginx -g "daemon off;" -c /nginx.conf
```
DenysVuika: why redirecting the config to the root folder? this will again make the ACA config stand out of all other apps while we are trying to unify everything

DenysVuika:
Let's discuss what changes are really needed for ACA-2749 and what can be left as it is.

Dockerfile
mmuller88: the nginx folder is owned by root. I will try if I can add the alfresco user to the nginx folder

PR #1221, opened 2019-10-11
Change the base href in index.html to make app work natively when deployed to a sub dir on a web server

marsv024:
The Digital Workspace app already have this fixed and have base href point to ./ in the source code of index.html.

Expected behaviour:
* Build the app and deploy it to a subdirectory on your apache server such as http://localhost/content-app/
* Access http://localhost/content-app/ in the web browser.
* The app should work

Observed behaviour:
* Build the app and deploy it to a subdirectory on your apache server such as http://localhost/content-app/
* Access http://localhost/content-app/ in the web browser.
* The app does not work since all resources will be loaded from relative the root of the url, not relative the current directory as expected.

Issue Number: N/A

## Changes
src/index.html

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,7 +11,7 @@
     <meta http-equiv="pragma" content="no-cache" />
 
     <title>Alfresco Content App</title>
-    <base href="/" />
+    <base href="./" />
 
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/png" href="favicon-96x96.png" sizes="96x96" />


src/index.html
```diff
@@ -11,7 +11,7 @@
     <meta http-equiv="pragma" content="no-cache" />
 
     <title>Alfresco Content App</title>
-    <base href="/" />
+    <base href="./" />
```
mauriziovitale: This is not the right way of changing the base ref
Angular provides an option that can be used during the build time
`ng build --prod --base-href .`
So, in my opinion, the package.json build command should be changed.

adinapitul: There is a script for that already in package.json called build.tomcat

DenysVuika: Yeah, this change will break all "normal", non-Tomcat deployments

PR #1225, opened 2019-10-16
[AAE-622] enable noImplicitReturns rule

DenysVuika:
Issue Number: AAE-622

## Changes

- enable `noImplicitReturns` rule
- upgrade to latest JS-API, fix issues

https://issues.alfresco.com/jira/browse/AAE-622

### DIFF
--- a/e2e/components/data-table/data-table.ts
+++ b/e2e/components/data-table/data-table.ts
@@ -125,16 +125,13 @@ export class DataTable extends Component {
     return await this.getSortedColumnHeader().getText();
   }
 
-  async getSortingOrder() {
+  async getSortingOrder(): Promise<string> {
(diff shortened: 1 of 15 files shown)


e2e/components/data-table/data-table.ts
```diff
@@ -125,16 +125,13 @@ export class DataTable extends Component {
...
     if (str.includes('asc')) {
       return 'asc';
     }
-    else {
-      if (str.includes('desc')) {
-        return 'desc';
-      }
-    }
+
+    return 'desc';
```
adinapitul: can we keep the "else" block and return 'none' at the end ?
`if (str.includes('asc')) {
      return 'asc';
    }
    else {
      if (str.includes('desc')) {
        return 'desc';
      }
    }
    return 'none';`

PR #1229, opened 2019-10-17
[ACA-2506] Header - Button does not have a role

mmuller88:
Give Avatar button aria label and role button

Issue Number: ACA-2506

### DIFF
--- a/src/app/components/current-user/current-user.component.html
+++ b/src/app/components/current-user/current-user.component.html
@@ -1,6 +1,11 @@
 <div title="{{ (profile$ | async)?.id }}">
   <div class="current-user__full-name">{{ (profile$ | async)?.userName }}</div>
-  <div class="current-user__avatar am-avatar" [matMenuTriggerFor]="userMenu">
+  <div
+    role="button"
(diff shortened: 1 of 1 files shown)


src/app/components/current-user/current-user.component.html
```diff
@@ -1,6 +1,11 @@
 <div title="{{ (profile$ | async)?.id }}">
   <div class="current-user__full-name">{{ (profile$ | async)?.userName }}</div>
-  <div class="current-user__avatar am-avatar" [matMenuTriggerFor]="userMenu">
+  <div
+    role="button"
+    [attr.aria-label]="(profile$ | async)?.initials | translate"
```
pionnegru: since its marked as a button, it also must be accessible via keyboard like the native button element. this means it needs also the tabindex="0" added to attribute list

pionnegru:
@mmuller88 small comment 👍

src/app/components/current-user/current-user.component.html
mmuller88: Thanks :)

PR #1230, opened 2019-10-18
[ACA-2775] e2e remove some redundant awaits

adinapitul:
- remove some awaits
- add try catch
- small refactoring

Issue Number: ACA-2775

### DIFF
--- a/e2e/components/breadcrumb/breadcrumb.ts
+++ b/e2e/components/breadcrumb/breadcrumb.ts
@@ -45,11 +45,11 @@ export class Breadcrumb extends Component {
   }
 
   async getNthItemName(nth: number) {
-      return await this.getNthItem(nth).getText();
+      return this.getNthItem(nth).getText();
   }
 
   async getItemsCount() {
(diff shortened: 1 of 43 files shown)


e2e/components/breadcrumb/breadcrumb.ts
```diff
@@ -45,11 +45,11 @@ export class Breadcrumb extends Component {
   }
 
   async getNthItemName(nth: number) {
```
DenysVuika: Please add return types to functions in this file

e2e/components/data-table/data-table.ts
```diff
@@ -104,7 +104,7 @@ export class DataTable extends Component {
 
   async getColumnHeadersText() {
```
DenysVuika: missing return types in functions

adinapitul:
@DenysVuika , return types are missing almost everywhere. I would like to address that in a separate PR and separate JIRA task.

DenysVuika:
Makes sense, thanks

PR #1231, opened 2019-10-18
[ACA-2507] Header - link does not have a role

mmuller88:
Add meaning full aria label to avatar button

Issue Number: ACA-2507


DenysVuika:
@mmuller88 the PR needs rebase I think

PR #1239, opened 2019-10-30
[ACA-2782] Header - sidenav menu button expanded/collapsed state not exposed via ARIA

mmuller88:
Issue Number: ACA-2782

## Changes
Accessiblity improvement with showing on the sidenav toggle button if the sidnav is expanded or not

### DIFF
--- a/src/app/components/header/header.component.ts
+++ b/src/app/components/header/header.component.ts
@@ -52,7 +52,7 @@ export class AppHeaderComponent implements OnInit {
   @Output()
   toggleClicked = new EventEmitter();
 
-  @Input() expandedSidenav: Boolean = true;
+  @Input() expandedSidenav = true;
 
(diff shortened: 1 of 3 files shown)


src/app/components/header/header.component.ts
```diff
@@ -51,6 +52,8 @@ export class AppHeaderComponent implements OnInit {
   @Output()
   toggleClicked = new EventEmitter();
 
+  @Input() expandedSidenav: Boolean = true;
```
DenysVuika: Unlike in Java, the TypeScript/JavaScript requires `boolean` ;)

mmuller88: Thanks Denys. VScode suggested to remove  the type declaration. I think that is fine and it can calculate the boolean type from the assignment "true"  :)

PR #1242, opened 2019-11-05
Update e2e Tests to allow ADW builds to run on them as well

mmuller88:
Issue Number: N/A

## Changes
* Make the e2e ECM URL configurable


mmuller88:
Do not Merge is because I would like to get Bamboo and Travis green with the tests :)

PR #1245, opened 2019-11-07
[ACA-2782] Header - sidenav menu button expanded/collapsed state not exposed via ARIA

mmuller88:
Issue Number: ACA-278

## Changes
Sidenav can be collapsed with other events like Search enter. I didn't know that. So the expanded aria label needs to be calculated bases on the correct state.

### DIFF
--- a/src/app/components/layout/app-layout/app-layout.component.html
+++ b/src/app/components/layout/app-layout/app-layout.component.html
@@ -9,13 +9,13 @@
     (expanded)="onExpanded($event)"
   >
     <adf-sidenav-layout-header>
-      <ng-template>
+      <ng-template let-isMenuMinimized="isMenuMinimized">
         <app-header
           role="heading"
           aria-level="1"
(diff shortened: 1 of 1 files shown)


src/app/components/layout/app-layout/app-layout.component.html
```diff
@@ -9,13 +9,13 @@
...
     <adf-sidenav-layout-header>
-      <ng-template>
+      <ng-template let-isMenuMinimized="isMenuMinimized">
         <app-header
           role="heading"
           aria-level="1"
           *ngIf="!hideSidenav"
           (toggleClicked)="layout.toggleMenu($event)"
-          [expandedSidenav]="expandedSidenav"
+          [expandedSidenav]="isMenuMinimized() ? 'false' : 'true'"
```
DenysVuika: I think this can be simplified?
```suggestion
          [expandedSidenav]="!isMenuMinimized()"
```

PR #1246, opened 2019-11-07
[ACA-2601] Search-date-range - fix content overlap

suzanadirla:
Issue Number: https://issues.alfresco.com/jira/browse/ACA-2601

## Changes
Fix the display for mat-form-field legacy style label that overlaps error message.

### DIFF
--- a/src/app/components/search/search-results/search-results.component.scss
+++ b/src/app/components/search/search-results/search-results.component.scss
@@ -73,4 +73,8 @@
       }
     }
   }
+
+  .adf-search-date-range .mat-form-field:not(:first-child) {
+    margin-top: 20px;
+  }
 }


src/app/components/search/search-results/search-results.component.scss
```diff
@@ -73,4 +73,8 @@
       }
     }
   }
+
+  .adf-search-date-range .mat-form-field:not(:first-child) {
+    margin-top: 20px;
```
DenysVuika: should this be fixed in ADF too?

suzanadirla: No, the problem does not happen on demo-shell - it is only on ACA.

DenysVuika: Ok, thanks for confirming

PR #1254, opened 2019-11-19
[ACA-2825] ACA dev and master using one backend

mmuller88:
As preparation for ACA-2825 it would be awesome to allow the image itself to provide an alternative server path.

Issue Number: ACA-2825

## Changes
Introduces an optional SERVER_PATH env which allow to serve aca on a path like .../content-app or in context of the jira task ../dev or .../master . That would reduce the efforts necessary to configure image (copy/paste, release image somewhere) + proxy (adjust haproxy) on rancher. For more information please look into the issue.


mmuller88:
Alternatively we could use the move command instead of the copy command to reduce redundancy but than we couldn't serve the app on root / and server path eg .../content-app

PR #1272, opened 2019-12-02
Hid AOS if record

mmuller88:
Issue Number: N/A

## Changes
Hide the "Edit in Office" action if the node is an Record

### DIFF
--- a/projects/adf-office-services-ext/src/lib/evaluators.spec.ts
+++ b/projects/adf-office-services-ext/src/lib/evaluators.spec.ts
@@ -83,6 +83,58 @@ describe('evaluators', () => {
       expect(canOpenWithOffice(context)).toBeFalsy();
     });
 
+    it('should return [false] if selected file is a record 1', () => {
+      const context: any = {
+        selection: {
+          file: {
+            entry: {
+              name: 'document.docx',
+              isLocked: false,
+              properties: {},
+              aspectName: ['rma:declaredRecord']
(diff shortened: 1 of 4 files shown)


projects/adf-office-services-ext/src/lib/evaluators.ts
```diff
@@ -98,5 +98,14 @@ export function canOpenWithOffice(
     return false;
   }
 
+  // check if record
+  if (
+    file.entry.aspectNames &&
+    (file.entry.aspectNames.indexOf('rma:declaredRecord') !== -1 ||
+      file.entry.aspectNames.indexOf('rma:record') !== -1)
```
pionnegru: array.includes() will return true or false

projects/adf-office-services-ext/src/lib/evaluators.spec.ts
```diff
@@ -83,6 +83,58 @@ describe('evaluators', () => {
...
+          }
+        },
+        permissions: {
+          check: () => true
+        }
+      };
+      expect(canOpenWithOffice(context)).toBeFalsy();
+    });
+
+    it('should return [false] if selected file is a record 1', () => {
```
adinapitul: duplicate test name

mmuller88: jup adjusted slighlty thanks

PR #1273, opened 2019-12-02
Allow other base urls like localhost:8080/workspace

mmuller88:
Issue Number: N/A

## Changes
Allows builds like ADW to specify /workspace for the affected test

### DIFF
--- a/e2e/suites/actions/share-file.test.ts
+++ b/e2e/suites/actions/share-file.test.ts
@@ -48,7 +48,7 @@ describe('Share a file', () => {
   const viewer = new Viewer();
   const page = new BrowsingPage();
   const { dataTable, toolbar } = page;
-  const shareLinkPreUrl = "/#/preview/s/";
+  const shareLinkPreUrl = `${browser.baseUrl}/#/preview/s/`;
 
   const apis = {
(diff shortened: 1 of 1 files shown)


e2e/suites/actions/share-file.test.ts
```diff
@@ -48,7 +48,7 @@ describe('Share a file', () => {
   const viewer = new Viewer();
   const page = new BrowsingPage();
   const { dataTable, toolbar } = page;
-  const shareLinkPreUrl = "/#/preview/s/";
+  const shareLinkPreUrl = `${browser.baseUrl}/#/preview/s/`;
```
adinapitul: this should not be needed.
this shareLinkPreUrl is used in browser.get(...) which is supposed to append it to browser.baseUrl 
is protractor not appending this fragment to browser.baseUrl ?

mmuller88: Yes seems necessary . I tested it locally. With that fix the ADW test passes using 
TEST_BASE_URL=http://localhost:8080/workspace SUITE="--suite sharingContent" npm run e2e 
otherwise I will be directed to the alfresco root page

PR #1285, opened 2019-12-27
[ACA-2849] [MNT] - ADW- Edit option 'Cancel editing'/'edit offline' is not refreshing automatically after uploading new version

mmuller88:
Issue Number: ACA-2849

## Changes
Reload preview if unlock to show "Edit Offline" instead of "Cancel Editing" after uploading new version

### DIFF
--- a/projects/aca-shared/src/lib/services/content-api.service.ts
+++ b/projects/aca-shared/src/lib/services/content-api.service.ts
@@ -287,7 +287,7 @@ export class ContentApiService {
     );
   }
 
-  unlockNode(nodeId: string, opts?: any) {
-    return this.api.nodesApi.unlockNode(nodeId, opts);
+  unlockNode(nodeId: string, opts?: any): Observable<any> {
+    return from(this.api.nodesApi.unlockNode(nodeId, opts));
   }
 }
--- a/src/app/services/content-management.service.ts
+++ b/src/app/services/content-management.service.ts
@@ -40,7 +40,8 @@ import {
   SnackbarInfoAction,
   SnackbarUserAction,
   SnackbarWarningAction,
-  UndoDeleteNodesAction
+  UndoDeleteNodesAction,
+  ViewFileAction
 } from '@alfresco/aca-shared/store';
 import {
   ConfirmDialogComponent,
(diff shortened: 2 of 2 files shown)


projects/aca-shared/src/lib/services/content-api.service.ts
```diff
@@ -287,7 +287,7 @@ export class ContentApiService {
     );
   }
 
-  unlockNode(nodeId: string, opts?: any) {
-    return this.api.nodesApi.unlockNode(nodeId, opts);
+  unlockNode(nodeId: string, opts?: any): Observable<any> {
```
pionnegru: I think it should be Observable <MinimalNodeEntity>

src/app/services/content-management.service.ts
```diff
@@ -1179,12 +1181,21 @@ export class ContentManagementService {
...
-          fileName: node.entry.name
-        })
-      );
-    });
+    this.contentApi.unlockNode(node.entry.id).subscribe(
+      nodeResult => {
+        delete node.entry.properties['cm:lockType'];
+        this.store.dispatch(new SetSelectedNodesAction([node]));
+        if (isPreview) {
+          this.store.dispatch(new ViewFileAction(nodeResult));
```
pionnegru: shouldn't store.select(getAppSelection) fire once you dispatch SetSelectedNodesAction?

src/app/services/content-management.service.ts
```diff
@@ -1179,12 +1181,21 @@ export class ContentManagementService {
...
-    this.contentApi.unlockNode(node.entry.id).catch(() => {
-      this.store.dispatch(
-        new SnackbarErrorAction('APP.MESSAGES.ERRORS.UNLOCK_NODE', {
-          fileName: node.entry.name
-        })
-      );
-    });
+    this.contentApi.unlockNode(node.entry.id).subscribe(
+      nodeResult => {
+        delete node.entry.properties['cm:lockType'];
```
pionnegru: can we fire the reload document-list action or update the Subject?

src/app/services/content-management.service.ts
```diff
@@ -1179,12 +1181,21 @@ export class ContentManagementService {
...
+        this.store.dispatch(new SetSelectedNodesAction([node]));
+        if (isPreview) {
```
pionnegru: don't think this works as expected since it's a function https://github.com/Alfresco/alfresco-content-app/blob/development/projects/aca-shared/rules/src/navigation.rules.ts#L32

PR #1290, opened 2020-01-03
[ACA-2755] SSO: Simplify changing alfresco-js-api

mmuller88:
Issue Number: ACA-2755 

## Changes
Simplify the build / change of an local alfresco-js-api with using git submodules.


mmuller88:
I merged it as I've got some positive feedback from @DenysVuika and @eromano . As well our team would like to try using the submodule feature for js-api changes.

PR #1300, opened 2020-01-10
Git Submodules improvements

mmuller88:
Issue Number: N/A

## Changes
Fix local adf script + attach _LOCAL to the local created libraries


pionnegru:
can we have some description on how the development flow works with this?

PR #1302, opened 2020-01-14
update copyright

pionnegru:
Issue Number: N/A


adinapitul:
I think the year needs to be updated in the license header also.

PR #1306, opened 2020-01-15
chore: a11y removed overriding role from button in menu

downzer0:
Issue Number: ACA-2543

## Changes

* removed the ng attribute which was applying an overriding role to the button, causing it to not read like a button
* added the applied class manually for styling

### DIFF
--- a/src/app/components/toolbar/toolbar-menu-item/toolbar-menu-item.component.html
+++ b/src/app/components/toolbar/toolbar-menu-item/toolbar-menu-item.component.html
@@ -27,9 +27,12 @@
   </ng-container>
 
   <ng-container *ngSwitchDefault>
+    <!--
+      ACA-2543: removed attribute mat-menu-item which applied a role, overriding the native button role; added the applied class directly for styling.
+    -->
     <button
       [id]="actionRef.id"
-      mat-menu-item
+      [class]="'mat-menu-item'"
       color="primary"
       [disabled]="actionRef.disabled"
       [attr.title]="


src/app/components/toolbar/toolbar-menu-item/toolbar-menu-item.component.html
```diff
@@ -27,9 +27,12 @@
   </ng-container>
 
   <ng-container *ngSwitchDefault>
+    <!--
```
DenysVuika: we do not store the change history in the code, please remove the comment

src/app/components/toolbar/toolbar-menu-item/toolbar-menu-item.component.html
```diff
@@ -29,7 +29,7 @@
   <ng-container *ngSwitchDefault>
     <button
       [id]="actionRef.id"
-      mat-menu-item
+      [class]="'mat-menu-item'"
```
pionnegru: are w sure this doesn't impact other things since `mat-menu-item` is a component with some internal  logic such as focus monitor? `mat-menu-item` can take also a role attribute 
```
...
<button
    mat-menu-item
    role="button" // this will override default menuitem role
</div>
...
```

downzer0: this works, thanks. it's not ideal markup-wise, but as you say it has the fewest potential impacts later. the additional role and the tabindex here are not needed, but also don't hurt anything.

PR #1311, opened 2020-01-21
refactor available actions e2e

adinapitul:
Issue Number: N/A

## Changes

refactor available actions e2e so that the preconditions are created just once and then used by multiple test suites


adinapitul:
execution time reduced by 4 mins

PR #1346, opened 2020-02-18
e2e refactorings

adinapitul:
Issue Number: N/A

## Changes

- create a generic-dialog class to remove some duplicated code
- add return types


adinapitul:
all Travis jobs are green, but for some reason the status here still appears as running.

PR #1373, opened 2020-03-24
upgrade to latest ADF alpha

DenysVuika:
Issue Number: N/A

## Changes
upgrade to latest ADF alpha


e2e/suites/actions/share-file.test.ts
```diff
@@ -94,7 +94,7 @@ describe('Share a file', () => {
   })
 
   describe('when logged in', () => {
-    const expiryDate: any = '2020-12-25T18:30:00.000+0000';
+    const expiryDate: any = '2020-12-25T23:59:59.000+0000';
```
adinapitul: please change this back.

the logic of the test is the following:
- a user has a file shared with an expiration date + time by any other tool (a rest API call, a CMIS call, through Share, through Desktop Sync)
- user opens the Shared Link Settings in ACA - to get the URL
- test now checks that the file properties have not been changed in any way simply by opening the dialog

and this test fails, cause as we can see, the original expiration was set to 25/12/2020, 18:30, but after simply opening the dialog, the node properties have been changed, and expiration time is now 23:59.

Even if in ACA setting an expiration with time is not possible, we should not change this value if it has been set through any other tools or clients.

PR #1390, opened 2020-03-28
Use the e2e lite-serve

mauriziovitale:
be able to run e2e against an external domain
be able to run e2e with a custom BE
to create a local env (docker-compose.e2e.yml) doesn’t have share or content app but only BE services


e2e/utilities/repo-client/apis/repo-api.ts
```diff
@@ -47,6 +46,9 @@ export abstract class RepoApi {
 
     protected handleError(message: string, response: any) {
       console.log(`\n--- ${message} error :`);
+      console.log('\t>>> username: ', this.username);
+      console.log('\t>>> password: ', this.password);
```
adinapitul: This would print the password in the travis log. Not sure it's a good idea from a security point of view

scripts/app-config-replace.js
```diff
@@ -0,0 +1,55 @@
...
+const OAUTH_HOST = process.env.OAUTH_HOST || 'keycloak';
+
+const options = {
+  apiHost: {
+    flags: '-a, --api-host',
+    description: "set apiHost's and ecmHost's value with API_HOST",
+    set: appConfig => {
+      appConfig.ecmHost = API_HOST;
+      appConfig.aosHost = API_HOST + '/alfresco/aos';
+      appConfig.baseShareUrl = API_HOST + '/#/preview/s';
```
adinapitul: this should be E2E_HOST

mauriziovitale: nope this is the backend the e2e_host is where the FE is

adinapitul: yes, and the baseShareUrl is the one used for public shared files, and should be the URL of the content-app / digital workspace, not the URL of the backend.

PR #1412, opened 2020-04-15
unified puppeteer setup

DenysVuika:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

- use the same `config.chromeDriver` field as AMA and APA
- use same postinstall approach to setup the webdriver
- pin the puppeteer version to the same as AMA an APA (we can upgrade later)

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "prettier": "^1.17.1",
     "protractor": "5.4.2",
     "protractor-screenshoter-plugin": "0.10.3",
-    "puppeteer": "1.20.0",
+    "puppeteer": "2.1.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -131,5 +131,8 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
+  },
+  "config": {
+    "chromeDriver": "78.0.3904.70"
```
adinapitul, line 136: can you try to use a more recent one? like maybe 80.0.3987.0

DenysVuika: Yes, that is the second step if the initial config works

PR #1478, opened 2020-05-26
[ACA-3315] - add predefined extension guard

SilviuCPopa:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ACA-3315

Added new predefined auth guard

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/projects/aca-shared/src/lib/adf-extensions/extensions-data-loader.guard.spec.ts
+++ b/projects/aca-shared/src/lib/adf-extensions/extensions-data-loader.guard.spec.ts
@@ -131,5 +131,32 @@ describe('ExtensionsDataLoaderGuard', () => {
       expect(erroredSpy).not.toHaveBeenCalled();
       expect(completedSpy).toHaveBeenCalled();
     });
+
+    it('should call canActivate only once', () => {
+      const subject1 = new Subject<true>();
+      const extensionLoaders = {
+        fct1: function() {
+          return subject1.asObservable();
+        }
+      };
+      const extensionLoaderSpy = spyOn(extensionLoaders, 'fct1');
+      const guard = new ExtensionsDataLoaderGuard([extensionLoaders.fct1]);
+
+      subject1.next();
+      expect(guard.involved).toEqual(false);
+      guard.canActivate(route).subscribe(emittedSpy, erroredSpy, completedSpy);
+      expect(guard.involved).toEqual(true);
+      expect(emittedSpy).toHaveBeenCalledWith(true);
+      expect(erroredSpy).not.toHaveBeenCalled();
+      expect(completedSpy).toHaveBeenCalled();
+      expect(extensionLoaderSpy).toHaveBeenCalled();
+
+      extensionLoaderSpy.calls.reset();
+      guard.canActivate(route).subscribe(emittedSpy, erroredSpy, completedSpy);
+      expect(guard.involved).toEqual(true);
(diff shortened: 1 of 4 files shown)


projects/aca-shared/src/lib/adf-extensions/extensions-data-loader.guard.ts
```diff
@@ -75,4 +75,8 @@ export class ExtensionsDataLoaderGuard implements CanActivate {
       })
     );
   }
+
+  canActivateChild(route: ActivatedRouteSnapshot): Observable<boolean> {
```
popovicsandras, line 79: I'm not sure it is needed. Wouldn't this result to load always on every child route? We want to load these loaders only once, not on every child routes.

SilviuCPopa:
> Test missing

I don't think it's needed since `canActivateChild` calls only the `canActivate` function which is already tested

projects/aca-shared/src/lib/adf-extensions/extensions-data-loader.guard.ts
```diff
@@ -45,34 +45,46 @@ export const EXTENSION_DATA_LOADERS = new InjectionToken<
 
 @Injectable({ providedIn: 'root' })
 export class ExtensionsDataLoaderGuard implements CanActivate {
+  involved = false;
```
popovicsandras, line 48: might be better to call it invoked? 🤔

popovicsandras: And it is supposed to be private. Only this class can modify it.

projects/aca-shared/src/lib/adf-extensions/extensions-data-loader.guard.spec.ts
```diff
@@ -131,5 +131,32 @@ describe('ExtensionsDataLoaderGuard', () => {
...
+      const extensionLoaders = {
+        fct1: function() {
+          return subject1.asObservable();
+        }
+      };
+      const extensionLoaderSpy = spyOn(extensionLoaders, 'fct1');
+      const guard = new ExtensionsDataLoaderGuard([extensionLoaders.fct1]);
+
+      subject1.next();
+      expect(guard.involved).toEqual(false);
```
popovicsandras, line 146: This needs to be private. Unit test needs to test the behaviour of the function. This way, you are doing nothing else, but testing the implementation, which is something which we should always avoid.

projects/aca-shared/src/lib/adf-extensions/extensions-data-loader.guard.spec.ts
```diff
@@ -131,5 +131,32 @@ describe('ExtensionsDataLoaderGuard', () => {
...
+      guard.canActivate(route).subscribe(emittedSpy, erroredSpy, completedSpy);
+      expect(guard.involved).toEqual(true);
+      expect(emittedSpy).toHaveBeenCalledWith(true);
+      expect(erroredSpy).not.toHaveBeenCalled();
+      expect(completedSpy).toHaveBeenCalled();
+      expect(extensionLoaderSpy).toHaveBeenCalled();
```
popovicsandras, line 152: It is enough to test if this callback has been called or not. This is the behaviour what we need to test.

PR #1570, opened 2020-07-31
Fix smart runner

eromano:

> - [x] Docs have been added / updated (for bug fixes / features)

> - [ ] Bugfix
> - [ ] Feature
> - [x] Build related changes
> - [ ] Documentation

Fix smart runner

> - [ ] Yes
> - [ ] No

**Other information**:


protractor.conf.js
```diff
@@ -5,14 +5,15 @@ const path = require('path');
...
+const resolve = require('path').resolve;
+
+require('dotenv').config({ path: process.env.ENV_FILE });
 
 const SmartRunner = require('protractor-smartrunner');
 const projectRoot = path.resolve(__dirname);
 const downloadFolder = `${projectRoot}/e2e-downloads`;
 const e2eFolder = path.resolve(projectRoot, 'e2e');
-const E2E_HOST = process.env.E2E_HOST || 'http://localhost';
-const E2E_PORT = process.env.E2E_PORT || 4200;
```
adinapitul, line 15: If E2E_HOST now includes the port as well, maybe call it E2E_URL ?

eromano: This is the name used also in other protractor file is better to keep this name to so for developers is easy to move from  project to another and remember the parameter

adinapitul: Just because it's used in other places doesn't make it "corect", does it?
If port is included, it should be called URL instead of HOST in order to have the name relevant to what the value represents.
I understand it will need changing in the other places as well, for consistency, and in travis variables, and still think it would be a good improvement to do.

eromano: Nope, I will prefer to address this in a separate task this was anyway already with a wrong name before

eromano: Actually is not soo bad as name:
Protocol: http://
Hostname:  contentapp.dev.com
Port: 4200
Host: contentapp.dev.4200
i think we can live with the protocol more in the host

PR #1575, opened 2020-08-03
[ACA-3796] - fixed error on sorting

VitoAlbano:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/src/app/components/files/files.component.html
+++ b/src/app/components/files/files.component.html
@@ -35,7 +35,7 @@
             <ng-template let-col>
               <adf-search-header
                 [col]="col"
-                [currentFolderNodeId]="node?.id"
+                [currentFolderNodeId]="nodeId"
                 [maxItems]="(documentList?.pagination | async)?.maxItems"
                 [skipCount]="(documentList?.pagination | async)?.skipCount"
                 (update)="onFilterUpdate($event)"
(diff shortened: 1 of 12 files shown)


src/app/components/files/files.component.html
```diff
@@ -55,6 +54,7 @@
                   [format]="column.format"
                   [class]="column.class"
                   [sortable]="column.sortable"
+                  [sortingKey]="column.sortingKey"
```
DenysVuika, line 57: can we do fallback to the `key` and not modify the .json files?
```suggestion
                  [sortingKey]="column.sortingKey || column.key"
```

src/app/components/files/files.component.html
```diff
@@ -70,6 +70,7 @@
...
```
DenysVuika, line 73: same as earlier, we can fallback and leave .json files as they are
```suggestion
                  [sortingKey]="column.sortingKey || column.key"
```

src/app/components/files/files.component.html
VitoAlbano: Isn't that value coming from the json file? that is needed as unfortunately the field used in orderBy and the column key sometimes differ (unfortunately), but maybe i'm missing something here :) i'll contact you.

DenysVuika: It's just optimisation, if the "sortingKey" not found, it will be using "key". and if you want a different sorting you just set the value in the "sortingKey". that can prevent modifying all the columns and adding "sortingKey" that is the same as "key"

PR #1602, opened 2020-08-13
[AAE-3278] Temporary fix for some flakyness: C297638 and C297638

popovicsandras:
https://issues.alfresco.com/jira/browse/AAE-3278

### DIFF
--- a/scripts/run
+++ b/scripts/run
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 const path = require('path');
-const tsConfigPath = path.resolve(__dirname, 'tsconfig.base.json');
+const tsConfigPath = path.resolve(__dirname, 'tsconfig.json');
 const tsConfig = require(tsConfigPath);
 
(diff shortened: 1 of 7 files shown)


scripts/run
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 const path = require('path');
-const tsConfigPath = path.resolve(__dirname, 'tsconfig.base.json');
+const tsConfigPath = path.resolve(__dirname, 'tsconfig.json');
```
DenysVuika, line 4: you should not be using tsconfig.json for now until Angular 10.1.0 (where they rollback the solution style configs). Right now tsconfig is for IDEs and tooling, as it contains all references.

popovicsandras: This tsconfig is only used for the ts-node compiler, it is not used anywhere else. The 'npm run lite-serve' script was errored because of it. I've just restored it, now it works.

PR #1669, opened 2020-09-11
ACA Run the test against the remote env

mauriziovitale:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:


scripts/app-config-replace.js
```diff
@@ -21,6 +21,7 @@ const options = {
       appConfig.apiHost = API_HOST;
```
adinapitul, line 21: would have to be `appConfig.apiHost = API_HOST + '/alfresco/aos'` to actually work, but it's not used by any tests at the moment afaik

scripts/app-config-replace.js
```diff
@@ -21,6 +21,7 @@ const options = {
...
       appConfig.bpmHost = API_PROCESS_HOST;
       appConfig.ecmHost = API_CONTENT_HOST;
+      appConfig.baseShareUrl = API_CONTENT_HOST + '/#/preview/s';
       appConfig.aosHost = API_AOS_HOST;
```
adinapitul, line 25: would have to be `appConfig.aosHost = API_AOS_HOST + '/alfresco/aos'` to actually work, but it's not used by any tests at the moment afaik

popovicsandras: I reckon aosHost should contain everything, `url` and `path` as well, this is why we moved this data into env vars. I believe, a code file should contain only behaviour and not data. For data, we can simply put them into environment variables.

popovicsandras (approved):

scripts/app-config-replace.js
adinapitul: I added this comment after I saw [this](https://github.com/Alfresco/alfresco-content-app/pull/1669/files#diff-354f30a63fb0907d4ad57269548329e3R46). This makes it necessary to add the path in here, as far as I can understand.

popovicsandras: Sorry, I'm not sure I can follow you. Why is it necessary to add this path in this codefile and not as part of an environment variable?

adinapitul: this should be set by an environment variable, as you say. But at the moment, in travis.yml this value is `- API_AOS_HOST=$API_CONTENT_HOST_REMOTE`
So, considering the env variable used on travis, then the path needs to be added  like a mentioned above. It's either this or correctly set the variable on travis.yml 😃

.travis.yml
```diff
@@ -21,8 +20,6 @@ branches:
     - /.*beta.*/
 
 before_install:
-  - sudo /etc/init.d/postgresql stop
-  - npm install -g npm@latest
   - sleep 3
```
DenysVuika, line 23: sleep not needed here as previous 2 lines were removed

e2e/suites/actions-available/files-folders/files-folders-actions.test.ts
```diff
@@ -63,11 +63,15 @@ describe('Files / folders actions : ', () => {
 
   const loginPage = new LoginPage();
 
+  let initialSearchTotalItems: number;
```
DenysVuika, line 66: use as a local const inside beforeAll

e2e/suites/actions-available/files-folders/files-folders-actions.test.ts
```diff
@@ -63,11 +63,15 @@ describe('Files / folders actions : ', () => {
...
+
   beforeAll(async () => {
     await adminApiActions.createUser({ username });
 
     parentId = (await userApi.nodes.createFolder(parent)).entry.id;
 
+    initialSearchTotalItems = await userApi.search.getTotalItems(username);
```
DenysVuika, line 73: not used outside the scope, so should be declared locally
```suggestion
    const initialSearchTotalItems = await userApi.search.getTotalItems(username);
```

e2e/suites/actions-available/libraries/library.test.ts
```diff
@@ -268,7 +268,7 @@ describe('Library actions : ', () => {
       await testUtil.checkContextMenu(testData.privateUserMemberFav.name, testData.privateUserMemberFav.contextMenu);
     });
 
-    it('[C290086] Moderated library, user is a member, favorite', async () => {
+    it('[C290086] Moderated library, user is a member, favorite ', async () => {
```
DenysVuika, line 271: useless space
```suggestion
    it('[C290086] Moderated library, user is a member, favorite', async () => {
```

DenysVuika (approved):

projects/aca-testing-shared/src/utilities/repo-client/apis/favorites/favorites-api.ts
```diff
@@ -130,7 +140,7 @@ export class FavoritesApi extends RepoApi {
       };
       return await Utils.retryCall(favorite);
     } catch (error) {
-      // this.handleError(`${this.constructor.name} ${this.isFavoriteWithRetry.name}`, error);
+      // this.handleError(`FavoritesApi isFavoriteWithRetry : catch : `, error);
```
DenysVuika, line 143: should this be commented?

projects/aca-testing-shared/src/utilities/repo-client/apis/shared-links/shared-links-api.ts
```diff
@@ -56,17 +56,17 @@ export class SharedLinksApi extends RepoApi {
         return this.shareFileById(current);
       }, Promise.resolve());
     } catch (error) {
-      this.handleError(`${this.constructor.name} ${this.shareFilesByIds.name}`, error);
+      this.handleError(`SharedLinksApi shareFilesByIds : catch : `, error);
     }
   }
 
   async getSharedIdOfNode(name: string) {
```
DenysVuika, line 63: missing return type
```suggestion
  async getSharedIdOfNode(name: string): Promise<string> {
```

projects/aca-testing-shared/src/utilities/repo-client/apis/sites/sites-api.ts
```diff
@@ -59,17 +59,27 @@ export class SitesApi extends RepoApi {
...
+    try {
+      await this.apiAuth();
+      return (await this.sitesApi.listSiteMembershipsForPerson(this.getUsername())).list.pagination.totalItems;
+    } catch (error) {
+      this.handleError(`SitesApi getSitesTotalItems : catch : `, error);
+      return -1;
+    }
+  }
+
   async getDocLibId(siteId: string) {
```
DenysVuika, line 77: missing return type
```suggestion
  async getDocLibId(siteId: string): Promise<string> {
```

DenysVuika (changes requested):

projects/aca-testing-shared/src/utilities/repo-client/apis/favorites/favorites-api.ts
adinapitul: Yes. During a good and clean run, this fails if the node is already a favorite. For the purpose of the test, all I care is for the items to be favorite, so these logs would only pollute

mauriziovitale: it would be better to use the Logger with a LOG_LEVEL debug in this case it would be visible only based on the config. Btw if we don’t need it just remove it

DenysVuika (approved):

PR #1868, opened 2020-12-08
disable about and settings by default

DenysVuika:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [x] Other... Please describe:

- disable About and Settings by default

> - [ ] Yes
> - [x] No

**Other information**:


eromano (approved):

src/assets/app.extensions.json
```diff
@@ -8,6 +8,7 @@
   "$runtime": "1.7.0",
   "$description": "Core application extensions and features",
   "$references": [],
+  "$ignoreReferenceList": ["about.plugin.json", "settings.plugin.json"],
```
mteodori, line 11: can we use the APP_EXTENSIONS_IGNORE_REFS variable directly here so we don't need the extra shell script? and add a replacer in the startup for the default value

docker/default.conf.template
```diff
@@ -23,6 +23,10 @@ server {
         alias  ${NGINX_ENVSUBST_OUTPUT_DIR}/app.config.json;
     }
 
+    location ~ /assets/app.extensions.json {
```
mteodori, line 26: did you test with the run.sh script the mapping is as you expect?

src/assets/app.extensions.json
DenysVuika: not everything runs via docker, using variable there will break the whole development process

docker/default.conf.template
DenysVuika: still having issues with running the scripts, it seems they need some fixes to be used dev environemnts

DenysVuika: no, the `run.sh` and `run-content-ce.sh` not working for me

src/assets/app.extensions.json
mteodori: the replacer like you have in the other projects supports plain javascript runs maybe @popovicsandras can explain better

mauriziovitale (approved):

PR #2047, opened 2021-03-10
[ACA-4287] Fix Reload library

dhrn:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACA-4287

fixed

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/src/app/services/content-management.service.ts
+++ b/src/app/services/content-management.service.ts
@@ -34,6 +34,7 @@ import {
   NavigateToParentFolder,
   NodeInfo,
   ReloadDocumentListAction,
+  ReloadLibraryAction
   SetSelectedNodesAction,
   SnackbarAction,
   SnackbarErrorAction,
--- a/src/app/services/content-management.service.ts
+++ b/src/app/services/content-management.service.ts
@@ -325,6 +325,7 @@ export class ContentManagementService {
       () => {
         this.libraryDeleted.next(id);
         this.store.dispatch(new SnackbarInfoAction('APP.MESSAGES.INFO.LIBRARY_DELETED'));
+        this.store.dispatch(new ReloadLibraryAction());
       },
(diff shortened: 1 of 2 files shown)


adinapitul:
this problem does NOT happen in ACA
this problem does NOT happen in ADW in My Libraries or Favorite Libraries
it only happens in ADW in All Libraries - which is only present in ADW.
so, why should a change in ACA be required?

dhrn:
ADW is depends on the underlying event hook from ACA. we don't have any other approach other than modifying the current ACA implementation. you can correct me if i am wrong. I will discuss with @DenysVuika to avoid this problems in future and  give more capability to extension users.

pionnegru:
and [these](https://github.com/Alfresco/alfresco-content-app/blob/develop/src/app/components/libraries/libraries.component.ts#L57-L59 ) don't do the same thing?
maybe we ned to retire them and use thee actions.

dhrn:
No this didn't help. we have own layout.

pionnegru:
I'm saying maybe we should remove [these](https://github.com/Alfresco/alfresco-content-app/blob/develop/src/app/components/libraries/libraries.component.ts#L57-L59) and use ReloadLibraryAction also

dhrn:
yes  you are correct. also planning to move the hooks to aca shared library like [this](https://github.com/Alfresco/alfresco-content-app/compare/dev-dhrn-move-hooks)

dhrn:
We can do that as well

dhrn:
Hmm, that is good approach. we can do that. Wondering how will impact the other community extension. so eventually we will remove all the subject based hook?

dhrn:
addressed in #2052

PR #2075, opened 2021-03-31
[AAE-4842] app config extensibility support

DenysVuika:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

- provide support for extending app config data from within the extension files

> - [ ] Yes
> - [x] No

**Other information**:
AAE-4842

### DIFF
--- a/projects/aca-shared/src/lib/services/app.extension.service.ts
+++ b/projects/aca-shared/src/lib/services/app.extension.service.ts
@@ -136,6 +136,7 @@ export class AppExtensionService implements RuleContext {
       return;
     }
 
+    this.loadConfiguration(config);
     this.settingGroups = this.loader.getElements<SettingsGroupRef>(config, 'settings');
 
     this.headerActions = this.loader.getContentActions(config, 'features.header');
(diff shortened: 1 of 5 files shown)


projects/aca-shared/src/lib/services/app.extension.service.ts
```diff
@@ -268,6 +269,16 @@ export class AppExtensionService implements RuleContext {
       });
   }
 
+  loadConfiguration(config: ExtensionConfig) {
+    if (config) {
+      const configuration = config['configuration'];
+
+      if (configuration) {
+        this.appConfig.config = mergeObjects(this.appConfig.config, configuration);
```
mauriziovitale, line 277: what is going to happen if different extensions contain the same properties? the last one wins? can we define the merge strategy somewhere?

mauriziovitale: something like 
`configuration: { 
   strategy: 'merge' | 'append',
   value: {}
}`

DenysVuika: the order is the same as the order for merging all other extension data - the order of file declaration, we don't need any merge strategy here because we already have it for the rest of the cases

PR #2094, opened 2021-04-27
Refactor env variables

mauriziovitale:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:


protractor.conf.js
```diff
@@ -21,13 +21,13 @@ const width = 1366;
 const height = 768;
 
 const SAVE_SCREENSHOT = process.env.SAVE_SCREENSHOT === 'true';
-const API_CONTENT_HOST = process.env.API_CONTENT_HOST || 'http://localhost:8080';
+const APP_CONFIG_ECM_HOST = process.env.APP_CONFIG_ECM_HOST || 'http://localhost:8080';
```
adinapitul, line 24: Shouldn't this be just API_HOST like it is on adw and apps?

protractor.conf.js
```diff
@@ -21,13 +21,13 @@ const width = 1366;
...
 const MAXINSTANCES = process.env.MAXINSTANCES || 1;
 const MAX_RETRIES = process.env.MAX_RETRIES || 1;
 const LOG_LEVEL = process.env.LOG_LEVEL || 'ERROR';
 
 const appConfig = {
-  hostEcm: API_CONTENT_HOST,
+  hostEcm: APP_CONFIG_ECM_HOST,
```
adinapitul, line 30: Same as before, should be API_HOST

protractor.conf.js
mauriziovitale: We are still discussing that. For now, there are different variables for ecm and bpm.

mauriziovitale: @popovicsandras am I correct ?

popovicsandras: @adinapitul We are changing and standardising these variables in every repository at the moment. What you can see here is going to be the new name of variables, which is going to be accross every repo, and with this we even get rid of the variable name differences between local dev and deployement time. (We are going to use the same var names as what the docker container needs)

popovicsandras (approved):

PR #2382, opened 2021-12-16
MNT-22687- Add new aosPlugin variable to hide or show the AOS plugin

antoniojfelix:

> - [X] Docs have been added / updated (for bug fixes / features)

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/MNT-22687

There's a new variable named "aosPlugin" in the **app.config.json** file where we can disable or enable the AOS plugin.

> - [ ] Yes
> - [X] No

**Other information**:


.travis.yml
```diff
@@ -49,6 +49,7 @@ env:
     - APP_CONFIG_AUTH_TYPE=BASIC
     - APP_CONFIG_OAUTH2_HOST=http://localhost:4200/auth/realms/alfresco
     - APP_CONFIG_OAUTH2_CLIENTID=alfresco
+    - APP_CONFIG_PLUGIN_AOS=true
```
mauriziovitale, line 52: are we sure that the default value should be true ? ( just asking )

package.json
```diff
@@ -7,7 +7,7 @@
...
+    "assemble-app-config": "envsub --env-file ./.env --env APP_CONFIG_ECM_HOST={protocol}//{hostname}{:port} --env APP_CONFIG_PROVIDER=ECM --env APP_CONFIG_AUTH_TYPE=BASIC --env APP_CONFIG_PLUGIN_AOS=true --env APP_CONFIG_OAUTH2_HOST=http://localhost:4200/auth/realms/alfresco --env APP_CONFIG_OAUTH2_CLIENTID=alfresco --env APP_CONFIG_OAUTH2_IMPLICIT_FLOW=true --env APP_CONFIG_OAUTH2_SILENT_LOGIN=true --env APP_CONFIG_OAUTH2_REDIRECT_SILENT_IFRAME_URI={protocol}//{hostname}{:port}/assets/silent-refresh.html --env APP_CONFIG_OAUTH2_REDIRECT_LOGIN=/ --env APP_CONFIG_OAUTH2_REDIRECT_LOGOUT=/logout --all ./app/src/app.config.json.tpl ./app/.tmp/app.config.json",
```
mauriziovitale, line 10: this is not needed, if you are testing it from local it should be in your .env
As part of the CI/CD the value is already part of the .travis file

projects/adf-office-services-ext/README.md
```diff
@@ -70,3 +70,9 @@ Update `app.extensions.json` and append a reference to the plugin definition:
   "$references": ["aos.plugin.json"]
 }
 ```
+
+## Disable and Enable the extension after it is installed
+
+There's an environment that can disable or enable the installed extension.
```
mauriziovitale, line 76: Could we specify the default value is true, please?

projects/adf-office-services-ext/src/lib/evaluators.spec.ts
```diff
@@ -268,6 +283,7 @@ describe('evaluators', () => {
 
     it('should return [false] if permissions check is false', () => {
       const context: any = {
+        aosPlugin: true,
```
mauriziovitale, line 286: do we really need to add it everywhere?

projects/adf-office-services-ext/src/lib/evaluators.spec.ts
```diff
@@ -220,6 +233,7 @@ describe('evaluators', () => {
 
     it('should return [false] if current user is not lock owner', () => {
       const context: any = {
+        aosPlugin: true,
```
mauriziovitale, line 236: do we really need to add it everywhere?

projects/adf-office-services-ext/src/lib/evaluators.spec.ts
```diff
@@ -303,5 +320,12 @@ describe('evaluators', () => {
 
       expect(canOpenWithOffice(context)).toBeTruthy();
     });
+
+    it('should return [false] when `aosPlugin` property is false', () => {
```
mauriziovitale, line 324: can we add a new test that checks only the asoPlugin in case of true?

app/src/app.config.json.tpl
```diff
@@ -6,6 +6,7 @@
   "providers": "${APP_CONFIG_PROVIDER}",
   "authType": "${APP_CONFIG_AUTH_TYPE}",
   "loginRoute": "login",
+  "aosPlugin" : ${APP_CONFIG_PLUGIN_AOS},
```
mauriziovitale, line 9: the position is completely wrong. There is a specific section that contains all the plugins

mauriziovitale: i.e. 

"plugins": {
    "microsoftOnline": "${APP_CONFIG_PLUGIN_MICROSOFT_ONLINE}",
    "processAutomation": true,
    "aiService": ${APP_CONFIG_PLUGIN_AI_SERVICE}
  },

mauriziovitale (changes requested):
Please make sure we use the same strategy used across plugins 
i.e. create a rule isMicrosoftOnlinePluginEnabled

export function isMicrosoftOnlinePluginEnabled(): boolean {
    return localStorage && localStorage.getItem('microsoftOnline') === 'true';
}

or 

export function isProcessServiceCloudPluginEnabled(): boolean {
    return localStorage && localStorage.getItem('processAutomation') === 'true';
}

.travis.yml
antoniojfelix: Yes, by default the AOS plugin is already enabled. The goal of this issue is to add the possibility of disabling it.

app/src/app.config.json.tpl
antoniojfelix: I couldn't find that section in the `app.config.json.tpl` file but I agree, I will add that section for this **aosPlugin** variable.

package.json
antoniojfelix: Ok thanks. I'll revert this file.

projects/adf-office-services-ext/README.md
antoniojfelix: Yes, of course. Thanks for pointing that out.

projects/adf-office-services-ext/src/lib/evaluators.spec.ts
antoniojfelix: Yes, I think it makes sense in the context of each test. With the introduction of this new boolean, the remain tests will only make sense if the **aosPlugin** is true. If the **aosPlugin** is false, that means we won't have an AOS plugin to test all the other things.

projects/adf-office-services-ext/src/lib/evaluators.spec.ts
antoniojfelix: It would be complicated. Since I added the validation of the AOS plugin at the beggining of the `canOpenWithOffice` method, I can't really return true unless I have all of the other validations to be true as well. And we are testing that already in the 'should return [true] if all checks succeed' test.

antoniojfelix:
Hi Maurizio,
I believe that for this case it would be a little bit different, since this is regarding ACA and there aren't the possibility of enabling/disabling other plugins by changing environment properties. But once this PR is merged I will open another one regarding ADW, where it would be possible to disable/enable this ACA plugin througt the ADW.

projects/aca-shared/src/lib/services/app.extension.service.ts
```diff
@@ -106,6 +106,7 @@ export class AppExtensionService implements RuleContext {
   profile: ProfileState;
   repository: RepositoryInfo;
   withCredentials: boolean;
+  aosPlugin: boolean;
```
mauriziovitale, line 109: not sure if this is the best place where to add the plugin logic. Usually, every plugin has its own service i.e. MicrosoftExtensionService or ProcessServiceExtensionService

mauriziovitale: An example [here](https://github.com/Alfresco/alfresco-apps/blob/66e0fe66adeeff9f8e6f80a15cc6af7d0f5b4158/libs/content-ee/microsoft-office-online-integration-extension/src/lib/services/microsoft-extension.service.ts#L70)

projects/adf-office-services-ext/src/lib/evaluators.spec.ts
```diff
@@ -29,6 +29,7 @@ describe('evaluators', () => {
   describe('canOpenWithOffice', () => {
     it('should return [false] if using SSO', () => {
       const context: any = {
+        aosPlugin: true,
```
mauriziovitale, line 32: it should not be necessary to add the true value, the plugin should be enabled is the plugin. aosPlugin is true or the property is missing to be backward compatible

projects/adf-office-services-ext/src/lib/evaluators.spec.ts
mauriziovitale: it should be working fine thanks to the default value the following methods return when the property is not found
this.appConfig.get<boolean>('plugins.aosPlugin', true);

projects/adf-office-services-ext/src/lib/evaluators.ts
```diff
@@ -26,7 +26,15 @@
 import { RuleContext } from '@alfresco/adf-extensions';
 import { getFileExtension, supportedExtensions } from './utils';
 
-export function canOpenWithOffice(context: RuleContext): boolean {
+export interface AcaRuleContext extends RuleContext {
```
mauriziovitale, line 29: Same here we are mixing the general acaRules with aosPlugin

mauriziovitale: @DenysVuika can you check if this is correct?

projects/adf-office-services-ext/src/lib/evaluators.spec.ts
DenysVuika: yeah, maybe worth having it in the "beforeEach" section

projects/adf-office-services-ext/src/lib/evaluators.ts
DenysVuika: looks okay, we used this kind of interfaces before just to use Typescript type checking better for the evaluators/rules... in the current case in ACA the App Extensions Service is the RuleContext at the same time, so in order for a rule to access it's properties we use interfaces... without the interface, it would probably be `any`

```
export function canOpenWithOffice(context: any): boolean {
  if (!context.aosPlugin) {
    return false;
  }
```

docker/docker-entrypoint.d/30-sed-on-appconfig.sh
```diff
@@ -93,6 +93,15 @@ echo "replace APP_CONFIG_ECM_HOST"
     -i "${NGINX_ENVSUBST_OUTPUT_DIR}/app.config.json"
 fi
 
+if [[ -n "${APP_CONFIG_PLUGIN_AOS}" ]]; then
```
mauriziovitale, line 96: this code should not be any more needed since we are using subenv on ACA
When the docker image is build Nginx should replace the values as part of the tpl with the env variable values

mauriziovitale (changes requested):
Please just remove the replacement.

mauriziovitale (approved):

PR #2569, opened 2022-07-19
Bug fix. Don't lose the contentService property after page refresh.

joerison:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

With each page refresh, the `contentService` property is reset to the environment variable setting. This makes setting this property through the interface meaningless. I'm refer to the configuration screen currently accessed via the `#/settings` route and checkbox "Enable Content Services Extensions".

The configuration preference will be the one in the browser's storage. If not exists, it will follow the flow that was already exists and create the respective property.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/app/src/app/services/content-service-extension.service.ts
+++ b/app/src/app/services/content-service-extension.service.ts
@@ -15,7 +15,17 @@ import { take } from 'rxjs/operators';
 })
 export class ContentServiceExtensionService {
   constructor(private appConfigService: AppConfigService) {
-    this.updateContentServiceAvailability();
+    this.prepareContentServiceProperty();
+  }
+
+  private prepareContentServiceProperty() {
+    if (!this.existsContentServicePropertyLocalStorage()) {
+      this.updateContentServiceAvailability();
+    }
+  }
+
+  private existsContentServicePropertyLocalStorage() {
+    return localStorage && localStorage.getItem('contentService');
   }
 
   updateContentServiceAvailability() {


app/src/app/services/content-service-extension.service.ts
```diff
@@ -15,7 +15,17 @@ import { take } from 'rxjs/operators';
 })
 export class ContentServiceExtensionService {
   constructor(private appConfigService: AppConfigService) {
-    this.updateContentServiceAvailability();
+    this.prepareContentServiceProperty();
+  }
+
+  private prepareContentServiceProperty() {
+    if (!this.existsContentServicePropertyLocalStorage()) {
```
DenysVuika, line 22: hmm, I think that this call prevents checking the app config service in case the flag was previously stored in the local storage. that could mean that if admin changes this setting on the backend, the app will still be using localstorage....

joerison: This is strange, because there are other options not have this behavior (don't lose the localstorage property after page refresh):
- Enable AI Extensions
- Enable Process Services Extensions

The only ones with this "bug" - lose/reset this localstorage property after page refresh:
- Enable Content Services Extensions
- Enable Microsoft Office Online Extension (I noticed that it is like that too)

DenysVuika:
closing for now, please reopen with unit tests if it is still valid

PR #2577, opened 2022-08-05
[ACS-3338] Manage Rules screen / listing - Folder rules service

nikita-web-ua:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

There is no way to fetch the rules in ACA yet.

To interact with the API, an Angular service was created. The service fetch the data from the API, and format it using the types defined in `src/lib/model` in the `aca-folder-rules` project. 

> - [ ] Yes
> - [x] No

**Other information**:


projects/aca-folder-rules/src/lib/services/folder-rules.service.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  it('should make 1 API call', async () => {
+    expect(apiCallSpy).toHaveBeenCalledTimes(1);
+  });
+
+  it('should format response to list of rules', async () => {
+    expect(rules).toEqual(dummyRules, 'The list of rules is incorrectly formatted');
+  });
+
+  it('should format each rule entity', async () => {
```
TomikaArome, line 174: This test is unnecessary as it's already covered by the previous one. If `rules` is different to `dummyRules` in any way, including nested objects, the statement on line 171 should fail.

projects/aca-folder-rules/src/lib/services/folder-rules.service.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+    const unformattedRule = dummyResponse.list.entries[0].entry;
+
+    expect(rules[0]).not.toEqual(unformattedRule as any, 'Rule entity is not formatted');
+    expect(rules[0]).toEqual(dummyRules[0], 'Rule entity is not formatted correctly');
+  });
+
+  it('should keep the order of the rules list', async () => {
```
TomikaArome, line 181: This test is also unnecessary for the same reason as the last. Just as a precaution though, could you try making these tests fail and see which ones fail each time.

TomikaArome (changes requested):
Good job with your first PR Mykyta 🎉 There's some good decisions in there.

A few extra comments:
- We should preferably not be changing the `package-lock.json` file. This happens when you run `npm install`. You can run `npm ci` instead to avoid this file being changed (see https://docs.npmjs.com/cli/v8/commands/npm-ci). Reset this file to get its original contents back and then run `npm ci` again.
- I like the simple, yet effective solution you came up with for mocking the `callApi` method. Generally I try not to spy on any of the methods of the service/component I'm testing but I think this is far easier.
- It might be simpler to combine all of your unit tests into one (with several different `expect` statements). Your current approach is fine though so this really is a non-issue if you prefer to keep it this way.

projects/aca-folder-rules/src/lib/mock/rules.mock.ts
```diff
@@ -0,0 +1,102 @@
...
+            }
+          ],
+          enabled: true
+        }
+      }
+    ]
+  }
+};
+
+export const dummyRules: Partial<Rule>[] = [
```
TomikaArome, line 58: Replace the `Partial<Rule>` with `Rule` here as well

projects/aca-folder-rules/src/lib/services/folder-rules.service.ts
```diff
@@ -0,0 +1,79 @@
...
+      );
+
+    this.loadingSource.next(true);
+  }
+
+  private apiCall(path: string, httpMethod: string, params?: any[]): Promise<any> {
+    return this.apiService.getInstance().contentClient.callApi(path, httpMethod, ...params);
+  }
+
+  formatRules(res): Rule[] {
```
TomikaArome, line 60: We could also make this method private actually as it's only used internally.

projects/aca-folder-rules/src/lib/services/folder-rules.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    return res.list.entries.map((entry) => this.formatRule(entry.entry));
+  }
+
+  formatRule(obj): Rule {
```
TomikaArome, line 64: Same here.

TomikaArome (approved):
Looking good. Just a couple very minor things you could change. Otherwise see if the Travis pipeline check passes, but in any case, don't merge until we're allowed to again.

projects/aca-folder-rules/src/lib/services/folder-rules.service.ts
```diff
@@ -0,0 +1,79 @@
...
+  providedIn: 'root'
+})
+export class FolderRulesService {
+  private rulesListingSource = new BehaviorSubject<Rule[]>([]);
+  rulesListing$: Observable<Rule[]> = this.rulesListingSource.asObservable();
+  private loadingSource = new BehaviorSubject<boolean>(false);
+  loading$ = this.loadingSource.asObservable();
+  constructor(private apiService: AlfrescoApiService) {}
+
+  loadRules(nodeId: string, ruleSetId: string = '-default-'): void {
```
KikoUK, line 42: Shouldn't we name this method `getRules`?

TomikaArome: `loadRules` is appropriate in my opinion, since it doesn't return the rules itself but outputs them to the observable defined higher up in the service

DenysVuika (approved):

PR #2596, opened 2022-08-16
[ACS-3339] Manage Rules screen / listing - Manage rules screen template & minimal listing

nikita-web-ua:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

Rules cannot currently be viewed or managed through the ACA interface.

Now a simple list of the rules set for the folder can be seen in the rules management screen.

Loading bar
Screenshot 2022-08-17 at 11 55 13

List of rules and rule details
Screenshot 2022-08-19 at 17 09 00

Folder has no rules defined yet.
Screenshot 2022-08-17 at 11 58 04

Wrong node id  
Screenshot 2022-08-17 at 11 58 19

> - [ ] Yes
> - [X] No

**Other information**:


projects/aca-folder-rules/src/lib/services/folder-rules.service.ts
```diff
@@ -0,0 +1,105 @@
...
+  private formatRule(obj): Rule {
+    return {
+      id: obj.id,
+      name: obj.name ?? '',
+      description: obj.description ?? '',
+      enabled: obj.enabled ?? true,
+      cascade: obj.cascade ?? false,
+      asynchronous: obj.asynchronous ?? false,
+      errorScript: obj.errorScript ?? '',
+      shared: obj.shared ?? false,
```
tpage-alfresco, line 99: This is now called `isShared`.  The other booleans also need updating at some point (see https://alfresco.atlassian.net/browse/ACS-3358 and https://alfresco.atlassian.net/browse/ACS-3359)

projects/aca-folder-rules/src/lib/model/rule.model.ts
```diff
@@ -34,7 +35,7 @@ export interface Rule {
   asynchronous: boolean;
   errorScript: string;
   shared: boolean;
-  triggers: ('inbound' | 'update' | 'outbound')[];
+  triggers: ('INBOUND' | 'UPDATE' | 'OUTBOUND')[];
```
tpage-alfresco, line 38: I think these should be lower case. If the API is returning upper case then it's a bug.

nikita-web-ua: > I think these should be lower case. If the API is returning upper case then it's a bug.

Yes, you are right. It was a bug, but now it is fixed as far as I can see.

projects/aca-folder-rules/src/lib/manage-rules/manage-rules.smart-component.html
```diff
@@ -0,0 +1,56 @@
...
+
+  <aca-page-layout-content>
+    <div class="main-content">
+
+      <ng-container *ngIf="isLoading$ | async; else onLoaded">
+        <mat-progress-bar color="primary" mode="indeterminate"></mat-progress-bar>
+      </ng-container>
+
+      <ng-template #onLoaded>
+        <ng-container *ngIf="folderInfo; else genericError">
```
TomikaArome, line 20: Use the `async` pipe here.

projects/aca-folder-rules/src/lib/manage-rules/manage-rules.smart-component.scss
```diff
@@ -0,0 +1,15 @@
+.aca-rules-actions-bar {
+  padding: 0 30px;
+
+  .aca-rules-breadcrumb {
```
TomikaArome, line 4: Any way you could try and fix the vertical alignment in the breadcrumb? If not, don't worry too much about it.

projects/aca-folder-rules/src/lib/manage-rules/manage-rules.smart-component.html
```diff
@@ -0,0 +1,56 @@
...
+              <mat-icon>folder</mat-icon>
+              <adf-breadcrumb root="{{folderInfo.name}}:{{'ACA_FOLDER_RULES.TOOLBAR.BREADCRUMB.RULES' | translate}}"
+                              class="aca-rules-breadcrumb"></adf-breadcrumb>
+            </adf-toolbar-title>
+          </adf-toolbar>
+          <mat-divider></mat-divider>
+
+          <div class="aca-manage-rules-container" *ngIf="(rules$ | async).length > 0 ; else emptyContent">
+            <aca-rules-list [rules]="rules$ | async"></aca-rules-list>
+            <aca-rule-details [readOnly]="true"></aca-rule-details>
```
TomikaArome, line 32: There is no value filled in here. However I realise now that my rule details component currently only takes an `initialValue`, rather than an input property you can actually change, so that's my bad. Maybe provide an `[initialValue]` here anyway with the value of the selected rule, and I'll change it so it actually changes the value of the form.

projects/aca-folder-rules/src/lib/mock/node.mock.ts
```diff
@@ -0,0 +1,19 @@
+export const dummyNodeInfo = {
+  aspectNames: ['rule:rules', 'cm:titled', 'cm:auditable'],
+  createdAt: '2022-08-16T07:58:21.416+0000',
+  isFolder: true,
+  isFile: false,
+  createdByUser: {
+    id: 'hruser',
```
TomikaArome, line 7: Perhaps let's not provide real data here for a mock. Even if there's no password, let's keep it generic enough.

projects/aca-folder-rules/src/lib/rules-list/rule/rule.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="aca-rule">
+  <div class="aca-rule-checkbox">
+    <mat-checkbox></mat-checkbox>
+  </div>
```
TomikaArome, line 4: I'd remove the checkbox for now until we're ready to add multi-select. It's not necessary for now.

projects/aca-folder-rules/src/lib/rules-list/rule/rule.component.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'aca-rule',
+  templateUrl: 'rule.component.html',
+  styleUrls: ['rule.component.scss']
+})
+export class RuleComponent {
```
TomikaArome, line 33: This is far too generic of a name for a component. It could refer to any component related to the rule. Name it `RuleListItemUiComponent`

projects/aca-folder-rules/src/lib/rules-list/rules-list.component.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { Component, Input } from '@angular/core';
+import { Rule } from '../model/rule.model';
+
+@Component({
+  selector: 'aca-rules-list',
+  templateUrl: 'rules-list.component.html',
+  styleUrls: ['rules-list.component.scss']
+})
+export class RulesListComponent {
```
TomikaArome, line 34: Rename to `RulesListUiComponent`, as well as naming the file `rules-list.ui-component.ts`. This is just a convention we've more recently adopted to distinguish UI components from smart components.

TomikaArome (changes requested):
Good start. There's something missing though: clicking on a rule list item should show the rule details on the right hand side. This is partially a miss on my part due to not yet providing a way of changing the value of the rule details component from the outside other than for an initial value, but the PR should still cover it.

I found it interesting the decision to make specific components for the rule list and the rule list item even for such simple templates. I'm not against it personally though, it's part of the SOLID principals to make sure each part of the code has a single responsibility and I suppose making more components fulfils this. But it also does bloat the number of files we have and could potentially make things slightly more complex later on if you need to bubble events up. Keep it how you did it for now though.

projects/aca-folder-rules/src/lib/mock/node.mock.ts
```diff
@@ -0,0 +1,19 @@
+export const dummyNodeInfo = {
```
TomikaArome, line 1: We still need licensing information on every TypeScript file. I'm surprised the linter doesn't pick up on that.

projects/aca-folder-rules/src/lib/rules-list/rules-list.ui-component.html
```diff
@@ -0,0 +1,3 @@
+<div class="aca-rules-list">
+  <aca-rule *ngFor="let rule of rules" [rule]="rule" (click)="onRuleClicked(rule)"></aca-rule>
```
TomikaArome, line 2: Is there any visual change (e.g. change of background colour) to indicate a selected rule?

TomikaArome (changes requested):
Very good progress. Besides the other comments in my review, two more things:
- Since you modified the folder rules service, have you added any unit tests/changed existing ones to test for the node info?
- You'll need to rebase your branch following my merge yesterday. This will be a typical rebase and shouldn't require going into interactive mode like last time as what's happened is we both modified the `en.json` and `folder-rules.module.ts` files.

TomikaArome (approved):

PR #2603, opened 2022-08-19
[ACS-3256] Add trigger / "when" section to rule details & edit rule dialog

TomikaArome:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACS-3256

Adds the trigger / "when" section to the rule details component (which is found itself in the edit rule dialog).

Also provides a couple of small fixes to the rule details components:
- Changes the input property of the rule details component from "initialValue" to just "value", meaning it can be modified beyond the initial construction of the rule details component.
- Handles the case in which an unknown field is retrieved from the API, in which case it will show it in the field select box but will disappear as soon as the user selects a different option.

Create dialog is now openable using the "New rule" button at the top of the manage rules screen. However, it does not yet perform any action on press of the "Submit" button.

Default state of the checkboxes when the create dialog is opened.
image

An error is displayed when no checkboxes are pressed
image

Only the selected triggers are shown (without their checkboxes) in read only mode.
image

> - [ ] Yes
> - [x] No

**Other information**:


projects/aca-folder-rules/src/lib/rule-details/rule-details.ui-component.ts
```diff
@@ -102,6 +123,7 @@ export class RuleDetailsUiComponent implements OnInit, OnDestroy {
     this.formValidationChanged.emit(this.form.valid);
 
     this.form.valueChanges.pipe(takeUntil(this.onDestroy$)).subscribe((newFormValue: any) => {
+      console.log(newFormValue);
```
DenysVuika, line 126: this line should not be committed

DenysVuika (changes requested):

DenysVuika (approved):

PR #2636, opened 2022-09-14
[ACS-3512] - ACA enable/disable rule from listing

nikita-web-ua:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

There is currently no option to enable/disable the rule.

For each rule, a toggle button has been added that enables / disables the rule:
Screenshot 2022-09-14 at 09 34 46

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/projects/aca-folder-rules/src/lib/rules-list/rule/rule-list-item.ui-component.ts
+++ b/projects/aca-folder-rules/src/lib/rules-list/rule/rule-list-item.ui-component.ts
@@ -39,15 +39,14 @@ export class RuleListItemUiComponent {
 
   constructor(private folderRulesService: FolderRulesService) {}
 
-  onToggleClick(){
-    this.folderRulesService.toggleRule(this.nodeId, this.rule.id, this.parseRule(this.rule))
+  onToggleClick() {
+    this.folderRulesService.toggleRule(this.nodeId, this.rule.id, this.parseRule(this.rule));
   }
 
   private parseRule(rule): Rule {
-    let parsedRule = {...rule , enabled: !this.rule.enabled }
+    const parsedRule = { ...rule, enabled: !this.rule.enabled };
     delete parsedRule.actions[0].params.actionContext;
     parsedRule.conditions.booleanMode = parsedRule.conditions.booleanMode.toUpperCase();
-    return parsedRule
+    return parsedRule;
   }
 }
-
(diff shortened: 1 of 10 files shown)


DenysVuika (approved):

projects/aca-folder-rules/src/lib/rules-list/rule/rule-list-item.ui-component.ts
```diff
@@ -34,4 +35,18 @@ import { Rule } from '../../model/rule.model';
...
+  @Input() nodeId: string;
+
+  constructor(private folderRulesService: FolderRulesService) {}
+
+  onToggleClick() {
+    this.folderRulesService.toggleRule(this.nodeId, this.rule.id, this.parseRule(this.rule));
+  }
+
+  private parseRule(rule): Rule {
+    const parsedRule = { ...rule, enabled: !this.rule.enabled };
```
TomikaArome, line 47: Could we try and obtain this value directly from the mat-slide-toggle rather than just negating it?

projects/aca-folder-rules/src/lib/rules-list/rule/rule-list-item.ui-component.ts
```diff
@@ -34,4 +35,18 @@ import { Rule } from '../../model/rule.model';
...
+    delete parsedRule.actions[0].params.actionContext;
+    parsedRule.conditions.booleanMode = parsedRule.conditions.booleanMode.toUpperCase();
```
TomikaArome, line 49: Why do we need to do this here?

TomikaArome (commented):
Good work. Just a couple of non-critical comments.

projects/aca-folder-rules/src/lib/rules-list/rule/rule-list-item.ui-component.ts
nikita-web-ua: - we need delete `actionContext`  as it is read-only property, if we include it to the request, we receive an error.
- server receives only uppercase 'AND / OR' (bug)

However, those two seems to be solved. Tom's PR about the first point is already open, and Kristian fixed the second issue, but the env is not updated yet.

TomikaArome (approved):

PR #2713, opened 2022-10-13
[ACS-3576] Map aspect actions with the aspect picker dropdown

nikita-web-ua:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

Currently there is a blank field for add/remove aspect actions 

Screenshot 2022-10-14 at 12 18 05

Now `CardViewSelectItemModel` is used to display a list of all aspects

Screenshot 2022-10-14 at 12 18 43

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.html
+++ b/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.html
@@ -16,6 +16,7 @@
   <adf-card-view
     data-automation-id="rule-action-card-view"
     [properties]="cardViewItems"
+    [ngStyle]="isFullWidth && {'width': '100%'}"
     [editable]="!readOnly">
   </adf-card-view>
 
--- a/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.html
+++ b/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.html
@@ -16,7 +16,7 @@
   <adf-card-view
     data-automation-id="rule-action-card-view"
     [properties]="cardViewItems"
-    [ngStyle]="isFullWidth && {'width': '100%'}"
+    [ngStyle]="cardViewStyle"
     [editable]="!readOnly">
   </adf-card-view>
 
--- a/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
+++ b/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
@@ -183,7 +183,7 @@ export class RuleActionUiComponent implements ControlValueAccessor, OnInit, OnDe
             value: this.parameters[paramDef.name] ?? false
           });
         case 'd:qname':
-          if (paramDef.name === 'aspect-name') {
+          if (paramDef.name === 'aspect-name' && !this.readOnly) {
             this.isFullWidth = true;
             return new CardViewSelectItemModel({
               ...cardViewPropertiesModel,
--- a/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
+++ b/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
@@ -102,7 +102,7 @@ export class RuleActionUiComponent implements ControlValueAccessor, OnInit, OnDe
   }
 
   get cardViewStyle() {
-    return this.isFullWidth ? { ' width': '100%' } : {};
+    return this.isFullWidth ? { width: '100%' } : {};
   }
 
   onChange: (action: RuleAction) => void = () => undefined;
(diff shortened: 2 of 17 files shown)


projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.html
```diff
@@ -16,6 +16,7 @@
   <adf-card-view
     data-automation-id="rule-action-card-view"
     [properties]="cardViewItems"
+    [ngStyle]="isFullWidth && {'width': '100%'}"
```
DenysVuika, line 19: should there be OR instead of AND ?

nikita-web-ua: > should there be OR instead of AND ?

I don't think so - when isFillWidth = true => set width to 100%.
Correct me if I am wrong

DenysVuika (commented):
Looks good, the only concern is the full width and "AND" clause

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.html
TomikaArome: If I'm not mistaken this should just evaluate to `true` if `isFullWidth` is true?

TomikaArome: Wait no I get it, it does evaluate to `{'width': '100%'}`. Can you create a getter for this?
```
get cardViewStyle {
  return this.isFullWidth ? {'width': '100%'} : {};
}
```

nikita-web-ua: done

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
```diff
@@ -82,10 +94,14 @@ export class RuleActionUiComponent implements ControlValueAccessor, OnInit, OnDe
...
 
+  get cardViewStyle() {
+    return this.isFullWidth ? { ' width': '100%' } : {};
+  }
+
   onChange: (action: RuleAction) => void = () => undefined;
   onTouch: () => void = () => undefined;
 
-  constructor(private cardViewUpdateService: CardViewUpdateService) {}
+  constructor(private cardViewUpdateService: CardViewUpdateService, private folderRulesService: FolderRulesService) {}
```
TomikaArome, line 104: This is a UI component, meaning it should not have any business logic in it. `CardViewUpdateService` is acceptable because it's a service which interacts directly with the card view component but the `FolderRuleService` calls the API. Can you do something similar to what I did with the `actionDefinitions`?

TomikaArome (changes requested):

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
TomikaArome: To be clear, ideally we'd want an input property which takes a list of aspects, which could originate from anywhere, not just the API. It makes the component more reusable. The `edit-rule-dialog` and `manage-rules` components are, on the other hand, smart component which can use these services.

nikita-web-ua: Should `rule-action.ui-component.ts` receive already parsed aspects or a raw list. In other words I am not sure if `parseAspectsToSelectOptions` method can be a part of the `rule-action.ui-component.ts` component?

nikita-web-ua:
Still not sure about `parseAspectsToSelectOptions` as it is a business logic which we should avoid in ui components, but at the same time it makes `rule-action.ui-component` more convenient

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
TomikaArome: `parseAspectsToSelectOptions` sounds to me like component logic. If it's something only the component needs for a select dropdown contained within, I think it's good where it is.

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
```diff
@@ -189,4 +223,24 @@ export class RuleActionUiComponent implements ControlValueAccessor, OnInit, OnDe
...
+          return 1;
+        }
+        if (b.label === '' || b.label === null) {
+          return -1;
+        }
+        if (a.label === b.label) {
+          return 0;
+        }
+        return a.label < b.label ? -1 : 1;
+      })
```
TomikaArome, line 240: You can use `localeCompare to simplify this:
```suggestion
      .sort((a, b) => a.label?.localeCompare(b.label) ?? -1)
```

projects/aca-folder-rules/src/lib/services/folder-rules.service.ts
```diff
@@ -174,13 +174,17 @@ export class FolderRulesService {
   }
 
   loadAspects(): void {
```
TomikaArome, line 176: This is more relevant to the actions service since it calls the `action-parameter-constraints` endpoint. Could you move it there?

TomikaArome (changes requested):
Good work! Sorry for the wait, here's my review finally. Just a couple things, although we could do with moving anything related to action constraints to the action service where it's more appropriate.

In general though this is good for now, I do wonder if we don't want to find a different solution for this in the future, since if/when we add more action constraints, it will start to crowd a the rule-action.ui-component with more and more specific case scenarios. Also right now we're loading in the aspects without necessarily needing them, but we can refactor later on with a better solution if we think of one.

projects/aca-folder-rules/src/lib/services/folder-rules.service.ts
nikita-web-ua: Moved.

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
nikita-web-ua: Applied, thanks.

DenysVuika (approved):

PR #2768, opened 2022-11-10
[ACS-3549] add role dialog in manage version dialog

rbahirsheth:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Element must only use allowed ARIA attributes.

Fix the above issue by adding role dialog in manage version dialog

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/app/src/app/services/content-management.service.ts
+++ b/app/src/app/services/content-management.service.ts
@@ -579,7 +579,7 @@ export class ContentManagementService {
         title: 'VERSION.DIALOG.TITLE'
       };
       this.newVersionUploaderService
-        .openUploadNewVersionDialog(newVersionUploaderDialogData, { width: '630px' })
+        .openUploadNewVersionDialog(newVersionUploaderDialogData, { width: '630px', role: 'dialog' })
         .subscribe((newVersionUploaderData: NewVersionUploaderData) => {
           switch (newVersionUploaderData.action) {
             case NewVersionUploaderDataAction.refresh:


app/src/app/services/content-management.service.ts
```diff
@@ -579,7 +579,7 @@ export class ContentManagementService {
         title: 'VERSION.DIALOG.TITLE'
       };
       this.newVersionUploaderService
-        .openUploadNewVersionDialog(newVersionUploaderDialogData, { width: '630px' })
+        .openUploadNewVersionDialog(newVersionUploaderDialogData, { width: '630px', role: 'dialog' })
```
MichalKinas, line 582: Is this role required? According to material documentation -> MatDialog creates modal dialogs that implements the ARIA role="dialog" pattern by default. You can check it out here: https://material.angular.io/components/dialog/overview#accessibility

rbahirsheth: You are right MatDialog adds ARIA role by default. But, somehow it is not creating the mat-dialog-container with role="dialog".
As you can see the below screenshot
image

MichalKinas: Have you tried investigating why this happens in our dialog? I mean why the role isn't applied as it should be

rbahirsheth: I tried but did not get the reason of this.

DenysVuika (approved):

PR #2800, opened 2022-11-21
[ACS-3860] Folder rules - Node picker for action parameters of type "d:noderef"

nikita-web-ua:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

Currently for action parameters of type "d:noderef" there is text input, where user should manually enter folder id.

Now for action parameters of type "d:noderef" node selection dialog opens to select a node rather than a simple text box.

Screenshot 2022-11-22 at 08 44 35
Screenshot 2022-11-22 at 08 46 15

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
+++ b/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
@@ -35,7 +35,7 @@ import {
   CardViewUpdateService,
   UpdateNotification
 } from '@alfresco/adf-core';
-import { ActionParameterDefinition } from '@alfresco/js-api';
+import { ActionParameterDefinition, Node } from '@alfresco/js-api';
 import { of, Subject } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 import { ActionParameterConstraint, ConstraintValue } from '../../model/action-parameter-constraint.model';
@@ -224,12 +224,12 @@ export class RuleActionUiComponent implements ControlValueAccessor, OnInit, OnDe
     });
   }
 
-  openSelectorDialog() {
+  private openSelectorDialog() {
     const data: ContentNodeSelectorComponentData = {
       title: this.translate.instant('ACA_FOLDER_RULES.RULE_DETAILS.PLACEHOLDER.CHOOSE_FOLDER'),
       actionName: NodeAction.CHOOSE,
       currentFolderId: this._nodeId,
-      select: new Subject<any>()
+      select: new Subject<Node[]>()
     };
 
     this.dialog.open(ContentNodeSelectorComponent, {
@@ -239,7 +239,7 @@ export class RuleActionUiComponent implements ControlValueAccessor, OnInit, OnDe
     });
 
     data.select.subscribe(
-      (selections) => {
+      (selections: Node[]) => {
         if (selections[0].id) {
           this.writeValue({
             actionDefinitionId: this.selectedActionDefinitionId,
--- a/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
+++ b/projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
@@ -39,6 +39,11 @@ import { ActionParameterDefinition } from '@alfresco/js-api';
 import { of, Subject } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 import { ActionParameterConstraint, ConstraintValue } from '../../model/action-parameter-constraint.model';
+import {
+  ContentNodeSelectorComponent,
+  ContentNodeSelectorComponentData, NodeAction
+} from '@alfresco/adf-content-services';
+import { MatDialog } from '@angular/material/dialog';
 
 @Component({
   selector: 'aca-rule-action',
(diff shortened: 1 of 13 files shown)


projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
```diff
@@ -182,8 +195,19 @@ export class RuleActionUiComponent implements ControlValueAccessor, OnInit, OnDe
...
+            return new CardViewTextItemModel({
+              ...cardViewPropertiesModel,
+              icon: 'folder',
+              default: this.translate.instant('ACA_FOLDER_RULES.RULE_DETAILS.PLACEHOLDER.CHOOSE_FOLDER'),
+              clickable: true,
+              clickCallBack: this.openSelectorDialog.bind(this),
+              value: this.parameters[paramDef.name]
+            });
+          }
+        //  falls through
```
AleksanderSklorz, line 209: Is that comment needed?

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
```diff
@@ -200,6 +224,45 @@ export class RuleActionUiComponent implements ControlValueAccessor, OnInit, OnDe
     });
   }
 
+  openSelectorDialog() {
+    const data: ContentNodeSelectorComponentData = {
+      title: this.translate.instant('ACA_FOLDER_RULES.RULE_DETAILS.PLACEHOLDER.CHOOSE_FOLDER'),
+      actionName: NodeAction.CHOOSE,
+      currentFolderId: this._nodeId,
+      select: new Subject<any>()
```
AleksanderSklorz, line 232: Can we use different type than any?

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
```diff
@@ -200,6 +224,45 @@ export class RuleActionUiComponent implements ControlValueAccessor, OnInit, OnDe
     });
   }
 
+  openSelectorDialog() {
```
AleksanderSklorz, line 227: Small thing but it may be private because it is unused outside component.

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
AleksanderSklorz, line 209: Is that comment needed?

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
AleksanderSklorz, line 227: Small comment - that function can be probably private because it is not used outside component

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
AleksanderSklorz, line 232: is that possible to use different type than any?

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
nikita-web-ua: Yes, it is needed for linter. Otherwise it gives "_Expected a 'break' statement before 'default_'" error.

AleksanderSklorz: And you don't need break right? Because it says about break probably because if that condition in if: "if (!constraintsForDropdownBox && !this.readOnly)" gives false then it will go to next case which is default if break is missing.

nikita-web-ua: Right

AleksanderSklorz (approved):

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
nikita-web-ua: Changed to private, thanks

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
nikita-web-ua: Changed, thanks

projects/aca-folder-rules/src/lib/rule-details/actions/rule-action.ui-component.ts
```diff
@@ -56,6 +59,15 @@ import { ActionParameterConstraint, ConstraintValue } from '../../model/action-p
...
 })
 export class RuleActionUiComponent implements ControlValueAccessor, OnInit, OnDestroy {
+  private _nodeId = '';
+  @Input()
+  get nodeId(): string {
+    return this._nodeId;
+  }
+  set nodeId(value) {
+    this._nodeId = value;
+  }
```
TomikaArome, line 69: There's no point to doing this as opposed to just a normal property. Using a setter is useful if you have something extra to do to a value every time it's passed into the component.
```suggestion
  @Input()
  nodeId = '';
```

TomikaArome (changes requested):
LGTM. Good idea using the clickable icon in the card view. Just one change I can see and it should be good!

TomikaArome (approved):

PR #2808, opened 2022-11-23
[ACS-3895] ACA - Folder Rules: inherit rule sets toggle

nikita-web-ua:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

Currently in ACA, the user does not have the ability to choose whether a folder should inherit folder rule sets from parent folders or not.

Each folder has a button to toggle whether it should inherit folder rule sets from its parent folders.
Screenshot 2022-11-23 at 17 53 08

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/projects/aca-folder-rules/src/lib/manage-rules/manage-rules.smart-component.html
+++ b/projects/aca-folder-rules/src/lib/manage-rules/manage-rules.smart-component.html
@@ -35,7 +35,7 @@
               {{ 'ACA_FOLDER_RULES.MANAGE_RULES.TOOLBAR.ACTIONS.INHERIT_RULES' | translate }}
             </mat-slide-toggle>
 
-            <mat-divider vertical [style]="{height:'50%', margin:'0 12px'}"></mat-divider>
+            <mat-divider vertical class="vertical-divider"></mat-divider>
 
             <div class="aca-manage-rules__actions-bar__buttons">
               <button
--- a/projects/aca-folder-rules/src/lib/manage-rules/manage-rules.smart-component.html
+++ b/projects/aca-folder-rules/src/lib/manage-rules/manage-rules.smart-component.html
@@ -26,7 +26,13 @@
                               class="aca-manage-rules__actions-bar__title__breadcrumb"></adf-breadcrumb>
             </adf-toolbar-title>
 
-            <mat-slide-toggle disableRipple [labelPosition]="'before'" [aria-label]="'Inherit rules'">Inherit rules</mat-slide-toggle>
+            <mat-slide-toggle
+              [checked]="isInheritanceEnabled"
+              (change)="onToggleInheritanceClick($event)"
+              [disabled]="isInheritanceToggleDisabled"
+              [labelPosition]="'before'">
+              {{ 'ACA_FOLDER_RULES.MANAGE_RULES.TOOLBAR.ACTIONS.INHERIT_RULES' | translate }}
+            </mat-slide-toggle>
 
             <mat-divider vertical [style]="{height:'50%', margin:'0 12px'}"></mat-divider>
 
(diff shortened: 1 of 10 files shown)


projects/aca-folder-rules/src/lib/manage-rules/manage-rules.smart-component.html
```diff
@@ -26,13 +26,24 @@
...
+              data-automation-id="manage-rules-inheritance-toggle-button"
+              [checked]="isInheritanceEnabled"
+              (change)="onInheritanceToggleChange($event)"
+              [disabled]="isInheritanceToggleDisabled"
+              [labelPosition]="'before'">
+              {{ 'ACA_FOLDER_RULES.MANAGE_RULES.TOOLBAR.ACTIONS.INHERIT_RULES' | translate }}
+            </mat-slide-toggle>
+
+            <ng-container *ngIf="canEditRule(mainRuleSet$ | async)">
+              <mat-divider vertical [style]="{height:'50%', margin:'0 12px'}"></mat-divider>
```
MichalKinas, line 39: Can we move inline styles to some stylesheet?

projects/aca-folder-rules/src/lib/manage-rules/manage-rules.smart-component.ts
```diff
@@ -57,6 +60,7 @@ export class ManageRulesSmartComponent implements OnInit, OnDestroy {
   hasMoreRuleSets$: Observable<boolean>;
   ruleSetsLoading$: Observable<boolean>;
   folderInfo$: Observable<NodeInfo>;
+  isInheritanceEnabled$: Observable<boolean>;
```
AleksanderSklorz, line 63: Seems unused?

projects/aca-folder-rules/src/lib/manage-rules/manage-rules.smart-component.html
nikita-web-ua: Moved

MichalKinas (approved):

AleksanderSklorz (approved):

projects/aca-folder-rules/src/lib/services/folder-rules.service.ts
```diff
@@ -152,6 +152,16 @@ export class FolderRulesService {
     );
   }
 
+  async getRuleSettings(nodeId: string, key: string = '-isInheritanceEnabled-'): Promise<RuleSettings> {
```
TomikaArome, line 155: Since the rule set service handles the node id, this should be in `FolderRuleSetsService`. We can load it at the same time in the `loadRuleSets` method by adding it in the `combineLatest`, here:
```
switchMap(() => combineLatest(
  this.getMainRuleSet(nodeId),
  loadInheritedRuleSets ? this.getInheritedRuleSets(nodeId) : of([])
))
```
This could become:
```
switchMap(() => combineLatest(
  this.getMainRuleSet(nodeId),
  loadInheritedRuleSets ? this.getInheritedRuleSets(nodeId) : of([]),
  this.getFolderInheritance(nodeId)
))
```
and then using the value in the subscribe to emit to the `isFolderInheritanceEnabledSource` BehaviourSubject.

TomikaArome (commented):
Good work!

Your method is fine, and I'm sure works well, but I'll admit it's not what I would have done. I'm not going to request changes due to the short time frame we have but see my review for what I think is a better approach.

Feel free to attempt it this afternoon if you're up to it and don't have anything else to do, otherwise I'm happy enough with this and you can merge it before the end of the day so it's hopefully propagated to the remote environment tomorrow.

projects/aca-folder-rules/src/lib/services/folder-rules.service.ts
nikita-web-ua: I made `getRuleSettings` more generic, so it could be used in future with other parameters. Do you suggest to redesign it to `getFolderInheritance` for now, so it will only load data for `isInheritanceEnabled$`?

TomikaArome: Considering it's likely the only rules / rule sets related setting for folders that we have, I don't think it's necessary to make it generic like this. Is this endpoint one of the new private ones? Did you check if there a generic js-api method that could allow you to call this endpoint and then use the folder rule (sets) service to be more specific about `isInheritanceEnabled`?

PR #2837, opened 2022-12-06
[ACS-3757] returning focus to element from which they were opened

AleksanderSklorz:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Modals which were opened from personal files/libraries list or toolbar or more menu, after closing, were not returning focus to element from which they were opened. 

After closing modals opened from personal files/libraries or toolbar or more menu, focus is returned to correct element. 

> - [ ] Yes
> - [x] No

**Other information**:
ADF PR: https://github.com/Alfresco/alfresco-ng2-components/pull/8034
APPS PR: https://github.com/Alfresco/alfresco-apps/pull/4630

### DIFF
--- a/app/src/app/content-plugin/services/content-management.service.ts
+++ b/app/src/app/content-plugin/services/content-management.service.ts
@@ -85,6 +85,8 @@ interface RestoredNode {
   providedIn: 'root'
 })
 export class ContentManagementService {
+  private static MORE_ACTIONS_BUTTON_SELECTOR = '#app\\.toolbar\\.more';
+
   constructor(
     private alfrescoApiService: AlfrescoApiService,
     private store: Store<AppStore>,
@@ -261,6 +263,7 @@ export class ContentManagementService {
       if (node) {
         this.alfrescoApiService.nodeUpdated.next(node);
       }
+      ContentManagementService.focusMoreActionsButton();
     });
   }
 
@@ -578,9 +581,8 @@ export class ContentManagementService {
         showVersionsOnly: true,
         title: 'VERSION.DIALOG.TITLE'
       };
-      this.newVersionUploaderService
-        .openUploadNewVersionDialog(newVersionUploaderDialogData, { width: '630px', role: 'dialog' })
-        .subscribe((newVersionUploaderData: NewVersionUploaderData) => {
+      this.newVersionUploaderService.openUploadNewVersionDialog(newVersionUploaderDialogData, { width: '630px', role: 'dialog' }).subscribe({
+        next: (newVersionUploaderData: NewVersionUploaderData) => {
           switch (newVersionUploaderData.action) {
             case NewVersionUploaderDataAction.refresh:
               this.store.dispatch(new ReloadDocumentListAction());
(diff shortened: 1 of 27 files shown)


MichalKinas (approved):

app/src/app/content-plugin/services/content-management.service.ts
```diff
@@ -1083,4 +1088,10 @@ export class ContentManagementService {
   private static focusCreateMenuButton(): void {
     document.querySelector<HTMLElement>('app-create-menu button').focus();
   }
+
+  private static focusAfterClose(focusedElementSelector: string): void {
```
DenysVuika, line 1092: why does it have to be `static` if it's private, you can remove the `static` and reduce the usage of `ContentManagementService.focusAfterClose` to just a method call.

app/src/app/content-plugin/services/node-actions.service.ts
```diff
@@ -687,4 +697,10 @@ export class NodeActionsService {
       return moveStatus;
     }
   }
+
+  private static focusAfterClose(focusedElementSelector: string): void {
```
DenysVuika, line 701: does not need to be private static as it does not bring any improvement, besides `NodeActionsService.focusAfterClose` in all the method calls does not make sense, as those are inside the same class

app/src/assets/app.extensions.json
```diff
@@ -401,7 +402,8 @@
         "title": "APP.ACTIONS.LEAVE",
         "icon": "exit_to_app",
         "actions": {
-          "click": "LEAVE_LIBRARY"
+          "click": "LEAVE_LIBRARY",
+          "focusedElementOnCloseSelector": "#app\\.toolbar\\.leaveLibrary"
```
DenysVuika, line 406: we are adding the notion of the application css and paths to the config file, not sure is a great idea

DenysVuika (changes requested):

app/src/assets/app.extensions.json
```diff
@@ -331,7 +331,7 @@
         "order": 100,
         "data": {
           "iconButton": true,
-          "focusedElementOnCloseSelector": "#share-action-button"
+          "autoFocusedTriggerAfterModalClose": true
```
DenysVuika, line 334: is it possible to make it `true` by default and avoid setting it for each entry here?

AleksanderSklorz: Well it was my first go before I change selector to boolean but when I remove that flag and assume that default is true, it breaks some options without modals. So I decided to leave that boolean to avoid more changes. 

But sure I will check it again to see how many extra changes is required to remove that flag. If there won't be too many then sure I will remove that flag. I will let you know when I check.

app/src/app/content-plugin/services/content-management.service.ts
AleksanderSklorz: done

app/src/app/content-plugin/services/node-actions.service.ts
AleksanderSklorz: done

app/src/assets/app.extensions.json
AleksanderSklorz: done

app/src/assets/app.extensions.json
AleksanderSklorz: @DenysVuika I removed these flags successfully so now there is no flags in json file.

AleksanderSklorz:
@DenysVuika I pushed interface as you suggested

DenysVuika (approved):

PR #2896, opened 2023-01-10
Revert "[ACS-4108] rule actions were showing id instead of names"

arohilaGL:
Reverts Alfresco/alfresco-content-app#2888


arohilaGL:
Had to revert the branch, made a wrong change.

DenysVuika (approved):

PR #2947, opened 2023-01-31
[ACS-4251] Added keyboard navigation for toolbar menu items

swapnil-verma-gl:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Not all toolbar menu items could be navigated to using keyboard shortcuts. Only items wrapped with `<adf-toolbar-menu-item>` could be navigated. Those wrapped with `<adf-dynamic-component>` could not be navigated.

Implemented workaround where all  menubar items can be navigated to. This was done by wrapping the items with a button with the mat-menu-item directive, and then manually calling the click event on them when the button was clicked. 

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
+++ b/projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
@@ -12,14 +12,19 @@
 </button>
 
 <mat-menu #menu="matMenu" [overlapTrigger]="false">
-  <ng-container *ngFor="let child of actionRef.children; trackBy: trackByActionId">
-    <ng-container [ngSwitch]="child.type">
+  <ng-container *ngFor="let child of actionRef.children; trackBy: trackByActionId" [ngSwitch]="child.type">
       <ng-container *ngSwitchCase="'custom'">
-        <adf-dynamic-component [id]="child.component" [data]="child.data"></adf-dynamic-component>
+        <button mat-menu-item style="padding: 0" (click)="onCustomItemContainerClick($event)">
+          <adf-dynamic-component [id]="child.component" [data]="child.data"></adf-dynamic-component>
+        </button>
       </ng-container>
-      <ng-container *ngSwitchDefault>
+      <ng-container *ngSwitchCase="'separator'">
         <app-toolbar-menu-item [actionRef]="child" [menuId]="actionRef.id"></app-toolbar-menu-item>
       </ng-container>
-    </ng-container>
+      <ng-container *ngSwitchDefault>
+        <button mat-menu-item style="padding: 0" (click)="onCustomItemContainerClick($event)">
+          <app-toolbar-menu-item [actionRef]="child" [menuId]="actionRef.id"></app-toolbar-menu-item>
+        </button>
+      </ng-container>
   </ng-container>
 </mat-menu>
--- a/projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
+++ b/projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
@@ -12,9 +12,10 @@
 </button>
 
 <mat-menu #menu="matMenu" [overlapTrigger]="false">
-  <ng-container *ngFor="let child of actionRef.children; trackBy: trackByActionId" [ngSwitch]="child.type">
+  <ng-container *ngFor="let child of actionRef.children; trackBy: trackByActionId">
+    <ng-container [ngSwitch]="child.type">
       <ng-container *ngSwitchCase="'custom'">
-        <button mat-menu-item style="padding: 0" (click)="onCustomItemContainerClick($event)">
+        <button mat-menu-item class="aca-mat-menu-item-container" (click)="onCustomItemContainerClick($event)">
           <adf-dynamic-component [id]="child.component" [data]="child.data"></adf-dynamic-component>
         </button>
       </ng-container>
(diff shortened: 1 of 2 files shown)


projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
```diff
@@ -12,14 +12,19 @@
 </button>
 
 <mat-menu #menu="matMenu" [overlapTrigger]="false">
-  <ng-container *ngFor="let child of actionRef.children; trackBy: trackByActionId">
-    <ng-container [ngSwitch]="child.type">
+  <ng-container *ngFor="let child of actionRef.children; trackBy: trackByActionId" [ngSwitch]="child.type">
       <ng-container *ngSwitchCase="'custom'">
-        <adf-dynamic-component [id]="child.component" [data]="child.data"></adf-dynamic-component>
+        <button mat-menu-item style="padding: 0" (click)="onCustomItemContainerClick($event)">
```
DenysVuika, line 17: don't embed styles in the components, we have css/scss for that

projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
```diff
@@ -12,14 +12,19 @@
...
+          <adf-dynamic-component [id]="child.component" [data]="child.data"></adf-dynamic-component>
+        </button>
       </ng-container>
-      <ng-container *ngSwitchDefault>
+      <ng-container *ngSwitchCase="'separator'">
         <app-toolbar-menu-item [actionRef]="child" [menuId]="actionRef.id"></app-toolbar-menu-item>
       </ng-container>
-    </ng-container>
+      <ng-container *ngSwitchDefault>
+        <button mat-menu-item style="padding: 0" (click)="onCustomItemContainerClick($event)">
```
DenysVuika, line 25: embedded styles

projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
```diff
@@ -12,14 +12,19 @@
 </button>
 
 <mat-menu #menu="matMenu" [overlapTrigger]="false">
-  <ng-container *ngFor="let child of actionRef.children; trackBy: trackByActionId">
-    <ng-container [ngSwitch]="child.type">
```
DenysVuika, line 16: is there a reason for removing this container?

DenysVuika (changes requested):

projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
swapnil-verma-gl: No there isn't a reason. Its likely that this was removed while trying out different approaches, and then slipped through during the code push. I'll revert this change in the next commit.

projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
swapnil-verma-gl: Agreed. I'll move the styles to the component.

projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
swapnil-verma-gl: Moved the styling to the component, and used class instead

projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
```diff
@@ -15,11 +15,18 @@
   <ng-container *ngFor="let child of actionRef.children; trackBy: trackByActionId">
     <ng-container [ngSwitch]="child.type">
       <ng-container *ngSwitchCase="'custom'">
-        <adf-dynamic-component [id]="child.component" [data]="child.data"></adf-dynamic-component>
+        <button mat-menu-item class="aca-mat-menu-item-container" (click)="onCustomItemContainerClick($event)">
```
DenysVuika, line 18: not sure this is correct, the dynamic component can be anything, and we should already have cases when it's a button (like toggle favorite)... so wrapping button with a button does not make sense

DenysVuika (changes requested):

projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.html
swapnil-verma-gl: I agree with the 'button in a button' aspect of this approach. However, the reason why we took this approach was due to how angular material handles menu items. The mat-menu-item directive is responsible for handling a lot of functionality associated with the menu items. This also includes features such as keyboard navigation and interactions. However, according to some of the comments made by some developers from the @angular/components repository as well as logged issues (links [here](https://github.com/angular/components/issues/25323#issuecomment-1222093609) and [here](https://github.com/angular/components/issues/18123), unfortunately I couldn't find any official 'documentation' for this point), this directive only works when it is a direct child of a mat-menu component. It WILL NOT WORK if the mat-menu item is added to the mat-menu via content-projection (for e.g. with custom components such as  adf-dynamic-components as well as adf-toolbar-menu-item). Right now, the buttons inside adf-dynamic-component and adf-toolbar-menu-item are the ones that have the mat-menu-item directive.
The reason why it seems to currently work with adf-toolbar-menu-items is due to a workaround implemented in that component where we manually push the menu items into the mat-menu (see [ngAfterViewInit()](https://github.com/Alfresco/alfresco-content-app/blob/develop/projects/aca-shared/src/lib/components/tool-bar/toolbar-menu/toolbar-menu.component.ts#L59) implementation). I think that this kind of approach might not be possible with the adf-dynamic-components because it can be designed in any manner outside of our control, by third party developers using our applications. Restricting them to follow a certain kind of component architecture, so that we can make changes similar to adf-toolbar-menu-items is something that can be done, but might not be desirable.
As for why we used a button, we could use a div/span with a custom click handler and keydown.enter/keydown.space events. However, there might be some accessibility issues related to using them (for e.g. those related to  focus and screen reader related). We can investigate into this aspect a bit more if needed though.

swapnil-verma-gl:
Accessibility issues have been re prioritised, and will be handled in the future if required. Closing this PR.

PR #2998, opened 2023-02-17
[MOBILEAPPS-1701] Custom URL schema issues fixed

jatin2008:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/MOBILEAPPS-1701 issues

1. Not displaying dialog once user close the dialog.
2. dialog name is localized now.
3. dialog is now converted to floating action button and will not dismiss until the user closed it.
4. dialog title is not scrollable now.

> - [ ] Yes
> - [x] No

**Other information**:


projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.ts
```diff
@@ -51,4 +52,10 @@ export class OpenInAppComponent {
   openInApp(): void {
     this.window.location.href = this.redirectUrl;
   }
+
+  onCloseDialog(): void {
```
DenysVuika, line 56: what about clicking outside the dialog? it will close it without calling this event

jatin2008: If user will click outside dialog will not close as this is the requirement until and unless user close the dialog with close icon.

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.scss
```diff
@@ -6,8 +6,12 @@
...
 .mat-dialog-container{
-  padding: 12px;
+  padding: 5px;
   border-radius: 36px;
   background-color: var(--theme-blue-button-color);
   color: var(--theme-about-panel-background-color);
 }
+
+.open-in-app {
+  overflow-x: hidden !important;
```
AleksanderSklorz, line 16: do you need !important? Can't you change somehow your selector used in scss so it won't need !important?

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.spec.ts
```diff
@@ -35,4 +48,10 @@ describe('OpenInAppComponent', () => {
 
     expect(currentLocation).toBe('mockRedirectUrl');
   });
+
+  it('should set the value `mobile_notification_expires_in` in session storage on dialog close', async () => {
+    component.onCloseDialog();
+    expect(sessionStorage.getItem('mobile_notification_expires_in')).not.toBe(null);
```
AleksanderSklorz, line 54: instead of .not.toBe(null) you can use: toBeDefined()/toBeTruthy()/not.toBeNull()

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.scss
jatin2008: yess i checked without !important as this style is applied to button from angular material but without !important it will not override the mat style.

projects/aca-shared/src/lib/services/aca-mobile-app-switcher.service.ts
```diff
@@ -46,31 +46,48 @@ export class AcaMobileAppSwitcherService {
...
-    const currentTime: number = new Date().getTime();
-    const sessionTime: string = sessionStorage.getItem('mobile_notification_expires_in');
+  resolveExistenceOfDialog(): void {
+    const url: string = this.getCurrentUrl();
+    const queryParams: string = url.split('?')[1];
+    let queryParamsList = [];
+    let hideBanner = false;
+    if (queryParams !== null && queryParams !== undefined) {
+      queryParamsList = queryParams.split('&');
+      queryParamsList.forEach((param) => {
```
AleksanderSklorz, line 56: You can replace forEach with 'some' function from array probably

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.scss
AleksanderSklorz: Did you try something like: 
.material-css-selector.open-in-app {
  ....
}

You may need to combine your css class with material class which is applied to button to override it without important

projects/aca-shared/src/lib/services/aca-mobile-app-switcher.service.ts
jatin2008: we are just iterating throught the array and retrieving the queryParams for checking only and not working on the result of the array so can't it be fine

AleksanderSklorz: well you can do something like: 
```ts
hideBanner = queryParamsList.some((param) => {
  const key = param.split('=')[0];
  const value = param.split('=')[1];
  return key === 'mobileapps' && value === 'true'; 
});
```

It wold be better because is more descriptive that you are looking for boolean and also loop will finish when at least one will be true instead of like for forEach - going till end of array.

AleksanderSklorz: You can even make it shorter like: 
```ts
hideBanner = queryParamsList.some((param) => return param.split('=')[0] === 'mobileapps' && param.split('=')[1] === 'true');
```

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.scss
jatin2008: it is not applying while i am doing like this, so can you please just check

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.spec.ts
jatin2008: instead of .not.toBe(null) I have used now - not.toBeNull()

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.scss
AleksanderSklorz: Not sure why it is not applying I checked below on some button: 
image


You see I added overflow-x: hidden to 'test' class which I applied to button and then in computed I can see it is changed: 
image


but as I said, I tried it on different button

projects/aca-shared/src/lib/services/aca-mobile-app-switcher.service.ts
jatin2008: I now changed to a more shorter form as you said

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.scss
jatin2008: It is not applying in my case, don't know why, I also checked with material-css-selector

Screenshot from 2023-02-17 20-47-30

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.spec.ts
```diff
@@ -35,4 +48,10 @@ describe('OpenInAppComponent', () => {
 
     expect(currentLocation).toBe('mockRedirectUrl');
   });
+
+  it('should set the value `mobile_notification_expires_in` in session storage on dialog close', async () => {
+    component.onCloseDialog();
```
DenysVuika, line 53: Can you also check that the local storage was empty before calling a dialog

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.scss
AleksanderSklorz: Just to ensure. When you are saying that you checked with "material-css-selector" do you mean that you used "material-css-selector" sentence in your styles or you replaced it with existing material class? I said "material-css-selector" as placeholder for real material class so it should be something like: 

.open-in-app.mat-button {
    overflow-x: hidden;
}

jatin2008: yess you are right, I now applied the desired style and it's working now :)

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.spec.ts
jatin2008: I now have cleared the session storage before each test run to avoid tearing mutated storage

DenysVuika (approved):

projects/aca-shared/src/lib/components/open-in-app/open-in-app.component.scss
AleksanderSklorz: thank you

AleksanderSklorz (approved):

PR #3080, opened 2023-03-21
[ACS-4436] comparator options set as Equals only when Encoding is selected in condition in Create/Edit a rule in Manage rules same as Share application

jatin2008:
…as Share application which was earlier “Equals, Contains, Starts with and Ends with”

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

we have comparator values “Equals, Contains, Starts with and Ends with” when Encoding is selected in Condition but on Share app we can see only “Equals” for Encoding. 

comparator options set as Equals only when Encoding is selected in condition in Create/Edit a rule in Manage rules same as Share application.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/projects/aca-folder-rules/src/lib/rule-details/conditions/rule-condition-fields.ts
+++ b/projects/aca-folder-rules/src/lib/rule-details/conditions/rule-condition-fields.ts
@@ -23,7 +23,7 @@
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  */
 
-export type RuleConditionFieldType = 'string' | 'number' | 'date' | 'type' | 'special' | 'equals_comparator';
+export type RuleConditionFieldType = 'string' | 'number' | 'date' | 'type' | 'special';
 
 export interface RuleConditionField {
   name: string;
@@ -50,7 +50,7 @@ export const ruleConditionFields: RuleConditionField[] = [
   {
     name: 'encoding',
     label: 'ACA_FOLDER_RULES.RULE_DETAILS.FIELDS.ENCODING',
-    type: 'equals_comparator'
+    type: 'special'
   },
   {
     name: 'category',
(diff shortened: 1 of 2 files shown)


projects/aca-folder-rules/src/lib/rule-details/conditions/rule-condition-fields.ts
```diff
@@ -23,7 +23,7 @@
  * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
  */
 
-export type RuleConditionFieldType = 'string' | 'number' | 'date' | 'type' | 'special';
+export type RuleConditionFieldType = 'string' | 'number' | 'date' | 'type' | 'special' | 'equals_comparator';
```
TomikaArome, line 26: "equals_comparator" isn't a type. Can we use `'special'` similar to how "category", "tag" and "aspect" do it. This  should hide the comparator field entirely (since equals is the only option"). See here how it is included as a comparator used for the `'special'` type (even though it is eventually hidden by the component): https://github.com/Alfresco/alfresco-content-app/blob/develop/projects/aca-folder-rules/src/lib/rule-details/conditions/rule-condition-comparators.ts#L40

TomikaArome (changes requested):

projects/aca-folder-rules/src/lib/rule-details/conditions/rule-condition-fields.ts
jatin2008: I have now used 'special' comparator instead of 'equals_comparator'.

TomikaArome (approved):

AleksanderSklorz (approved):

nikita-web-ua (approved):

PR #3150, opened 2023-04-26
[ACA-4690] added width and max-width to sidebar

arohilaGL:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

The width of the Create tag keeps increasing indefinitely.

Limiting the width.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/projects/aca-shared/src/lib/components/page-layout/page-layout.component.scss
+++ b/projects/aca-shared/src/lib/components/page-layout/page-layout.component.scss
@@ -56,6 +56,14 @@
     }
   }
 
+  .sidebar {
+    display: block;
+    height: 100%;
+    overflow-y: scroll;
+    max-width: 350px;
+    width: 350px;
+  }
+
   .aca-page-layout-content {
     @include flex-row;
   }


projects/aca-shared/src/lib/components/page-layout/page-layout.component.scss
```diff
@@ -56,6 +56,14 @@
     }
   }
 
+  .sidebar {
```
AleksanderSklorz, line 59: I'm wondering if that is correct place for these styles. In html for this component I can't see that this class is used. Also note that this component has encapsulation: ViewEncapsulation.None, so having this class not nested in scss may cause that you affect other places with sidebar class too. @MichalKinas what do you think?

MichalKinas: Yeah I agree, component itself shouldn't style elements that it's not aware of. In this case element containing sidebar class comes from ng-content

arohilaGL: This was removed in a previous commit, I only identified what got removed and placed it back.

MichalKinas (approved):

AleksanderSklorz (approved):

projects/aca-shared/src/lib/components/page-layout/page-layout.component.scss
AleksanderSklorz: discussed on slack, please leave and for now and raise jira as suggested by @MichalKinas

PR #3195, opened 2023-05-16
[ACS-5171] Facets section - UI changes

nikita-web-ua:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [X] Other... Please describe: UI changes 

https://alfresco.atlassian.net/browse/ACS-5171
Screenshot 2023-05-16 at 22 43 07

Facets have been updated to follow new UI
Screenshot 2023-05-16 at 22 10 09

> - [ ] Yes
> - [X] No

**Other information**:


MichalKinas (approved):

projects/aca-content/src/lib/components/search/search-results/search-results.component.scss
```diff
@@ -1,6 +1,7 @@
 @import '../../../ui/mixins';
 
 $adf-chip-background: #efefef;
+$adf-facets-background: white;
```
AleksanderSklorz, line 4: If you open browser's console and scroll down you can see there is list of already defined variables: 
image

Some of them have already white value. Could you please verify if we can reuse them or you need create new one?

projects/aca-content/src/lib/components/search/search-results/search-results.component.html
```diff
@@ -15,20 +15,25 @@
...
-            <button
-              mat-button
-              adf-reset-search
-              class="content__reset-action"
-              title="{{'APP.BROWSE.SEARCH.RESET_ACTION' | translate }}"
-              [attr.aria-label]="'APP.BROWSE.SEARCH.RESET_ACTION' | translate ">
-              <mat-icon> refresh </mat-icon>
-            </button>
+            <div class="content__advanced-filters">
+              <div class="content__advanced-filters--header">
```
AleksanderSklorz, line 24: hmm should not we use some prefixes for classes? Like aca-?

AleksanderSklorz (approved):

projects/aca-content/src/lib/components/search/search-results/search-results.component.scss
nikita-web-ua: Yep, I saw those variables, but I was not sure if I should use them, what do you think?

AleksanderSklorz: I would say it is good to reuse existing assuming that its name matches to your case. For example if there is any variable named "modal" then this does not match to your case but if there are some more generic name then should be fine in my opinion. 

Generally variables are helpful to share style values between places so then we can change style for each place by changing just single value. If we define separate variable for each place then we lose that advantage because we need then change each place instead of changing just one place

projects/aca-content/src/lib/components/search/search-results/search-results.component.scss
```diff
@@ -24,6 +25,7 @@ $contrast-gray: #646569;
     padding: 0 25px;
     flex-direction: row;
     align-items: center;
+    border-top: 1px solid #eee;
```
swapnil-verma-gl, line 28: I feel like maybe we should have a variable here for the border color, instead of a set hex code? I do see a variable already defined for border colors - --theme-border-color, although it needs to be confirmed whether it can be applied here or not.

projects/aca-content/src/lib/components/search/search-results/search-results.component.html
nikita-web-ua: added prefixes

AleksanderSklorz: thanks

projects/aca-content/src/lib/components/search/search-results/search-results.component.scss
nikita-web-ua: good point, thanks

swapnil-verma-gl (approved):

DenysVuika (approved):

PR #3295, opened 2023-06-22
[MOBILEAPPS-1927] popup now visible after login without manual refresh

jatin2008:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

When we generate a private Url and open it in Incognito tab then after login popup was visible after login.

Now after geberating a private Url and opening it in an Incognito tab then after login popup is visible after few seconds without manual refresh of the page.
Jira id - https://alfresco.atlassian.net/browse/MOBILEAPPS-1927

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/projects/aca-shared/src/lib/services/app.service.ts
+++ b/projects/aca-shared/src/lib/services/app.service.ts
@@ -166,9 +166,11 @@ export class AppService implements OnDestroy {
       if (isReady) {
         this.loadRepositoryStatus();
         this.loadUserProfile();
-        setTimeout(() => {
-          this.openMobileAppDialog();
-        });
+        if (document.readyState === 'complete') {
+          setTimeout(() => {
+            this.openMobileAppDialog();
+          }, 10000);
+        }
       }
     });
 


projects/aca-shared/src/lib/services/app.service.ts
```diff
@@ -166,9 +166,11 @@ export class AppService implements OnDestroy {
       if (isReady) {
         this.loadRepositoryStatus();
         this.loadUserProfile();
-        setTimeout(() => {
-          this.openMobileAppDialog();
-        });
+        if (document.readyState === 'complete') {
```
DenysVuika, line 169: how that fixes the problem? if was not showing previously, before adding check for "readyState", it will not going to fire with this check too, as readyState might be loading... what does this check actually do to fix?

DenysVuika (changes requested):

projects/aca-shared/src/lib/services/app.service.ts
jatin2008: basically I added timer in settimeout, if you want i can remove this check.

projects/aca-shared/src/lib/services/app.service.ts
```diff
@@ -166,9 +166,11 @@ export class AppService implements OnDestroy {
...
+          setTimeout(() => {
```
AleksanderSklorz, line 170: We should have tests for bug fixes to prevent the same problem in the future. Can we add some tests?

projects/aca-shared/src/lib/services/app.service.ts
```diff
@@ -166,9 +166,11 @@ export class AppService implements OnDestroy {
...
+            this.openMobileAppDialog();
+          }, 10000);
```
AleksanderSklorz, line 172: Do we really need 10 seconds of delay? We should avoid high delays and check if there is better way. Why you need this? Is that required because you need wait until page is loaded? If yes then maybe you can use following:

https://stackoverflow.com/questions/50345082/angular-4-check-if-all-page-resources-are-fully-loaded-make-loader-spins-unti

(ngAfterViewInit)

@MichalKinas what do you think?

MichalKinas: Yeah I agree, timeout with 10 seconds doesn't seem right and there is a high chance that `ngAfterViewInit` will help

PR #3371, opened 2023-07-26
ADF and JS-API version update

MichalKinas:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [x] Other... Please describe: Dependecy update

ACA dependencies has been updated, all affected E2Es has been fixed. https://alfresco.atlassian.net/browse/ACS-5632

> - [ ] Yes
> - [x] No

**Other information**:


e2e/protractor/protractor.excludes.json
```diff
@@ -6,8 +6,6 @@
   "C286252": "temp, see https://alfresco.atlassian.net/browse/ACS-5189",
   "C284666": "temp, see https://alfresco.atlassian.net/browse/ACS-5189",
   "C286269": "temp, see https://alfresco.atlassian.net/browse/ACS-5189",
-  "C280051": "temp, will be fixed in https://alfresco.atlassian.net/browse/ACS-5632",
-  "C308042": "temp, will be fixed in https://alfresco.atlassian.net/browse/ACS-5632",
   "C279186": "temp, will be fixed in https://alfresco.atlassian.net/browse/ACS-4985",
```
AleksanderSklorz, line 9: @MichalKinas I think there should be removed also some other e2e from excluded list. Actually we removed only ACS-5632 but I see that PR fixes also e2e for ACS-4985 probably?

swapnil-verma-gl: I doubt that the e2e errors for ACS-4985 would be fixed. That's because the old protractor e2es are still relying on the older date UI for the search filter, not the newer one.

On a side note, I thought the plan was to accommodate the new advanced search E2Es when migrating to playwright, and not to fix them in protractor?

AleksanderSklorz: ah ok, I got it. I thought these are fixed because I see some changes for them in that PR. 

Regarding to your question - I think we don't write new e2es in Protractor but if there is quick fix for old existing Protractor tests then we fix them while waiting for migration to Playwright. It is always good to have working E2Es because they prevent bugs so if fix is easy then should be fine to have it fixed to have protection. 

I think the same we did for previous jiras, @MichalKinas can confirm I guess.

AleksanderSklorz (approved):

e2e/protractor/protractor.excludes.json
MichalKinas: I think there is no point in enabling them since I saw that Swapnil is excluding them in his PR for created and modified date with link to Playwright migration, so they will be excluded soon either way

nikita-web-ua (approved):

swapnil-verma-gl (approved):

PR #3467, opened 2023-10-06
[ACS-6078] Fixed layout for create from template component

AleksanderSklorz:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACS-6078

Layout for 'Create from template' component is corrected now. Every field is displayed in separate row.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/projects/aca-content/src/lib/dialogs/node-template/create-from-template.dialog.scss
+++ b/projects/aca-content/src/lib/dialogs/node-template/create-from-template.dialog.scss
@@ -23,6 +23,7 @@
 
   .mat-form-field {
     margin-bottom: 20px;
+    width: 100%;
   }
 
   .mat-dialog-container {
--- a/projects/aca-content/src/lib/dialogs/node-template/create-from-template.dialog.scss
+++ b/projects/aca-content/src/lib/dialogs/node-template/create-from-template.dialog.scss
@@ -21,9 +21,12 @@
     }
   }
 
-  .mat-form-field {
-    margin-bottom: 20px;
+  .app-create-from-template-field {
     width: 100%;
+
+    &:not(:last-of-type) {
+      margin-bottom: 20px;
+    }
   }
 
   .mat-dialog-container {
--- a/projects/aca-content/src/lib/dialogs/node-template/create-from-template.dialog.scss
+++ b/projects/aca-content/src/lib/dialogs/node-template/create-from-template.dialog.scss
@@ -1,7 +1,8 @@
-.aca-create-from-template-dialog {
-  ng-component {
-    overflow: visible;
-  }
+app-create-from-template-dialog {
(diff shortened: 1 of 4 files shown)


MichalKinas (approved):

projects/aca-content/src/lib/dialogs/node-template/create-from-template.dialog.scss
```diff
@@ -23,6 +23,7 @@
 
   .mat-form-field {
     margin-bottom: 20px;
+    width: 100%;
```
nikita-web-ua, line 26: should not we avoid referring to external library classes?

DenysVuika: it's the problem we have with the Angular upgrade, should not be touching internal classes of material

AleksanderSklorz: thanks, done

nikita-web-ua (approved):

projects/aca-content/src/lib/dialogs/node-template/create-from-template.dialog.ts
```diff
@@ -40,6 +40,7 @@ import { MatButtonModule } from '@angular/material/button';
   imports: [CommonModule, TranslateModule, MatDialogModule, ReactiveFormsModule, MatFormFieldModule, MatInputModule, MatButtonModule],
   templateUrl: './create-from-template.dialog.html',
   styleUrls: ['./create-from-template.dialog.scss'],
+  selector: 'app-create-from-template-dialog',
```
swapnil-verma-gl, line 43: Is this selector being used/going to be used somewhere? Also, should we change the selector being used in the [CSS](https://github.com/Alfresco/alfresco-content-app/pull/3467/files#diff-832a272b8c99e0ad8ead7c2fb7b42fbe4e37f0ef1601e47e9e5da23f2fef6ee2L1) for this component? It refers to aca-create-from-template-dialog, rather than app-create-from-template-dialog

AleksanderSklorz: This component is displayed by service as it is dialog. So there is no usage of that selector in component's template directly. But we decided to add this selector because if we skip that then in HTML in browser there is 'ng-component' instead of real selector for that component (it is mentioned in jira). I discussed that with @MichalKinas and we decided that we can add selector as it will make easier finding this component in browser in html when you have real selector instead of just general ng-component. 

I used this selector in css now.

swapnil-verma-gl (approved):

PR #3546, opened 2023-11-29
 manual upstream

DominikIwanek:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:


package.json
```diff
@@ -37,7 +37,7 @@
     "@alfresco/adf-core": ">=6.6.0-7017572520",
     "@alfresco/adf-extensions": ">=6.6.0-7017572520",
     "@alfresco/eslint-plugin-eslint-angular": ">=6.6.0-7017572520",
```
MichalKinas, line 39: What about the adf version? Should we bump it as well?

DominikIwanek: adf has: 
>=6.6.0-7017572520
so I think it's fine?

MichalKinas: That depends on the version you want to use later on in ADW, ideally they should be the same to mitigate the risk of NPM conflicts during the install

MichalKinas: Good practice is to use latest ADF alpha in both ACA and ADW

MichalKinas (approved):

PR #3565, opened 2023-12-18
[ACS-6456] Migrated Share File e2es to Playwright

SheenaMalhotra182:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Protractor is deprecated and e2es for _'share-file.test.ts'_ are written in Protractor.

Migrated the Protractor test cases into Playwright for _'share-file.test.ts'_ file.

> - [ ] Yes
> - [x] No

**Other information**:
https://alfresco.atlassian.net/browse/ACS-6456


e2e/playwright/actions/src/tests/share/share-file.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+  test.afterAll(async () => {
+    await apiClientFactory.nodes.deleteNodes([parentId]);
+  });
+
+  test.describe('when logged out', () => {
+    let file6SharedLink: string;
+    let file6Id: string;
+
+    test.beforeAll(async () => {
+      test.setTimeout(timeouts.extendedTest);
```
akashrathod28, line 65: test SetTimeout will not require if beforeAll require very less time.

projects/aca-playwright-shared/src/page-objects/components/dialogs/share-dialog.component.ts
```diff
@@ -0,0 +1,111 @@
...
+    const toggleClass = await toggle.getAttribute('class');
+    return toggleClass.includes(status);
+  }
+
+  async isShareToggleChecked(): Promise<boolean> {
+    const shareToggleElement = await this.shareToggle.elementHandle();
+    return this.isToggleStatus(shareToggleElement, 'checked');
+  }
+
+  async isShareToggleDisabled(): Promise<boolean> {
```
akashrathod28, line 98: function not used, can be removed.

projects/aca-playwright-shared/src/page-objects/components/dialogs/share-dialog.component.ts
```diff
@@ -0,0 +1,111 @@
...
+  dateTimePicker = new DateTimePicker(this.page);
+
+  getDialogLabel = () => this.getChild('label').innerText();
+  getErrorByText = (text: string): Locator => this.page.locator('mat-error', { hasText: text });
+
+  async getLabels() {
+    return await this.page.$$eval('.adf-share-link__label', (elements) => elements.map((element) => element.textContent));
+  }
+
+  async isErrorMessageDisplayed(errorText: string): Promise<boolean> {
```
akashrathod28, line 57: function not used, can be removed.

projects/aca-playwright-shared/src/api/shared-links-api.ts
```diff
@@ -77,6 +77,19 @@ export class SharedLinksApi {
     }
   }
 
+  async shareNodes(nodeIds: string[], expiresAt?: Date): Promise<any> {
```
akashrathod28, line 80: can you already cleared function , shareFilesByIds

SheenaMalhotra182: removed it, and modified the existing function to be used in tests.

projects/aca-playwright-shared/src/page-objects/components/dialogs/share-dialog.component.ts
SheenaMalhotra182: removed

projects/aca-playwright-shared/src/page-objects/components/dialogs/share-dialog.component.ts
SheenaMalhotra182: removed

e2e/playwright/actions/src/tests/share/share-file.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+  const file7 = `file7-${random}.txt`;
+  const file8 = `file8-${random}.txt`;
+  const file9 = `file9-${random}.txt`;
+
+  const shareLinkPreUrl = `/#/preview/s/`;
+
+  const apiClientFactory = new ApiClientFactory();
+
+  test.beforeAll(async () => {
+    test.setTimeout(timeouts.extendedTest);
```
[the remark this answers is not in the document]
SheenaMalhotra182: removed timeouts where beforeAll is executing properly.

e2e/playwright/actions/src/tests/share/share-file.spec.ts
SheenaMalhotra182: removed timeouts where beforeAll is executing properly

e2e/playwright/actions/src/tests/share/share-file.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+      test('[C286333] Disable the share link expiration', async ({ personalFiles, nodesApiAction, page }) => {
+        expect(await personalFiles.dataTable.performActionFromExpandableMenu(file7, 'Share'));
+
+        expect(await personalFiles.shareDialog.isExpireToggleEnabled()).toBe(true);
+        expect(await personalFiles.shareDialog.getExpireDate()).not.toBe('');
+
+        await personalFiles.shareDialog.expireToggle.click();
+        expect(await personalFiles.shareDialog.isExpireToggleEnabled()).toBe(false);
+
+        page.waitForTimeout(timeouts.tiny);
```
akashrathod28, line 216: await is missing,  Also we can have page from  personalFiles.page

e2e/playwright/actions/src/tests/share/share-file.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+
+        const url = await personalFiles.shareDialog.getLinkUrl();
+        expect(url).toContain(shareLinkPreUrl);
+
+        await personalFiles.shareDialog.urlAction.click();
+
+        const shareSnackBar = personalFiles.snackBar;
+        await expect(shareSnackBar.getByMessageLocator('Link copied to the clipboard')).toBeVisible();
+
+        await page.goto(url);
```
akashrathod28, line 169: can check if able to use  personalFiles.navigate( remoteUrl ) method

e2e/playwright/actions/src/tests/share/share-file.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+        expect(await personalFiles.shareDialog.isExpireToggleEnabled()).toBe(true);
+
+        await personalFiles.shareDialog.datetimePickerButton.click();
+        expect(await personalFiles.shareDialog.dateTimePicker.isCalendarOpen()).toBe(true);
+
+        await personalFiles.shareDialog.dateTimePicker.pickDateTime();
+
+        const inputDate = await personalFiles.shareDialog.getExpireDate();
+
+        await page.waitForTimeout(timeouts.normal);
```
akashrathod28, line 192: can use personalFiles.page

e2e/playwright/actions/src/tests/share/share-file.spec.ts
SheenaMalhotra182: added await. waitForTimeout is an inbuilt method of Page Interface provided by Playwright so I guess we don't need  personalFiles.page

e2e/playwright/actions/src/tests/share/share-file.spec.ts
SheenaMalhotra182: goto is an inbuilt method provided by Playwright and the navigate methods expects options, not the url so this should be fine.

e2e/playwright/actions/src/tests/share/share-file.spec.ts
SheenaMalhotra182: waitForTimeout is an inbuilt method of Page Interface provided by Playwright so I guess we don't need personalFiles.page. Otherwise, we will have to implement wateForTimeout for other pages as well which would be redundant.

e2e/playwright/actions/src/tests/share/share-file.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+      await personalFiles.viewer.toolbar.sharedDownloadButton.click();
+      const download = await downloadPromise;
+      expect(download.suggestedFilename()).toBe(file6);
+    });
+  });
+
+  test.describe('when logged in', () => {
+    const expiryDateObj: Date = new Date();
+    expiryDateObj.setFullYear(expiryDateObj.getFullYear() + 1);
+    const expiryDate: any = expiryDateObj.toISOString().replace('Z', '+0000');
```
MichalKinas, line 92: Please don't use `any`

e2e/playwright/actions/src/tests/share/share-file.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+
+      const downloadPromise = personalFiles.page.waitForEvent('download');
+      await personalFiles.viewer.toolbar.sharedDownloadButton.click();
+      const download = await downloadPromise;
+      expect(download.suggestedFilename()).toBe(file6);
+    });
+  });
+
+  test.describe('when logged in', () => {
+    const expiryDateObj: Date = new Date();
```
MichalKinas, line 90: You don't have to specify a type when you're creating the object right away

projects/aca-playwright-shared/src/page-objects/components/datetime-picker/datetime-picker.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    const element = this.rootElemLocator;
+    return element.isVisible();
+  }
+
+  async pickDateTime(): Promise<void> {
+    const today = new Date();
+    const nextAvailableDay = new Date();
+    nextAvailableDay.setDate(today.getDate() + 2);
+    if (nextAvailableDay.getMonth() !== today.getMonth()) {
+      await await this.nextMonthBtn.click();
```
MichalKinas, line 49: Why double await?

projects/aca-playwright-shared/src/page-objects/components/dialogs/share-dialog.component.ts
```diff
@@ -0,0 +1,101 @@
...
+  expireToggle = this.getChild(`[data-automation-id='adf-expire-toggle']`);
+  expireInput = this.getChild('input[formcontrolname="time"]');
+  datetimePickerButton = this.getChild('.mat-datepicker-toggle');
+
+  dateTimePicker = new DateTimePicker(this.page);
+
+  getDialogLabel = () => this.getChild('label').innerText();
+  getErrorByText = (text: string): Locator => this.page.locator('mat-error', { hasText: text });
+
+  async getLabels() {
```
MichalKinas, line 53: Missing return type

e2e/playwright/actions/src/tests/share/share-file.spec.ts
SheenaMalhotra182: removed type declaration

projects/aca-playwright-shared/src/page-objects/components/datetime-picker/datetime-picker.component.ts
SheenaMalhotra182: fixed typo

projects/aca-playwright-shared/src/page-objects/components/dialogs/share-dialog.component.ts
SheenaMalhotra182: added return type Promise<Array<string>>

e2e/playwright/actions/src/tests/share/share-file.spec.ts
SheenaMalhotra182: removing 'any' here is giving a red line error on _expiryDate_ here 
`await shareApi.shareFilesByIds([file6Id, file7Id], expiryDate);`
because _shareFileByIds()_ and other nested methods expect _expireDate_ to be of type 'Date'. But strangely, the _shareFilesByIds()_ method works only if the _expireDate_ is of type 'string', otherwise the tests don't work as expected.
image


Due to this strange behavior, having 'any' helps but if I remove 'any', I will have to modify the data type of _expireDate_ from 'Date' to 'string' at multiple places and in JS-API which will need upstream.

I have tried multiple ways to make _expiryDate_ an object of type 'Date' so that no changes are required on the API level but that does not seem to work

What do you suggest here? Should we go ahead with 'any'? or can we use some other approach here? @MichalKinas

MichalKinas: If you need to pass a `Date` there you can use `expiryDateObj` object, `expiryDate` should be a string and using `any` here is incorrect

SheenaMalhotra182: I have tried using expiryDateObj but this does not help as 'shareFilesByIds()' is working properly only with expiryDate as string. 
If we remove any, we get an error because 'shareFilesByIds()' expects a Date but is somehow working only with string

MichalKinas: You can try creating new date from `expiryDate` but if it won't work it means that your code is not correct somewhere, `any` is anti-pattern and it should be used only in specifi cases and this is not one of them. `any` shouldn't be used to fix an error or not working code

SheenaMalhotra182: Yes, I understand what you are saying but the problem I am facing in doing so is that
1. the API actually expects 'expiryDate' to be of type string. (see sharedLinkBodyCreate https://api-explorer.alfresco.com/api-explorer/?urls.primaryName=Core%20API#/shared-links/createSharedLink)
image


2. But, in the code, the expected type of 'expiryDate' is set to Date in JS API and all related methods like 'shareNodes()' or 'shareFileByIds()'
image


So even if I create new date from 'expiryDate' , the API will still not run as expected. 
So far, this approach (using 'any') was being used in Protractor e2es so if we remove 'any', we either get some red line errors but the tests run fine 
OR 
we might need to make changes in JS API which I am not sure is feasible.

MichalKinas: In this case you still don't have to use `any`, you can declare `expiryDate` as `let expiryDate: Date | string = ...` and then pass it to the correct function like `expiryDate as Date`

SheenaMalhotra182: Passing expiryDate as Date does not help because `this.apiService.share.createSharedLink(data);` fails if expiry is of type 'Date' and gives status code 400.
It gives 200 only with type string (as per swagger https://api-explorer.alfresco.com/api-explorer/?urls.primaryName=Core%20API#/shared-links/createSharedLink)
and I am blocked because I cannot set the type of expiryDate to string. 
image (3)
image (2)

@DenysVuika, Michal suggested double-checking with you if we should make changes for this in JS-API for e2es. 
Can you please suggest what would be the feasible solution?

SheenaMalhotra182: @MichalKinas, as per my discussion with @DenysVuika, we can use 'any' for now and have a separate ticket to fix typings in JS-API. 

I have created this JIRA https://alfresco.atlassian.net/browse/ACS-6533 for the same.

MichalKinas (approved):

PR #3603, opened 2024-01-17
[ACS-5133] view details button is inactive in expanded view

AleksanderSklorz:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACS-5133

"View Details" button is active in expanded view and after clicking closes view. 

> - [ ] Yes
> - [x] No

**Other information**:


projects/aca-content/src/lib/store/effects/node.effects.ts
```diff
@@ -308,6 +308,12 @@ export class NodeEffects {
       this.actions$.pipe(
         ofType<ExpandInfoDrawerAction>(NodeActionTypes.ExpandInfoDrawer),
         map((action) => {
+          this.router.events
+            .pipe(
+              filter((event) => event instanceof NavigationEnd),
+              first()
```
MichalKinas, line 314: I think you don't have to use filter, in this case you could with something like `first(event => event instanceof NavigationEnd)` which will emit only first value that matches the predicate

projects/aca-content/src/lib/components/details/details.component.ts
```diff
@@ -99,6 +99,22 @@ export class DetailsComponent extends PageComponent implements OnInit, OnDestroy
       .subscribe((aspectActions) => {
         this.aspectActions = aspectActions;
       });
+    this.infoDrawerOpened$
+      .pipe(
+        filter((opened) => !opened),
+        first(),
```
MichalKinas, line 105: Same here, you can use only `first` to filter

projects/aca-content/src/lib/components/details/details.component.ts
```diff
@@ -99,6 +99,22 @@ export class DetailsComponent extends PageComponent implements OnInit, OnDestroy
...
+        takeUntil(
```
MichalKinas, line 106: I think that you don't need `takeUntil` when you're using `first` as after emitting the first value `first` then completes according to rxjs docs

projects/aca-content/src/lib/components/details/details.component.ts
```diff
@@ -99,6 +99,22 @@ export class DetailsComponent extends PageComponent implements OnInit, OnDestroy
...
+          merge(
+            this.onDestroy$,
+            this.router.events.pipe(
+              filter((event) => event instanceof NavigationStart),
```
MichalKinas, line 110: Same points here for `filter` and `takeUntil`

projects/aca-content/src/lib/components/details/details.component.ts
AleksanderSklorz: Yes but what in case if somebody will exit page before any value will be emitted? Then first won't be executed but still we would need to cancel subscription

projects/aca-content/src/lib/components/details/details.component.ts
AleksanderSklorz: same answer like above

projects/aca-content/src/lib/store/effects/node.effects.ts
AleksanderSklorz: good point, done

projects/aca-content/src/lib/components/details/details.component.ts
AleksanderSklorz: done

projects/aca-content/src/lib/components/details/details.component.ts
MichalKinas: Right good point since `infoDrawerOpened$` comes from the store so it won't complete when user will exit the page

MichalKinas (approved):

nikita-web-ua (approved):

swapnil-verma-gl (approved):

g-jaskowski (approved):

PR #3631, opened 2024-02-05
[ACA-4728] fix file order in viewer

g-jaskowski:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Files browsed in viewer with previous/next buttons have different order than in document list.
https://alfresco.atlassian.net/browse/ACA-4728

Getting file order in viewer is set accordingly to sorting mode and unified between pages and viewer.
Last sorting key is saved and used to prevent datatable rows changing order after closing viewer.
Folder structure is changed to reduce code duplication.

> - [x] Yes
> - [ ] No

**Other information**:


MichalKinas (approved):

projects/aca-content/viewer/src/lib/components/viewer/viewer.component.spec.ts
AleksanderSklorz, line 1: @g-jaskowski @MichalKinas what do you think about creating single component instead of two different components (preview and viewer)? 
Both components look almost the same, they have even similar unit tests. It would simplify maintenance in the future because changes would be required only in single component instead of in two. 
This task is just bug fix so probably we would not like to do it here but maybe in separate task, what do you think? Would that be possible and good idea or better keep them as separate components?

AleksanderSklorz (approved):

AleksanderSklorz:
@g-jaskowski before merging that PR please check sonarcloud issues

projects/aca-content/viewer/src/lib/components/viewer/viewer.component.spec.ts
MichalKinas: I think it might not be that easy as preview has another custom use case, when client has a custom plugin e.g. starting a new process he can use preview component to check if file that he wants to attach to the process is correct. We would need to check that use case and investigate if those can be merged then

AleksanderSklorz: Thanks for response @MichalKinas. So should @g-jaskowski raise task to investigate that possibility and eventually merge these 2 components together if possible?

MichalKinas: Yes that's good idea to investigate that separately

AleksanderSklorz: great, then @g-jaskowski could you please raise task for this and put it to your team's board?

DenysVuika (approved):

g-jaskowski:
I added `do not merge` label, as additional fixes are needed

g-jaskowski:
@AleksanderSklorz @MichalKinas I tried to reduce code duplication with the refactor, it's still quite high, but can be lowered with some unit test optimization.
Also another small PR will be needed for ADW, but I wanted to make sure I took the right approach.

projects/aca-content/viewer/src/lib/components/preview/preview.component.ts
```diff
@@ -52,55 +52,31 @@ import { AlfrescoViewerModule, NodesApiService } from '@alfresco/adf-content-ser
   host: { class: 'app-preview' }
 })
 export class PreviewComponent extends PageComponent implements OnInit, OnDestroy {
-  previewLocation: string = null;
-  routesSkipNavigation = ['shared', 'recent-files', 'favorites'];
+  contentExtensions: Array<ViewerExtensionRef> = [];
```
AleksanderSklorz, line 55: is that used somewhere? I can't see usages so probably can be removed

projects/aca-content/viewer/src/lib/components/viewer/viewer.component.spec.ts
```diff
@@ -22,41 +22,332 @@
...
+    });
+    spyOn(component, 'displayNode').and.stub();
+
+    component.ngOnInit();
+
+    expect(component.folderId).toBe('folder1');
+    expect(component.displayNode).toHaveBeenCalledWith('node1');
+  });
+
+  it('should not display node when id is missing', async () => {
```
AleksanderSklorz, line 185: I would change a bit title because this test checks if getNodeInfo and navigate functions are not called. In reality it does not check if node is not displayed because there is no 'expect' which directly checks displaying of node

projects/aca-content/viewer/src/lib/components/viewer/viewer.component.spec.ts
```diff
@@ -22,41 +22,332 @@
...
+
+    expect(store.dispatch).toHaveBeenCalledWith(new ReloadDocumentListAction());
+    expect(component['navigateToFileLocation']).toHaveBeenCalled();
+    expect(router.navigateByUrl).toHaveBeenCalledWith(fakeLocation);
+  });
+
+  it('should navigate to original location in case of Alfresco API errors', async () => {
+    component['previewLocation'] = 'personal-files';
+    spyOn(contentApi, 'getNodeInfo').and.returnValue(throwError('error'));
+    spyOn(JSON, 'parse').and.returnValue({ error: { statusCode: '123' } });
```
AleksanderSklorz, line 225: instead of mocking JSON.parse please try to prepare error object returned by getNodeInfo which gives you expected result in JSON.parse

projects/aca-content/viewer/src/lib/components/viewer/viewer.component.ts
```diff
@@ -74,62 +75,37 @@ export class AcaViewerComponent implements OnInit, OnDestroy {
...
   navigateSource: string = null;
-  previousNodeId: string;
+  navigationSources = ['favorites', 'libraries', 'personal-files', 'recent-files', 'shared'];
   nextNodeId: string;
-  navigateMultiple = true;
+  node: Node;
+  nodeId: string = null;
+  openWith: ContentActionRef[] = [];
+  previousNodeId: string;
   routesSkipNavigation = ['shared', 'recent-files', 'favorites'];
```
AleksanderSklorz, line 87: As you are doing changes here, could you remove that field because it seems unused?

projects/aca-content/viewer/src/lib/services/viewer.service.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * from Hyland Software. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { TestBed } from '@angular/core/testing';
+import { TranslationMock, TranslationService, UserPreferencesService } from '@alfresco/adf-core';
+import { ContentApiService } from '@alfresco/aca-shared';
+import { FavoritePaging, NodePaging, SharedLinkPaging } from '@alfresco/js-api';
+import { ViewerService } from '../services/viewer.service';
```
AleksanderSklorz, line 29: path can be just:  './viewer.service'

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,266 @@
...
+    '-TYPE:"lnk:link"'
+  ];
+
+  /**
+   * Retrieves nearest node information for the given node and folder.
+   *
+   * @param nodeId Unique identifier of the document node
+   * @param folderId Unique identifier of the containing folder node.
+   */
+  async getNearestNodes(nodeId: string, folderId: string, source: string): Promise<{ left: string; right: string }> {
```
AleksanderSklorz, line 66: source parameter seems missing in JSDoc above

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,266 @@
...
+      return empty;
+    }
+  }
+
+  /**
+   * Retrieves a list of node identifiers for the folder and data source.
+   *
+   * @param folderId Optional parameter containing folder node identifier for 'personal-files' and 'libraries' sources.
+   */
+  async getFileIds(source: string, folderId?: string): Promise<string[]> {
```
AleksanderSklorz, line 98: source parameter seems missing in JSDoc above

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,266 @@
...
+  ];
+
+  /**
+   * Retrieves nearest node information for the given node and folder.
+   *
+   * @param nodeId Unique identifier of the document node
+   * @param folderId Unique identifier of the containing folder node.
+   */
+  async getNearestNodes(nodeId: string, folderId: string, source: string): Promise<{ left: string; right: string }> {
+    const empty = {
```
AleksanderSklorz, line 67: can we have a type for this object and then please use this type as return type of this function

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,266 @@
...
+      left: null,
+      right: null
+    };
+
+    if (nodeId && folderId) {
```
AleksanderSklorz, line 72: this code can be a bit shorter: 
```ts
if (nodeId && folderId) {
      try {
        const ids = await this.getFileIds(source, folderId);
        const idx = ids.indexOf(nodeId);

        if (idx >= 0) {
          return {
            left: ids[idx - 1] || null,
            right: ids[idx + 1] || null
          };
        }
      } catch {}
}
return empty;

File: projects/aca-content/viewer/src/lib/services/viewer.service.ts [hunk 10]
```diff
@@ -0,0 +1,266 @@
...
+      const orderBy = isClient ? null : ['isFolder desc', `${sortKey} ${sortDirection}`];
+      const nodes = await this.contentApi
+        .getNodeChildren(folderId, {
+          orderBy: orderBy,
+          fields: this.getFields(sortKey, previousSortKey),
+          where: '(isFile=true)'
+        })
+        .toPromise();
+
+      const entries = nodes.list.entries.map((obj) => obj.entry);
```
AleksanderSklorz, line 112: This part: 
```ts
const entries = nodes.list.entries.map((obj) => obj.entry);
      if (isClient) {
        if (previousSortKey) {
          this.sort(entries, previousSortKey, previousSortDir);
        }
        this.sort(entries, sortKey, sortDirection);
      }
      return entries.map((entry) => entry.id);
```

looks similarly for every 'if'. Can you remove duplications? They are multiple ways to achieve that, for example it can be moved after all 'ifs' at the end of function so will be executed for every case

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,266 @@
...
+    return [];
+  }
+
+  /**
+   * Get the root field name from the property path.
+   * Example: 'property1.some.child.property' => 'property1'
+   *
+   * @param path Property path
+   */
+  getRootField(path: string) {
```
AleksanderSklorz, line 219: missing return type

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,266 @@
...
+    if (path) {
+      return path.split('.')[0];
+    }
+    return path;
+  }
+
+  private sort(items: any[], key: string, direction: string) {
```
AleksanderSklorz, line 226: can we use some type instead of any?

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,266 @@
...
+    const options: Intl.CollatorOptions = {};
+    if (key.includes('sizeInBytes') || key === 'name') {
+      options.numeric = true;
+    }
+
+    items.sort((a: any, b: any) => {
```
AleksanderSklorz, line 232: same here, can we remove any?

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,266 @@
...
+      let right = ObjectUtils.getValue(b, key) ?? '';
+      right = right instanceof Date ? right.valueOf().toString() : right.toString();
+
+      return direction === 'asc' || direction === 'ASC'
+        ? left.localeCompare(right, undefined, options)
+        : right.localeCompare(left, undefined, options);
+    });
+  }
+
+  private getFields(sortKey: string, previousSortKey?: string) {
```
AleksanderSklorz, line 245: missing return type

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,266 @@
...
+    return ['id', this.getRootField(sortKey), this.getRootField(previousSortKey)];
+  }
+
+  private getSortKeyDir(source: string) {
```
AleksanderSklorz, line 249: missing return type

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,266 @@
...
+    const previousSortKey = this.preferences.get(`${source}.sorting.previousKey`);
+    const previousSortDir = this.preferences.get(`${source}.sorting.previousDirection`);
+    const sortKey = this.preferences.get(`${source}.sorting.key`) || this.getDefaults(source)[0];
+    const sortDirection = this.preferences.get(`${source}.sorting.direction`) || this.getDefaults(source)[1];
+    return [sortKey, sortDirection, previousSortKey, previousSortDir];
+  }
+
+  private getDefaults(source: string) {
```
AleksanderSklorz, line 257: missing return type

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,229 @@
...
+    '-TYPE:"ia:calendarEvent"',
+    '-TYPE:"lnk:link"'
+  ];
+
+  /**
+   * Retrieves nearest node information for the given node and folder.
+   *
+   * @param nodeId Unique identifier of the document node
+   * @param folderId Unique identifier of the containing folder node.
+   * @param source Data source name. Allowed values are: personal-files, libraries, favorites, shared, recent-files.
```
AleksanderSklorz, line 70: if allowed values for source are 

personal-files, libraries, favorites, shared, recent-files

then maybe we can limit type of source parameter to only these values instead of general string which allows every string?

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,229 @@
...
+    return empty;
+  }
+
+  /**
+   * Retrieves a list of node identifiers for the folder and data source.
+   *
+   * @param source Data source name. Allowed values are: personal-files, libraries, favorites, shared, recent-files.
+   * @param folderId Optional parameter containing folder node identifier for 'personal-files' and 'libraries' sources.
+   */
+  async getFileIds(source: string, folderId?: string): Promise<string[]> {
```
AleksanderSklorz, line 100: similarly like above - maybe we can limit type of source to only these allowed values instead of general string?

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,229 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * from Hyland Software. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { ObjectUtils, UserPreferencesService } from '@alfresco/adf-core';
+import { FavoritePaging, Node, NodePaging, SearchRequest, ResultSetPaging, SharedLink, SharedLinkPaging } from '@alfresco/js-api';
+import { Injectable } from '@angular/core';
+import { ContentApiService } from '@alfresco/aca-shared';
+
+interface adjacentFiles {
```
AleksanderSklorz, line 30: type should start from upper case

projects/aca-content/viewer/src/lib/services/viewer.service.ts
g-jaskowski: @AleksanderSklorz I indiscriminately reverted to old param description. Now the description is corrected, so is the getFileIds function with 'wrong' source. Your solution would be cleaner, but the source is tied to pages, and due to libraries pages' logic I believe this solution to be sufficient.

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,229 @@
...
+   * @param path Property path
+   */
+  getRootField(path: string): string {
+    if (path) {
+      return path.split('.')[0];
+    }
+    return path;
+  }
+
+  private sort(items: Node[] | SharedLink[], key: string, direction: string) {
```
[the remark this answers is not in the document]
g-jaskowski: Throughout whole ACA 'asc' is used, 'ASC' version occurred in one template and caused small bugs to happen with sorting indicator in column header. The direction is used in sorting: string[] property, so I left it without a type, as "...==='asc'"

projects/aca-content/viewer/src/lib/services/viewer.service.ts
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    entries = nodes.list.entries.map((obj) => obj.entry.target?.file ?? obj.entry);
+    if (isClient) {
+      if (previousSortKey) {
+        this.sort(entries, previousSortKey, previousSortDir);
+      }
+      this.sort(entries, sortKey, sortDirection);
+    }
+    return nodes === undefined ? [] : entries.map((entry) => entry.id ?? entry.nodeId);
```
AleksanderSklorz, line 175: No need to check if nodes is undefined because of multiple reasons: 
1. There is Intellij hint which says that it will give always false (because nodes will be never undefined): 
image

2. It can't be undefined because source can have following values: personal-files, libraries, favorites, shared and recent-files and you have 'if' for each of these values and in every 'if' you set nodes variable. 

3. You still refer to field of nodes variable (nodes.list) in 168th line so that means it is never undefined as you never get issue in that line that you refer to field of undefined. 

Hence please remove that condition which checks if nodes is undefined.

AleksanderSklorz (approved):

DenysVuika (approved):

PR #3637, opened 2024-02-12
[ACS-6427] Add search highlighting

MichalKinas:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

Nothing is highlighted in search results.

Search phrase is highlighted in node name, title, description and content. Description is now part of name column. Highlighting config is added to each existing filters sets. https://alfresco.atlassian.net/browse/ACS-6427
Zrzut ekranu 2024-02-12 o 08 21 56

> - [ ] Yes
> - [x] No

**Other information**:


projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
```diff
@@ -86,13 +94,42 @@ export class SearchResultsRowComponent implements OnInit, OnDestroy {
     this.node = this.context.row.node;
     this.isFile = this.node.entry.isFile;
 
-    const { name, properties } = this.node.entry;
-    const title = properties ? properties['cm:title'] : '';
-
+    const highlights: SearchEntryHighlight[] = this.node.entry['search']?.['highlight'];
```
AleksanderSklorz, line 97: Maybe it would be good idea to update model to have search and highlight fields defined inside models? Thanks that we won't need to do some workarounds with [] and IDE will hint available fields. 

Not sure if that is easy to update and if that is possible so if not in that PR we can also raise task for that. What do you think?

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
```diff
@@ -86,13 +94,42 @@ export class SearchResultsRowComponent implements OnInit, OnDestroy {
...
+    let name = this.node.entry.name;
+    const properties = this.node.entry.properties;
+    let title = properties ? properties['cm:title'] : '';
```
AleksanderSklorz, line 100: can be just: 
let title =  properties?.['cm:title'] || '';

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
```diff
@@ -114,4 +151,10 @@ export class SearchResultsRowComponent implements OnInit, OnDestroy {
     event.stopPropagation();
     this.store.dispatch(new NavigateToFolder(this.node));
   }
+
+  private stripHighlighting(highlightedContent: string): string {
+    return highlightedContent && highlightedContent !== ''
```
AleksanderSklorz, line 156: '' is falsy so you don't need second condition (highlightedContent !== '' -> this can be removed) because first condition will allow for truthy only

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
```diff
@@ -114,4 +151,10 @@ export class SearchResultsRowComponent implements OnInit, OnDestroy {
...
+      ? highlightedContent.replace(new RegExp(this.highlightPrefix, 'g'), '').replace(new RegExp(this.highlightPostfix, 'g'), '')
```
AleksanderSklorz, line 157: I'm wondering - instead of two calls of replace function and two regexps can we use OR inside regexp so then it will match both - prefix and postfix at once? Is that possible?

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
```diff
@@ -46,6 +46,8 @@ import { MatDialogModule } from '@angular/material/dialog';
   host: { class: 'aca-search-results-row' }
 })
 export class SearchResultsRowComponent implements OnInit, OnDestroy {
+  private readonly highlightPrefix = "<span class='aca-highlight'>";
```
AleksanderSklorz, line 49: maybe we should keep private readonly separately than private (so just have empty line between them)

projects/aca-content/assets/app.extensions.json
```diff
@@ -1532,7 +1532,27 @@
               "visible": "app.areCategoriesEnabled"
             }
           }
-        ]
+        ],
+        "highlight": {
+          "prefix": "<span class='aca-highlight'>",
```
DenysVuika, line 1537: maybe worth having some sort of `template` property that includes both pre- and post-? i.e.

```json
{
  "template": "<span>{{value}}</span>"
}
```

where you substitute `{{value}}` with the highlighted term

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
MichalKinas: Basically `ResultSetRowEntry` contains search properties but the component itself needs regular node to pass it to subsequent methods accepting node only, and node itself shouldn't contain the search properties imo @DenysVuika what do you think?

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
MichalKinas: I tired but for this to work we would need to escape every special character in the prefix and postfix, and the plan is to start taking them from config file which won't be that easy to implement, we would have to write another function to escape each special character and perform another replace

projects/aca-content/assets/app.extensions.json
MichalKinas: How would that work? BE expects us to provide prefix and postfix separately to properly highlight the search term, and what then is passed inside prefix and postfix is decided by search API

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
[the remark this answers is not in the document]
MichalKinas: We could but then the question comes which filter set we should base on, also there might be new ones added by the customers so this wouldn't be very reliable solution as well

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
MichalKinas: Fixed

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
MichalKinas: Right, fixed

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
MichalKinas: Done

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
AleksanderSklorz: ok, then you can leave as it is, thanks for explanation

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
AleksanderSklorz: @MichalKinas small thing but you missed this for description too

AleksanderSklorz (approved):

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
MichalKinas: Ah thanks, will update it in next commit

nikita-web-ua (approved):

projects/aca-content/assets/app.extensions.json
DenysVuika: ah, ok, no worries then, I thought it's combined in the FE

DenysVuika (approved):

PR #3700, opened 2024-03-11
[ACS-6954] Refactored search playwright tests

katarzynakita:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/projects/aca-playwright-shared/src/page-objects/pages/search.page.ts
+++ b/projects/aca-playwright-shared/src/page-objects/pages/search.page.ts
@@ -28,6 +28,8 @@ import { DataTableComponent, MatMenuComponent, ViewerComponent, SearchInputCompo
 import { AcaHeader } from '../components/aca-header.component';
 import { AdfConfirmDialogComponent, AdfFolderDialogComponent } from '../components/dialogs';
 
+type SearchType = 'files' | 'folders' | 'filesAndFolders' | 'libraries';
+
 export class SearchPage extends BasePage {
   private static pageUrl = 'search';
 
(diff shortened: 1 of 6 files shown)


projects/aca-playwright-shared/src/page-objects/pages/search.page.ts
```diff
@@ -44,4 +44,36 @@ export class SearchPage extends BasePage {
...
+    await this.searchInput.searchButton.click();
+    await this.searchInput.checkOnlyFolders();
+    await this.searchInput.searchFor(folderName);
+    await this.dataTable.progressBarWaitForReload();
+  }
+
+  async searchLibraries(libraryName: string): Promise<void> {
+    await this.acaHeader.searchButton.click();
+    await this.searchInput.searchButton.click();
+    await this.searchInput.checkLibraries();
```
MichalKinas, line 75: Basically each of those function is exactly the same except this line, I think we can make one function out of that with additional param which might be something like `searchType` which can accept values like `files | folders | libraries | filesAndFolders` and then you can use `switch case` to properly call correct action based on  `searchType` value

MichalKinas (changes requested):

projects/aca-playwright-shared/src/page-objects/pages/search.page.ts
katarzynakita: Done

MichalKinas (approved):

AleksanderSklorz (approved):

AleksanderSklorz (approved):

PR #3798, opened 2024-04-17
[ACS-7337] [E2E] protractor to playwright e2e test suites search filters categories filter

datguychen:

> - [x] Docs have been added / updated (for bug fixes / features)

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [x] Other... Please describe:

https://hyland.atlassian.net/browse/ACS-7337

protractor to playwright e2e test suites search filters categories filter

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
+++ b/e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
@@ -1,0 +1,80 @@
+/*!
+ * Copyright © 2005-2024 Hyland Software, Inc. and its affiliates. All rights reserved.
+ *
+ * Alfresco Example Content Application
+ *
+ * This file is part of the Alfresco Example Content Application.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail. Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * The Alfresco Example Content Application is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * The Alfresco Example Content Application is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * from Hyland Software. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { expect } from '@playwright/test';
+import { ApiClientFactory, Utils, test, NodesApi, TrashcanApi, FileActionsApi, TEST_FILES, CategoriesApi } from '@alfresco/playwright-shared';
+import { CategoryLinkBody } from '@alfresco/js-api';
+
+test.describe.only('Search - Filters - Categories', () => {
+  let nodesApi: NodesApi;
+  let trashcanApi: TrashcanApi;
+  let categoriesApi: CategoriesApi;
+  let categoryData;
+
+  const randomId = Utils.random();
+  const username = `user-${randomId}`;
+  const fileNamePdf = `${randomId}-fileNameKb.pdf`;
+  const fileNameJpg = `${randomId}-fileNameMb.jpg`;
+
+  const newSubcategories = [{ name: 'e2e' }];
+
+  test.beforeEach(async ({ loginPage }) => {
+    await Utils.tryLoginUser(loginPage, username, username, 'beforeEach failed');
+  });
+
+  test.beforeAll(async () => {
+    try {
+      const apiClientFactory = new ApiClientFactory();
+      await apiClientFactory.setUpAcaBackend('admin');
+      await apiClientFactory.createUser({ username });
+      nodesApi = await NodesApi.initialize(username, username);
+      trashcanApi = await TrashcanApi.initialize(username, username);
+      categoriesApi = await CategoriesApi.initialize('admin');
+      const fileActionsApi = await FileActionsApi.initialize(username, username);
+      await fileActionsApi.uploadFileWithRename(TEST_FILES.PDF.path, fileNamePdf, '-my-');
+      const jpgFileId = (await fileActionsApi.uploadFileWithRename(TEST_FILES.JPG_FILE.path, fileNameJpg, '-my-')).entry.id;
+      categoryData = await categoriesApi.createCategory('-root-', newSubcategories);
+      const categoryLinkBodyCreate: CategoryLinkBody[] = [{ categoryId: categoryData.entry.id }];
+      await categoriesApi.linkNodeToCategory(jpgFileId, categoryLinkBodyCreate);
+    } catch (error) {
+      console.error(`beforeAll failed: ${error}`);
+    }
+  });
+
+  test.afterAll(async () => {
+    await Utils.deleteNodesSitesEmptyTrashcan(nodesApi, trashcanApi, 'afterAll failed');
+    await categoriesApi.deleteCategory(categoryData.entry.id);
+  });
+
+  test('[C699498] Filter by categories', async ({ searchPage }) => {
+    await searchPage.searchWithin(randomId, 'files');
+    await searchPage.searchFilters.categoriesFilter.click();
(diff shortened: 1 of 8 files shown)


projects/aca-playwright-shared/src/api/categories-api.ts
```diff
@@ -0,0 +1,70 @@
...
+      console.error(`${this.constructor.name} ${this.deleteCategory.name}: ${error}`);
+    }
+  }
+
+  async linkNodeToCategory(nodeId: string, categoryLinkBodyCreate: CategoryLinkBody[], opts?: CategoryQuery): Promise<CategoryPaging | CategoryEntry> {
+    try {
+      return await this.apiService.categoriesApi.linkNodeToCategory(nodeId, categoryLinkBodyCreate, opts);
+    } catch (error) {
+      console.error(`${this.constructor.name} ${this.linkNodeToCategory.name}: ${error}`);
+      throw error;
```
akashrathod28, line 67: do we need to throw this error again? I think we use this on Major logic error like login.

datguychen: No we dont need it, thanks for catching it

akashrathod28 (approved):

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
```diff
@@ -0,0 +1,86 @@
...
+      await apiClientFactory.setUpAcaBackend('admin');
+      await apiClientFactory.createUser({ username });
+      nodesApi = await NodesApi.initialize(username, username);
+      trashcanApi = await TrashcanApi.initialize(username, username);
+      categoriesApi = await CategoriesApi.initialize('admin');
+      const fileActionsApi = await FileActionsApi.initialize(username, username);
+      await fileActionsApi.uploadFileWithRename(TEST_FILES.PDF.path, fileNamePdf, '-my-');
+      const jpgFileId = (await fileActionsApi.uploadFileWithRename(TEST_FILES.JPG_FILE.path, fileNameJpg, '-my-')).entry.id;
+      categoryData = await categoriesApi.createCategory('-root-', newSubcategories);
+      if (categoryData && 'entry' in categoryData) {
```
MichalKinas, line 59: What are you trying to check here? If you need to check the type you can use `instanceof` or if you want to check if `categoryData` has `entry` property you can use `categoryData?.entry`

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
```diff
@@ -0,0 +1,86 @@
...
+        categoryId = categoryData.entry.id;
+        const categoryLinkBodyCreate: CategoryLinkBody[] = [{ categoryId: categoryData.entry.id }];
+        await categoriesApi.linkNodeToCategory(jpgFileId, categoryLinkBodyCreate);
+      } else {
+        categoryId = null;
```
MichalKinas, line 64: Is this correct? It will call the delete in afterAll with null

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
datguychen: This is the only way I came up with that let me use ".entry.id" on "categoryData". The main issue here is that "categoryData" has the "CategoryPaging | CategoryEntry" type, and only "CategoryEntry" lets the use of ".entry.id".

I tried many different iterations to make it look nice, but this is the only one that works and keeps the "CategoryPaging | CategoryEntry" type.

I tried using "categoryData?.entry" but the "CategoryPaging" does let that.

Using "instanceof" doesnt change much so I am not sure here.

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
datguychen: I can either delete the else, or leave it and add another "if not null" in afterAll like this:
"    if (categoryId !== null) {
      await categoriesApi.deleteCategory(categoryId);
    }"
    
   What do you think?

akashrathod28: removing this line ' categoryId = null' , and you can add null check in deleteCategory method before you try to delete category

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
akashrathod28: I think we don't need this if check, createCategory return entry or it will fail. this we are already catch createCategory function.

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
MichalKinas: Yeah I agree with Akash

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
MichalKinas: Using `instanceof` verifies if passed object has correct type hence has correct properties so in this case I think the correct option would be to use `categoryData != null && categoryData instanceof CategoryEntry`

datguychen: Without this if I get this error:
"Property 'entry' does not exist on type 'CategoryPaging | CategoryEntry'.
  Property 'entry' does not exist on type 'CategoryPaging'."

MichalKinas: If is required in this case, like I said you need to use this condition: `categoryData != null && categoryData instanceof CategoryEntry`

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
datguychen: Done, but I am not sure if correctly

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
```diff
@@ -0,0 +1,84 @@
...
+        await categoriesApi.linkNodeToCategory(jpgFileId, categoryLinkBodyCreate);
+      }
+    } catch (error) {
+      console.error(`beforeAll failed: ${error}`);
+    }
+  });
+
+  test.afterAll(async () => {
+    await Utils.deleteNodesSitesEmptyTrashcan(nodesApi, trashcanApi, 'afterAll failed');
+    await categoriesApi.deleteCategory(categoryId);
```
MichalKinas, line 71: You need to add a if here to check if `categoryId` is valid before trying to remove it

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
```diff
@@ -0,0 +1,84 @@
...
+      await apiClientFactory.setUpAcaBackend('admin');
+      await apiClientFactory.createUser({ username });
+      nodesApi = await NodesApi.initialize(username, username);
+      trashcanApi = await TrashcanApi.initialize(username, username);
+      categoriesApi = await CategoriesApi.initialize('admin');
+      const fileActionsApi = await FileActionsApi.initialize(username, username);
+      await fileActionsApi.uploadFileWithRename(TEST_FILES.PDF.path, fileNamePdf, '-my-');
+      const jpgFileId = (await fileActionsApi.uploadFileWithRename(TEST_FILES.JPG_FILE.path, fileNameJpg, '-my-')).entry.id;
+      categoryData = await categoriesApi.createCategory('-root-', newSubcategories);
+      if (categoryData instanceof CategoryEntry && 'entry' in categoryData) {
```
MichalKinas, line 59:
```suggestion
      if (categoryData instanceof CategoryEntry) {
```
`entry` is not optional in `CategoryEntry` it will always be there if category is created

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
datguychen: I updated with `categoryData != null && categoryData instanceof CategoryEntry`.

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
datguychen: I added the null check in the deleteCategory method

MichalKinas (approved):

e2e/playwright/search/src/tests/search-filters-categories.e2e.ts
datguychen: correction, I updated with `if (categoryData instanceof CategoryEntry) {`

PR #3971, opened 2024-07-25
[ACS-8326] enable running actions from bulk dropdown and resetting selection

g-jaskowski:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8326

Actions are run on bulk dropdown selection.
There is a possibility to reset bulk dropdown selection.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/projects/aca-content/src/lib/components/bulk-actions-dropdown/bulk-actions-dropdown.component.ts
+++ b/projects/aca-content/src/lib/components/bulk-actions-dropdown/bulk-actions-dropdown.component.ts
@@ -25,8 +25,8 @@
 import { ContentActionRef } from '@alfresco/adf-extensions';
 import { AppStore, getSearchItemsTotalCount } from '@alfresco/aca-shared/store';
 import { CommonModule } from '@angular/common';
-import { Component, Input, OnDestroy, OnInit, ViewChild, ViewEncapsulation } from '@angular/core';
-import { MatSelect, MatSelectModule } from '@angular/material/select';
+import { Component, Input, OnDestroy, OnInit, ViewEncapsulation } from '@angular/core';
+import { MatSelectModule } from '@angular/material/select';
 import { Store } from '@ngrx/store';
 import { TranslateModule } from '@ngx-translate/core';
 import { combineLatest, Observable, Subject } from 'rxjs';
@@ -46,11 +46,9 @@ import { AppExtensionService } from '@alfresco/aca-shared';
 export class BulkActionsDropdownComponent implements OnInit, OnDestroy {
   @Input() items: ContentActionRef[];
 
-  @ViewChild(MatSelect) select: MatSelect;
-
   placeholder: string;
   tooltip: string;
-  disableControl = new FormControl();
+  bulkSelectControl = new FormControl();
 
   private readonly totalItems$: Observable<number> = this.store.select(getSearchItemsTotalCount);
   private readonly onDestroy$ = new Subject();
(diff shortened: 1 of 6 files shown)


projects/aca-content/src/lib/components/bulk-actions-dropdown/bulk-actions-dropdown.component.ts
```diff
@@ -80,10 +83,20 @@ export class BulkActionsDropdownComponent implements OnInit, OnDestroy {
         this.tooltip = title;
         this.placeholder = placeholder;
       });
+
+    this.extensions.resetBulkActionsSubject$.pipe(takeUntil(this.onDestroy$)).subscribe(() => {
```
swapnil-verma-gl, line 87: Instead of injecting MatSelect using ViewChild, you can also use the disableControl form control, and set the value through that. Also, while you are at it, maybe rename the property to be something other than disableControl? Maybe something like bulkSelectControl

projects/aca-shared/src/lib/services/app.extension.service.ts
```diff
@@ -571,4 +572,8 @@ export class AppExtensionService implements RuleContext {
 
     return true;
   }
+
+  resetBulkActions(): void {
```
MichalKinas, line 576: Do we want it this to be exposed to extension service? I think selected action can be reset upon executing an action in dropdown component

projects/aca-content/src/lib/components/bulk-actions-dropdown/bulk-actions-dropdown.component.ts
DaryaBalvanovich, line 87: could you please avoid adding `Subject` to the end of property, `$` is enough

g-jaskowski: Done, thanks!

projects/aca-shared/src/lib/services/app.extension.service.ts
g-jaskowski: @MichalKinas The Bulk Actions Dropdown is visible above the dialogs and resetting it after executing an action from the component would make the selection not visible at all. That's the only reason for using the Extension Service, because it allows resetting selection on dialog's `afterClosed`. Is it okay in this case or should I look for another solution, possibly through another actions executed from ADW level?

projects/aca-content/src/lib/components/bulk-actions-dropdown/bulk-actions-dropdown.component.ts
g-jaskowski: Fixed, thanks!

swapnil-verma-gl (approved):

DaryaBalvanovich (approved):

projects/aca-shared/src/lib/services/app.extension.service.ts
MichalKinas: Okay if the design requires it then it is fine, I would maybe rename it a little bit since now it's not clear why extension service would want to reset bulk actions and it creates a tight-coupling with the bulk dropdown component so instead let's rename it to `bulkActionExecuted` subject which will be a generic subject not coupled directly to any component and then the dropdown component can just reset whenever this event will happen, later on this subject can be even reused by different components if they will need to react on bulk action being executed, does this approach sound good to you?

MichalKinas (approved):

PR #4030, opened 2024-08-13
[ACS-8588] Navigation is triggered twice when leaving Knowledge Retrieval page

jacekpluta:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8588

4 issues described in the ticket has been fixed

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input/search-ai-input.component.spec.ts
+++ b/projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input/search-ai-input.component.spec.ts
@@ -33,7 +33,7 @@ import { of, Subject } from 'rxjs';
 import { AgentWithAvatar, NodeEntry } from '@alfresco/js-api';
 import { FormControlDirective } from '@angular/forms';
 import { DebugElement } from '@angular/core';
-import { AvatarComponent, IconComponent, NotificationService, UserPreferencesService } from '@alfresco/adf-core';
+import { AvatarComponent, IconComponent, NotificationService, UnsavedChangesDialogComponent, UserPreferencesService } from '@alfresco/adf-core';
 import { HarnessLoader } from '@angular/cdk/testing';
 import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
 import { MatSelectHarness } from '@angular/material/select/testing';
@@ -43,7 +43,7 @@ import { MatButton } from '@angular/material/button';
 import { MatInputHarness } from '@angular/material/input/testing';
 import { SelectionState } from '@alfresco/adf-extensions';
 import { MatSnackBarRef } from '@angular/material/snack-bar';
-import { MatDialog, MatDialogRef } from '@angular/material/dialog';
+import { MatDialog, MatDialogConfig, MatDialogRef } from '@angular/material/dialog';
 import { ActivatedRoute } from '@angular/router';
 import { ModalAiService } from '../../../../services/modal-ai.service';
 
@@ -258,13 +258,13 @@ describe('SearchAiInputComponent', () => {
       let queryInput: MatInputHarness;
       let submittingTrigger: () => void;
       const query = 'some query';
-      let dialogOpenSpy: jasmine.Spy<(component: any, config?: any) => MatDialogRef<any, any>>;
-      let modalAi: ModalAiService;
+      let dialogOpenSpy: jasmine.Spy<<T, R>(component: typeof UnsavedChangesDialogComponent, config?: MatDialogConfig) => MatDialogRef<T, R>>;
+      let modalAiService: ModalAiService;
 
       beforeEach(async () => {
         prepareBeforeTest();
 
-        modalAi = TestBed.inject(ModalAiService);
+        modalAiService = TestBed.inject(ModalAiService);
         checkSearchAvailabilitySpy = spyOn(TestBed.inject(SearchAiService), 'checkSearchAvailability');
         notificationService = TestBed.inject(NotificationService);
         userPreferencesService = TestBed.inject(UserPreferencesService);
@@ -400,13 +400,15 @@ describe('SearchAiInputComponent', () => {
         expect(dialogOpenSpy).toHaveBeenCalled();
       });
 
-      it('should open Unsaved Changes Modal', () => {
-        const modalAiSpy = spyOn(modalAi, 'openUnsavedChangesModal').and.callFake((callback) => callback());
+      it('should open Unsaved Changes Modal and run callback successfully', () => {
+        const modalAiSpy = spyOn(modalAiService, 'openUnsavedChangesModal').and.callThrough();
+        spyOn(component.searchSubmitted, 'emit');
 
         fixture.detectChanges();
 
         submittingTrigger();
         expect(modalAiSpy).toHaveBeenCalledWith(jasmine.any(Function));
+        expect(component.searchSubmitted.emit).toHaveBeenCalled();
       });
     });
   }
--- a/projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-results/search-ai-results.component.spec.ts
+++ b/projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-results/search-ai-results.component.spec.ts
@@ -31,9 +31,22 @@ import { UserPreferencesService } from '@alfresco/adf-core';
 import { MatDialogModule } from '@angular/material/dialog';
 import { AppTestingModule } from '../../../../testing/app-testing.module';
 import { MatIconTestingModule } from '@angular/material/icon/testing';
-import { SearchAiService } from '@alfresco/adf-content-services';
+import { NodesApiService, SearchAiService } from '@alfresco/adf-content-services';
 import { By } from '@angular/platform-browser';
 import { ModalAiService } from '../../../../services/modal-ai.service';
+import { delay } from 'rxjs/operators';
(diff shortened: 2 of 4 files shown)


projects/aca-content/assets/app.extensions.json
```diff
@@ -212,6 +212,9 @@
             "order": 100,
             "title": "APP.BROWSE.FILE.SIDENAV_LINK.LABEL",
             "description": "APP.BROWSE.FILE.SIDENAV_LINK.TOOLTIP",
+            "data": {
+              "canBeInactive": true
```
AleksanderSklorz, line 216: Did you check if that still works in ADW?

projects/aca-content/src/lib/components/sidenav/directives/expansion-panel.directive.spec.ts
```diff
@@ -118,5 +118,25 @@ describe('AcaExpansionPanel', () => {
 
       expect(router.navigate).not.toHaveBeenCalled();
     });
+
+    it('should not navigate to first child if none is active route and acaExpansionPanel has canBeInactive property', () => {
+      const router: any = new RouterStub('dummy-route-2');
```
AleksanderSklorz, line 123: please remove any

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input/search-ai-input.component.spec.ts
```diff
@@ -396,6 +399,15 @@ describe('SearchAiInputComponent', () => {
 
         expect(dialogOpenSpy).toHaveBeenCalled();
       });
+
+      it('should open Unsaved Changes Modal', () => {
+        const modalAiSpy = spyOn(modalAi, 'openUnsavedChangesModal').and.callFake((callback) => callback());
```
AleksanderSklorz, line 404: We should test what callback for openUnsavedChangesModal does. Actually we only check if that callback was called but we don't know what happens after that so we don't know if functionality works after calling callback.

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-results/search-ai-results.component.spec.ts
```diff
@@ -179,4 +149,22 @@ describe('SearchAiResultsComponent', () => {
...
+  describe('Unsaved Changes Modal', () => {
+    let modalAi: ModalAiService;
+
+    beforeEach(() => {
+      modalAi = TestBed.inject(ModalAiService);
+      spyOn(userPreferencesService, 'get').and.returnValue(knowledgeRetrievalNodes);
+    });
+
+    it('should open Unsaved Changes Modal', () => {
+      const modalAiSpy = spyOn(modalAi, 'openUnsavedChangesModal').and.callFake((callback) => callback());
```
AleksanderSklorz, line 162: same like for previous testing file - we need to test what callback does because we need to ensure that logic inside callback works

projects/aca-content/src/lib/services/modal-ai.service.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+import { ContentTestingModule } from '@alfresco/adf-content-services';
+import { MatDialog, MatDialogModule, MatDialogRef } from '@angular/material/dialog';
+import { ActivatedRoute, Params } from '@angular/router';
+import { of, Subject } from 'rxjs';
+import { StorageService, UnsavedChangesDialogComponent } from '@alfresco/adf-core';
+
+describe('ModalAiService', () => {
+  const mockQueryParams = new Subject<Params>();
+  let service: ModalAiService;
+  let dialogOpenSpy: jasmine.Spy<(component: any, config?: any) => MatDialogRef<any>>;
```
AleksanderSklorz, line 36: We should avoid any. You can change it to: 
let dialogOpenSpy: jasmine.Spy<<T, D, R>(component: ComponentType<T>, config?: MatDialogConfig<D>) => MatDialogRef<T, R>>;

or any other if it avoids any.

projects/aca-content/src/lib/services/modal-ai.service.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+import { TestBed } from '@angular/core/testing';
+import { ContentTestingModule } from '@alfresco/adf-content-services';
+import { MatDialog, MatDialogModule, MatDialogRef } from '@angular/material/dialog';
+import { ActivatedRoute, Params } from '@angular/router';
+import { of, Subject } from 'rxjs';
+import { StorageService, UnsavedChangesDialogComponent } from '@alfresco/adf-core';
+
+describe('ModalAiService', () => {
+  const mockQueryParams = new Subject<Params>();
+  let service: ModalAiService;
```
AleksanderSklorz, line 35: please add empty line to separate const and let variables

projects/aca-content/src/lib/services/modal-ai.service.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+          useValue: {
+            getItem: () => storageGetItem
+          }
+        },
+        {
+          provide: ActivatedRoute,
+          useValue: {
+            queryParams: mockQueryParams.asObservable(),
+            snapshot: {
+              queryParams: { query: query }
```
AleksanderSklorz, line 54: can be just: 
queryParams: { query }

projects/aca-content/src/lib/services/modal-ai.service.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    });
+
+    dialog = TestBed.inject(MatDialog);
+    dialogOpenSpy = spyOn(dialog, 'open').and.returnValue({
+      afterClosed: () => of(true)
+    } as MatDialogRef<any>);
+    service = TestBed.inject(ModalAiService);
+  };
+
+  describe('when there is UNSAVED_CHANGES_MODAL_HIDDEN in storage', () => {
```
AleksanderSklorz, line 68: this describe has the same title like describe below, please change one of them

projects/aca-content/src/lib/services/modal-ai.service.spec.ts
AleksanderSklorz, line 1: When I run these tests I get following error: 
image

"Some of your tests did a full page reload!"

Could you please have a look?

AleksanderSklorz: I have one observation here - I'm getting this error for each of these files so please check if that is caused by our changes. If not then you can skip that comment

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-results/search-ai-results.component.ts
AleksanderSklorz, line 1: I see each test causes some backend call, looks like some service is not spied (mocked). Could you please have a look?
image

projects/aca-content/src/lib/services/modal-ai.service.ts
```diff
@@ -0,0 +1,62 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class ModalAiService {
+  private route = inject(ActivatedRoute);
+  private dialog = inject(MatDialog);
+  private storageService = inject(StorageService);
+
+  openUnsavedChangesModal(callback: () => void): void {
+    const hasPreviousSearch = this.route.snapshot?.queryParams?.query?.length > 0;
+    const modalHidden = this.storageService.getItem(AppConfigValues.UNSAVED_CHANGES_MODAL_HIDDEN) === 'true';
```
AleksanderSklorz, line 38: I see one bug. Please look at below screenshot and description below that: 
image

If you check checkbox then there is added following property to storage: 
unsaved_changes__modal_hidden: true 

There is one big problem with that: please look at rest of properties in storage, some of them are prefixed by username if users may want different values for them. I believe here we have that case because not every user may want to skip this dialog. If we have for example 2 users which use the same machine then we can expect that one may want to have this dialog and second may want to skip this.

There is easy way to do this. Use "set" function from UserPreferencesService (this service can be imported from ADF's core). That services allows you to add this property to storage but additionally it prefixes it with username automatically.

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input/search-ai-input.component.spec.ts
AleksanderSklorz: I see my comment confused you a bit because I meant following: 
image

You can see that as callback we call search function so I meant that we can't test it in this way, we need to ensure that as callback logic from that function is triggered. In other case if we pass mocked function as callback then we still don't know if that logic works.

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-results/search-ai-results.component.spec.ts
AleksanderSklorz: Here the same comment like above, please check

projects/aca-content/src/lib/services/modal-ai.service.spec.ts
AleksanderSklorz: ok then let's leave that for now, thanks for checking

projects/aca-content/src/lib/services/modal-ai.service.ts
AleksanderSklorz: thank you for fixing that :)

AleksanderSklorz (approved):

PR #4147, opened 2024-09-30
[ACS-8779] Keep selections and question after going to the previous page

jacekpluta:
https://hyland.atlassian.net/browse/ACS-8779


projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input-container/search-ai-input-container.component.ts
```diff
@@ -28,15 +28,15 @@ import { MatIconModule } from '@angular/material/icon';
 import { SearchAiInputComponent } from '../search-ai-input/search-ai-input.component';
 import { MatDividerModule } from '@angular/material/divider';
 import { SearchAiNavigationService } from '../../../../services/search-ai-navigation.service';
-import { NavigationStart, Router } from '@angular/router';
-import { filter, takeUntil } from 'rxjs/operators';
 import { SearchAiService } from '@alfresco/adf-content-services';
 import { TranslateModule } from '@ngx-translate/core';
-import { Subject } from 'rxjs';
+import { Observable, Subject } from 'rxjs';
+import { SearchAiInputState } from '@alfresco/adf-content-services/lib/search-ai/models/search-ai-input-state';
```
AleksanderSklorz, line 34: import should be just: 
'@alfresco/adf-content-services'

AleksanderSklorz:
what with this case? When I click on X then files are selected but input is not pre-filled. Is that expected?

https://github.com/user-attachments/assets/a1ecf117-c64e-40f1-9e1b-67098274549d

AleksanderSklorz:
When we select files and run KR on some page for example personal files and then we will go to other page from KR for example share or favourites then files are not selected but question is refilled. I guess we should hide question as well in this case as we don't select any files for this page (as previous page was personal files)?


https://github.com/user-attachments/assets/c06f20bf-ac36-4e3d-aa29-1669fed29cf3

AleksanderSklorz:
When you run query for KR and then you will open for example personal files in separated tab and then you will run refresh on KR page in first tab then you will get error page. Before that change we could refresh page and use url to open results for last question's answer. 
I think the reason is that you reuse the same knowledgeRetrieval entry from localstorage which was used for that purpose. Solution can be introducing one more entry which keeps nodes (or maybe some flag instead of removing content for that entry for restoring?)

https://github.com/user-attachments/assets/5b133fdc-e24e-450e-a2f6-218efd11a440

AleksanderSklorz:
If I copy url to other page and then open that other page from KR by pasting url and then go to page like personal files then no files are selected but question is prefilled. We should hide question as well: 

https://github.com/user-attachments/assets/cc9f274b-d131-41bb-801b-0f796e966612

projects/aca-shared/src/lib/services/navigation-history.service.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  listenToRouteChanges() {
+    this.router.events
+      .pipe(
+        startWith(new NavigationEnd(0, this.router.url, this.router.url)),
+        filter((event: NavigationEnd) => event instanceof NavigationEnd)
+      )
+      .subscribe((event: NavigationEnd) => {
+        this.history.push(event.urlAfterRedirects);
+        if (this.history.length === 4) {
```
AleksanderSklorz, line 43: why 4? What is meaning for 4?

projects/aca-shared/src/lib/services/navigation-history.service.ts
```diff
@@ -0,0 +1,56 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * from Hyland Software. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { Injectable } from '@angular/core';
+import { NavigationEnd, Router } from '@angular/router';
+import { filter, startWith } from 'rxjs/operators';
+
+@Injectable({ providedIn: 'root' })
+export class NavigationHistoryService {
```
AleksanderSklorz, line 30: As this service is in shared - can we make it more reusable so if somebody would like to use it for something different than KR it would still work? For example make hardcoded '/knowledge-retrieval' in shouldReturnLastSelection function as parameter of function

projects/aca-shared/src/lib/services/navigation-history.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  private onDestroy$ = new Subject<boolean>();
+
+  history: string[] = [];
+
+  constructor(private router: Router) {}
+
+  listenToRouteChanges(maxHistoryLength) {
```
AleksanderSklorz, line 38: please set type to maxHistoryLength

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input-container/search-ai-input-container.component.spec.ts
AleksanderSklorz, line 1: You removed tests from this file as you removed some code but you added also some changes to implementation so you can write unit test for that change. 

In your test you can check if in html you assign searchTerm based on inputState$ which is set based on this.searchAiService.toggleSearchAiInput$

projects/aca-content/src/lib/components/sidenav/sidenav.component.ts
```diff
@@ -34,6 +34,7 @@ import { CommonModule } from '@angular/common';
 import { SidenavHeaderComponent } from './components/sidenav-header.component';
 import { MatListModule } from '@angular/material/list';
 import { ExpandMenuComponent } from './components/expand-menu.component';
+import { NavigationHistoryService } from '../../../../../aca-shared/src/lib/services/navigation-history.service';
```
AleksanderSklorz, line 37: should be imported from '@alfresco/aca-shared' I guess

projects/aca-shared/src/lib/services/navigation-history.service.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+ */
+
+import { NavigationEnd, NavigationStart, Router } from '@angular/router';
+import { TestBed } from '@angular/core/testing';
+import { NavigationHistoryService } from './navigation-history.service';
+import { Subject } from 'rxjs';
+
+describe('NavigationHistoryService', () => {
+  let service: NavigationHistoryService;
+  let routerEvents$: Subject<any>;
```
AleksanderSklorz, line 32: please try to avoid any, instead can be for example
```ts
Subject<NavigationEnd | NavigationStart>

File: projects/aca-shared/src/lib/components/document-base-page/document-base-page.spec.ts [hunk 7]
```diff
@@ -199,8 +209,18 @@ describe('Info Drawer state', () => {
 
   let store: MockStore<{ app: Partial<AppState> }>;
   let appState: Partial<AppState> = {};
+  let userPreferencesService: any;
```
AleksanderSklorz, line 212: we should avoid using any, please try use for all below services/router following: 
```ts
SpyObj<UserPreferencesService>
```

SpyObj type should resolve problem. When you assign these variables during injection you can do if needed: 
```ts
TestBed.inject(UserPreferencesService) as SpyObj<UserPreferencesService>
```

projects/aca-shared/src/lib/components/document-base-page/document-base-page.spec.ts
```diff
@@ -199,8 +209,18 @@ describe('Info Drawer state', () => {
...
+  let navigationHistoryService: any;
+  let searchAiService: any;
+  let router: any;
+  const mockNodes = JSON.stringify({ node: 'mockNode' });
```
AleksanderSklorz, line 216: please add empty line between let and const

AleksanderSklorz (approved):

PR #4189, opened 2024-10-23
[ACS-8907] Remove query / question from input on the results page

jacekpluta:
https://hyland.atlassian.net/browse/ACS-8907

### DIFF
--- a/projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input-container/search-ai-input-container.component.spec.ts
+++ b/projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input-container/search-ai-input-container.component.spec.ts
@@ -126,11 +126,11 @@ describe('SearchAiInputContainerComponent', () => {
       expect(inputComponent.agentId).toBe(component.agentId);
     });
 
-    it('should have assigned correct useStoredNodes flag', () => {
-      component.useStoredNodes = true;
+    it('should have assigned correct usedInAiResultsPage flag', () => {
+      component.usedInAiResultsPage = true;
       fixture.detectChanges();
 
-      expect(inputComponent.useStoredNodes).toBeTrue();
+      expect(inputComponent.usedInAiResultsPage).toBeTrue();
     });
 
     it('should set inputState$ to toggleSearchAiInput$ from the service on ngOnInit', () => {
--- a/projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input/search-ai-input.component.spec.ts
+++ b/projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input/search-ai-input.component.spec.ts
@@ -156,9 +156,9 @@ describe('SearchAiInputComponent', () => {
       expect(component.queryControl.value).toBe(query);
     });
 
-    it('should set queryControl value to "some new query" if useStoredNodes is equal to false', () => {
+    it('should set queryControl value to "some new query" if usedInAiResultsPage is equal to false', () => {
       const query = 'some new query';
-      component.useStoredNodes = false;
+      component.usedInAiResultsPage = false;
       component.searchTerm = query;
 
       component.ngOnInit();
@@ -166,9 +166,9 @@ describe('SearchAiInputComponent', () => {
       expect(component.queryControl.value).toBe('some new query');
     });
 
-    it('should set queryControl value to empty string if useStoredNodes is equal to true', () => {
+    it('should set queryControl value to empty string if usedInAiResultsPage is equal to true', () => {
       const query = 'some new query';
-      component.useStoredNodes = true;
+      component.usedInAiResultsPage = true;
       component.searchTerm = query;
 
       component.ngOnInit();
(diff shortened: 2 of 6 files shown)


tamaragruszka (approved):

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input/search-ai-input.component.ts
```diff
@@ -128,11 +128,16 @@ export class SearchAiInputComponent implements OnInit, OnDestroy {
...
   ngOnInit(): void {
-    if (this.searchTerm) {
+    const isKnowledgeRetrievalPage = this.router.url.startsWith('/knowledge-retrieval');
+
+    if (isKnowledgeRetrievalPage) {
+      this.queryControl.setValue('');
+    } else if (this.searchTerm) {
       this.queryControl.setValue(this.searchTerm);
     } else if (this.route.snapshot?.queryParams?.query?.length > 0) {
       this.queryControl.setValue(this.route.snapshot.queryParams.query);
```
AleksanderSklorz, line 143: Is this case still needed? Wasn't it needed for a case when you are on results page and refresh it but now we don't display it on results page so maybe it's redundant?

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input/search-ai-input.component.ts
```diff
@@ -128,11 +128,16 @@ export class SearchAiInputComponent implements OnInit, OnDestroy {
...
     private translateService: TranslateService,
     private modalAiService: ModalAiService,
-    private route: ActivatedRoute
+    private route: ActivatedRoute,
+    private router: Router
   ) {}
 
   ngOnInit(): void {
-    if (this.searchTerm) {
+    const isKnowledgeRetrievalPage = this.router.url.startsWith('/knowledge-retrieval');
```
AleksanderSklorz, line 136: I have one idea to avoid using url here because it would be safer to not use urls, just in case they are changed because when somebody changes url for KR then it can break this place. 

Above we have following input: 
useStoredNodes

This input is set only for search results to true, for other places it's not true. Can we rename this input to something like: "usedInAiResultsPage"? So then this flag can be used here instead of url as it will be true only for KR results page. This seems safer because it's easier to break something when string changes so if anyone changes url then it will be harder to detect

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input/search-ai-input.component.ts
```diff
@@ -127,20 +126,15 @@ export class SearchAiInputComponent implements OnInit, OnDestroy {
...
 
   ngOnInit(): void {
-    if (this.searchTerm) {
-      this.queryControl.setValue(this.searchTerm);
-    } else if (this.route.snapshot?.queryParams?.query?.length > 0) {
-      this.queryControl.setValue(this.route.snapshot.queryParams.query);
-    } else {
-      this.queryControl.setValue(null);
-    }
+    const usedInAiResultsPage = this.useStoredNodes;
```
AleksanderSklorz, line 133: Could you please rename that input useStoredNodes to usedInAiResultsPage? Actually it looks strange that we assign useStoredNodes to usedInAiResultsPage as the relationship is not clear. It will be easier to understand when we rename whole input (and usages in parents - please see it's used also in SearchAiInputContainerComponent)

g-jaskowski (approved):

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input-container/search-ai-input-container.component.spec.ts
```diff
@@ -127,10 +127,10 @@ describe('SearchAiInputContainerComponent', () => {
     });
 
     it('should have assigned correct useStoredNodes flag', () => {
```
AleksanderSklorz, line 129: should be 'usedInAiResultsPage' instead of 'useStoredNodes'

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input/search-ai-input.component.spec.ts
```diff
@@ -144,12 +157,24 @@ describe('SearchAiInputComponent', () => {
       expect(component.queryControl.value).toBe(query);
     });
 
-    it('should set queryControl value to query param if searchTerm is not defined', () => {
-      component.searchTerm = undefined;
+    it('should set queryControl value to "some new query" if useStoredNodes is equal to false', () => {
```
AleksanderSklorz, line 160: should be 'usedInAiResultsPage' instead of 'useStoredNodes'

projects/aca-content/src/lib/components/knowledge-retrieval/search-ai/search-ai-input/search-ai-input.component.spec.ts
```diff
@@ -144,12 +157,24 @@ describe('SearchAiInputComponent', () => {
...
+      component.usedInAiResultsPage = false;
+      component.searchTerm = query;
 
       component.ngOnInit();
 
-      expect(component.queryControl.value).toBe('some query');
+      expect(component.queryControl.value).toBe('some new query');
+    });
+
+    it('should set queryControl value to empty string if useStoredNodes is equal to true', () => {
```
AleksanderSklorz, line 170: should be 'usedInAiResultsPage' instead of 'useStoredNodes'

AleksanderSklorz (approved):

PR #4210, opened 2024-10-30
[ACS-8951] [E2E] tags and categories e2es added in ACA

datguychen:
https://hyland.atlassian.net/browse/ACS-8951

### DIFF
--- a/e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
+++ b/e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
@@ -58,7 +58,7 @@ test.describe('Info Drawer - File Folder Properties', () => {
   const categoryId2 = `-root-`;
   const categoryName = Utils.random();
 
-  async function createCategoryGetId() {
+  async function createCategoryGetId(): Promise<string> {
     const requestData = await categoriesApi.createCategory(categoryId2, [{ name: categoryName }]);
     if ('entry' in requestData) {
       return requestData.entry.id;
@@ -68,10 +68,10 @@ test.describe('Info Drawer - File Folder Properties', () => {
     }
   }
 
-  async function createTagGetId() {
+  async function createTagGetId(): Promise<string> {
     const requestTagData = await tagsApi.createTags([tagBody]);
-    if (Object.keys(requestTagData).includes('entry')) {
-      return (requestTagData as unknown as { entry: { id: string } }).entry.id;
+    if ('entry' in requestTagData) {
+      return (requestTagData as { entry: { id: string } }).entry.id;
     } else {
       console.error('Unexpected response format:', requestTagData);
       return null;
@@ -228,7 +228,7 @@ test.describe('Info Drawer - File Folder Properties', () => {
     await expect(personalFiles.infoDrawer.categoriesAccordion).toContainText('There are currently no categories added');
   });
 
-  test('[XAT-17243] Cancel adding a category to a node', async ({ personalFiles }) => {
+  test('[XAT-17244] Cancel adding a category to a node', async ({ personalFiles }) => {
     await setupFolderAndNavigate(personalFiles, Folder17244);
     await personalFiles.infoDrawer.categoriesAccordionPenButton.click();
     await personalFiles.infoDrawer.categoriesInput.fill('*');
--- a/e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
+++ b/e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
@@ -55,11 +55,12 @@ test.describe('Info Drawer - File Folder Properties', () => {
   const Folder17243 = `xat-17243-e2e-${Utils.random()}`;
   const Folder17244 = `xat-17244-e2e-${Utils.random()}`;
   const tagBody = { tag: `tag-${Utils.random()}` };
-  const categoryId2 = `-root-`;
   const categoryName = Utils.random();
+  const noCategoriesText = 'There are currently no categories added';
+  const noTagsText = 'There are currently no tags added';
 
   async function createCategoryGetId(): Promise<string> {
-    const requestData = await categoriesApi.createCategory(categoryId2, [{ name: categoryName }]);
+    const requestData = await categoriesApi.createCategory(`-root-`, [{ name: categoryName }]);
     if ('entry' in requestData) {
       return requestData.entry.id;
     } else {
@@ -138,10 +139,10 @@ test.describe('Info Drawer - File Folder Properties', () => {
   test('[XAT-17238] State for no tags and categories - accordion expanded', async ({ personalFiles }) => {
     await setupFolderAndNavigate(personalFiles, Folder17238);
     await personalFiles.infoDrawer.tagsAccordion.click();
-    await expect(personalFiles.infoDrawer.tagsAccordion).toContainText('There are currently no tags added');
+    await expect(personalFiles.infoDrawer.tagsAccordion).toContainText(noTagsText);
 
     await personalFiles.infoDrawer.categoriesAccordion.click();
-    await expect(personalFiles.infoDrawer.categoriesAccordion).toContainText('There are currently no categories added');
+    await expect(personalFiles.infoDrawer.categoriesAccordion).toContainText(noCategoriesText);
   });
 
   test('[XAT-17239] Add a new tag to a node', async ({ personalFiles }) => {
@@ -175,7 +176,7 @@ test.describe('Info Drawer - File Folder Properties', () => {
     await personalFiles.infoDrawer.tagsChipsXButton.first().click();
     await personalFiles.infoDrawer.tagsAccordionConfirmButton.click();
     await expect(personalFiles.infoDrawer.tagsChips.first()).toBeHidden();
-    await expect(personalFiles.infoDrawer.tagsAccordion).toContainText('There are currently no tags added');
+    await expect(personalFiles.infoDrawer.tagsAccordion).toContainText(noTagsText);
   });
 
   test('[XAT-17243] Cancel adding a tag to a node', async ({ personalFiles }) => {
@@ -191,7 +192,7 @@ test.describe('Info Drawer - File Folder Properties', () => {
     await personalFiles.infoDrawer.tagsInput.fill(manualTagName);
     await personalFiles.infoDrawer.tagsAccordionCancelButton.click();
     await expect(personalFiles.infoDrawer.tagsAccordionPenButton).toBeVisible();
-    await expect(personalFiles.infoDrawer.tagsAccordion).toContainText('There are currently no tags added');
+    await expect(personalFiles.infoDrawer.tagsAccordion).toContainText(noTagsText);
   });
 
   test('[XAT-17241] Add a new category to a node', async ({ personalFiles }) => {
@@ -206,7 +207,7 @@ test.describe('Info Drawer - File Folder Properties', () => {
     await personalFiles.infoDrawer.categoriesAccordionConfirmButton.click();
     await expect(personalFiles.infoDrawer.tagsChipsXButton.first()).toBeHidden();
     await expect(personalFiles.infoDrawer.tagsAccordionPenButton).toBeVisible();
-    await expect(personalFiles.infoDrawer.categoriesAccordion).not.toContainText('There are currently no categories added');
+    await expect(personalFiles.infoDrawer.categoriesAccordion).not.toContainText(noCategoriesText);
     await expect(personalFiles.infoDrawer.categoriesCreatedList.first()).toBeVisible();
   });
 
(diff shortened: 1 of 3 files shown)


datguychen:
One small issue: One test is creating a tag manually, and I didn't automate deleting it because I would have to create a function that listens to the request-response and extracts the ID, which could take a lot of time. The tags are deleted when the environment is cleared, so I decided to not spend time on it.

Let me know if I should work on it anyway.
@AleksanderSklorz @MichalKinas

e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
```diff
@@ -50,41 +101,139 @@ test.describe('Info Drawer - File Folder Properties', () => {
...
+
+    await personalFiles.infoDrawer.categoriesAccordion.click();
+    await expect(personalFiles.infoDrawer.categoriesAccordion).not.toContainText('There are currently no categories added');
+    await personalFiles.infoDrawer.categoriesAccordionPenButton.click();
+    await personalFiles.infoDrawer.categoriesItemRemoveButton.first().click();
+    await personalFiles.infoDrawer.categoriesAccordionConfirmButton.click();
+    await expect(personalFiles.infoDrawer.categoriesAccordion).toContainText('There are currently no categories added');
+  });
+
+  test('[XAT-17243] Cancel adding a category to a node', async ({ personalFiles }) => {
```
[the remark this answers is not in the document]
datguychen: Small mistake, will be fixed in the next push. Will wait for more reviews first.

e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
```diff
@@ -23,13 +23,60 @@
...
+      return requestData.entry.id;
+    } else {
+      console.error('Unexpected response format:', requestData);
+      return null;
+    }
+  }
+
+  async function createTagGetId() {
+    const requestTagData = await tagsApi.createTags([tagBody]);
+    if (Object.keys(requestTagData).includes('entry')) {
```
akashrathod28, line 73: we can replace this if with better code like below to replace unknown type checking.
if ('entry' in requestTagData) {
            return (requestTagData as { entry: { id: string } }).entry.id;
        }

datguychen: I was using 'entry' in request but sonar cloud is giving me this error:
`**_Use "indexOf" or "includes" (available from ES2016) instead._**`

I switched it back to using 'in' keyword and the issue occurs again. What should I do here?

datguychen: https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4210&id=Alfresco_alfresco-content-app&open=AZLhr2TdcCVkb1Q-mrj7

AleksanderSklorz:
> One small issue: One test is creating a tag manually, and I didn't automate deleting it because I would have to create a function that listens to the request-response and extracts the ID, which could take a lot of time. The tags are deleted when the environment is cleared, so I decided to not spend time on it.
> 
> Let me know if I should work on it anyway. @AleksanderSklorz @MichalKinas

@datguychen me and @MichalKinas we think that it would be good to delete tag if possible because in theory it can affect on later added tests which we can add in the future. Each test should be independent of others but leaving tag from some test can affect on some later test.

projects/aca-playwright-shared/src/page-objects/components/adf-info-drawer.component.ts
```diff
@@ -63,6 +63,24 @@ export class AdfInfoDrawerComponent extends BaseComponent {
   public descriptionField = this.page.locator('textarea[placeholder=Description]');
   public visibilityField = this.infoDrawerPanel.getByRole('combobox');
   public selectVisibility = (visibilityOption: string) => this.page.getByRole('listbox').getByRole('option', { name: visibilityOption }).click();
+  public tagsAccordion = this.getChild('[data-automation-id="adf-content-metadata-tags-panel"]');
+  public categoriesAccordion = this.getChild('[data-automation-id="adf-content-metadata-categories-panel"]');
+  public tagsAccordionPenButton = this.tagsAccordion.locator('[data-automation-id="showing-tag-input-button"]');
+  public categoriesAccordionPenButton = this.categoriesAccordion.locator('[data-automation-id="meta-data-categories-edit"]');
+  private readonly tagsCreator = this.getChild('adf-tags-creator');
```
AleksanderSklorz, line 70: can you keep private readonly fields separately from public fields?

projects/aca-playwright-shared/src/page-objects/components/adf-info-drawer.component.ts
```diff
@@ -63,6 +63,24 @@ export class AdfInfoDrawerComponent extends BaseComponent {
...
+  public tagsInput = this.tagsCreator.locator('input');
+  public createTagButton = this.tagsCreator.locator('[role="button"]');
+  public tagsChips = this.tagsCreator.locator('mat-chip');
+  public tagsChipsXButton = this.tagsChips.locator('button');
+  public tagsAccordionCancelButton = this.getChild('[data-automation-id="reset-tags-metadata"]');
+  public tagsAccordionConfirmButton = this.getChild('[data-automation-id="save-tags-metadata"]');
+  public categoriesAccordionCancelButton = this.getChild('[data-automation-id="reset-metadata"]');
+  public categoriesAccordionConfirmButton = this.getChild('[data-automation-id="save-categories-metadata"]');
+  private readonly categoriesManagement = this.getChild('adf-categories-management');
```
AleksanderSklorz, line 79: same like above

datguychen:
> > One small issue: One test is creating a tag manually, and I didn't automate deleting it because I would have to create a function that listens to the request-response and extracts the ID, which could take a lot of time. The tags are deleted when the environment is cleared, so I decided to not spend time on it.
> > Let me know if I should work on it anyway. @AleksanderSklorz @MichalKinas
> 
> @datguychen me and @MichalKinas we think that it would be good to delete tag if possible because in theory it can affect on later added tests which we can add in the future. Each test should be independent of others but leaving tag from some test can affect on some later test.

@AleksanderSklorz Okay, I will implement this in my tests. Thanks for answering!

AleksanderSklorz:
@datguychen there is some sonar issue, please fix: 
image

e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
datguychen: @AleksanderSklorz the sonar cloud issue is discussed here.

AleksanderSklorz: Hmm I see. Honestly not sure why Sonar is complaining about that. Using 'in' should be fine as well in my opinion  as it looks simpler. @MichalKinas is there any reason why we can't use 'in'? Should we somehow change it for sonar?

MichalKinas: In this case I think Sonar is right since `requestTagData` is an array an `in` keyword is TS work for objects, in case of arrays it won't check values but rather indexes of array so this could lead to some errors, here since you're creating only one tag you can use `'entry' in requestTagData[0]` and Sonar should be okay with that

AleksanderSklorz: ah, good catch :D I didn't notice it's array. Variable is named requestTagData so I thought it's some object with tag data :D

MichalKinas: Also this line `return (requestTagData as { entry: { id: string } }).entry.id;` looks a bit weird, since you know you created only one tag you can just return `return requestTagData[0].entry.id;`

e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
```diff
@@ -50,41 +102,139 @@ test.describe('Info Drawer - File Folder Properties', () => {
 
   test.afterAll(async () => {
     await Utils.deleteNodesSitesEmptyTrashcan(nodesApi, trashcanApi, 'afterAll failed');
+    await categoriesApi.deleteCategory(responseCategoryId);
+    await tagsApi.deleteTag(responseTagsId);
```
MichalKinas, line 106: `responseTagsId` may be null as well

datguychen: Fixed, please check if properly.

e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
```diff
@@ -50,41 +102,139 @@ test.describe('Info Drawer - File Folder Properties', () => {
 
   test.afterAll(async () => {
     await Utils.deleteNodesSitesEmptyTrashcan(nodesApi, trashcanApi, 'afterAll failed');
+    await categoriesApi.deleteCategory(responseCategoryId);
```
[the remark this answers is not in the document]
datguychen: Fixed, please check if properly.

e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
datguychen: @MichalKinas With my code the tests pass normally, but with your suggestions I get this error:

beforeAll failed : TypeError: Cannot use 'in' operator to search for 'entry' in undefined

MichalKinas: Can you please put a breakpoint there or console log what api returns? I just want to make sure what is the type of returned object

datguychen: Weirdly, it creates 3 tags.
const createdTag = await tagsApi.createTags([tagBody]);

returns this:
{
  entry: { tag: 'tag-1jj01', id: '52d24d73-48b3-418a-924d-7348b3e18a2f' }
}
{
  entry: { tag: 'tag-19453', id: '963abe54-1e21-4e78-babe-541e21ce7877' }
}
{
  entry: { tag: 'tag-13d1t', id: '9db166f6-8c5c-4ac8-b166-f68c5c7ac8eb' }
}

and all of them are deleted after the test.

MichalKinas: Yeah so as we were speaking it returns array of tags (don't know why 3 though) so you can't use `if ('entry' in createdTag)` but rather `if ('entry' in createdTags[0])`

datguychen: The tests just doesn't work when I try to use `if ('entry' in createdTags[0])`

`  async function createTagGetId(): Promise<string> {
    const createdTag = await tagsApi.createTags([tagBody]);
    if ('entry' in createdTag[0]) {
      return createdTag[0].entry.id;
    } else {
      console.error('Unexpected response format:', createdTag);
      return null;
    }
  }`

I get this error:
beforeAll failed : TypeError: Cannot use 'in' operator to search for 'entry' in undefined

Then the tests all fail.

datguychen: The tests just doesn't work when I try to use `if ('entry' in createdTags[0])`

`  async function createTagGetId(): Promise<string> {
    const createdTag = await tagsApi.createTags([tagBody]);
    if ('entry' in createdTag[0]) {
      return createdTag[0].entry.id;
    } else {
      console.error('Unexpected response format:', createdTag);
      return null;
    }
  }`

I get this error:
beforeAll failed : TypeError: Cannot use 'in' operator to search for 'entry' in undefined

Then the tests all fail.

datguychen: and if I try this:

  async function createTagGetId(): Promise<string> {
    const createdTag = await tagsApi.createTags([tagBody]);
    if ('entry' in createdTag) {
      return createdTag[0].entry.id;
    } else {
      console.error('Unexpected response format:', createdTag);
      return null;
    }
  }

I get this error:
beforeAll failed : TypeError: Cannot read properties of undefined (reading 'entry')

I guess the TagEntry type is the problem here, but I don't understand why. @MichalKinas

datguychen: That's why I used this line of code `return (createdTag as { entry: { id: string } }).entry.id;`, because it treats the createdTag as a specific object with entry and id, and not as TagEntry which is somehow interpreted differently in this case.

MichalKinas: So in your first case you got `beforeAll failed : TypeError: Cannot use 'in' operator to search for 'entry' in undefined` which means that in that case `createdTag[0]` is undefined which means that `tagsApi.createTags` didn't return tags, the rest of the code should work fine I even tested that in jsconsole -> 
Zrzut ekranu 2024-11-5 o 09 45 43

MichalKinas: If you still get the error please let me know how to run this case locally I will try to debug

datguychen: @MichalKinas in both cases the tags are created. This is the case with createdTag[0]. I still don't understand why it creates 3 tags, but I don't think this is the issue right now. The code is up-to-date on this branch, so if you want to debug it locally, you can do it. You can message me if you encounter any issues with running the e2es.

createdTag: {
  entry: { tag: 'tag-1ld4r', id: '222ce83c-e6a1-487c-ace8-3ce6a1487c98' }
}
beforeAll failed : TypeError: Cannot use 'in' operator to search for 'entry' in undefined
createdTag: {
  entry: { tag: 'tag-10mri', id: '2c78ac98-e2a9-44c8-b8ac-98e2a9a4c894' }
}
beforeAll failed : TypeError: Cannot use 'in' operator to search for 'entry' in undefined
createdTag: {
  entry: { tag: 'tag-137wp', id: 'bcf4dce5-93be-4955-b4dc-e593be395537' }
}
beforeAll failed : TypeError: Cannot use 'in' operator to search for 'entry' in undefined

AleksanderSklorz (approved):

PR #4226, opened 2024-11-06
[ACS-8981] [E2E] added deleteTagsByTagName method for TagsApiin ACA

datguychen:
https://hyland.atlassian.net/browse/ACS-8981


e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
```diff
@@ -132,18 +132,6 @@ test.describe('Info Drawer - File Folder Properties', () => {
     await personalFiles.acaHeader.viewDetails.click();
   }
 
-  async function waitForTagToBeAdded(folderId: string, personalFiles: PersonalFilesPage, maxRetries: number) {
-    let retries = 0;
-
-    while ((await tagsApi.listTagsForNode(folderId)).list.entries.length === 0) {
```
AleksanderSklorz, line 138: listTagsForNode is unused function now so can be removed

projects/aca-playwright-shared/src/api/tags-api.ts
```diff
@@ -72,4 +74,23 @@ export class TagsApi {
       return null;
     }
   }
+
+  async listTags(params?: { tag?: string; matching?: boolean }): Promise<TagPaging> {
```
AleksanderSklorz, line 78: I think there is no sense to create separate function just to wrap listTags function from TagsApi. I can see it's used now only in single place (below in deleteTagsByTagName function). Actually that extra wrapper doesn't give any value because in deleteTagsByTagName you have already try catch so you can call listTags from TagsApi in deleteTagsByTagName function directly and thanks that here will be less redundant code.

e2e/playwright/info-drawer/src/tests/file-folder-properties.e2e.ts
datguychen: I would leave it because it's in the tagsApi, so it can be used in the future. Its not a custom method, its just an api call from swagger.

AleksanderSklorz (approved):

dominikiwanekhyland (approved):

PR #4229, opened 2024-11-07
[ACS-8634] Add new option to edit changes in saved search or save as new

MichalKinas:
https://hyland.atlassian.net/browse/ACS-8634

Previously there was no option to edit params of saved search, now when user navigates from saved search we can choose to either save changes or save new search.

### DIFF
--- a/projects/aca-content/assets/i18n/en.json
+++ b/projects/aca-content/assets/i18n/en.json
@@ -211,6 +211,8 @@
               "RESET_ACTION": "Reset search filters",
               "SAVE_SEARCH": {
                 "ACTION_BUTTON": "Save Search",
+                "SAVE_CHANGES": "Save changes",
+                "SAVE_AS_NEW": "Save as new",
                 "MODAL_HEADER": "Save this search",
                 "NAME_LABEL": "Name",
                 "NAME_REQUIRED_ERROR": "This field is required",
--- a/projects/aca-content/assets/i18n/en.json
+++ b/projects/aca-content/assets/i18n/en.json
@@ -217,6 +217,7 @@
                 "DESCRIPTION_LABEL": "Description",
                 "SAVE_SUCCESS": "Search Saved",
                 "SAVE_ERROR": "Error occurred. Search could not be saved.",
+                "SEARCH_NAME_NOT_UNIQUE_ERROR": "Saved Search with {{ name }} name already exists.",
                 "NAVBAR": {
                   "TITLE": "Saved Searches ({{ number }})",
                   "MANAGE_BUTTON": "Manage searches"
--- a/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
+++ b/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
@@ -26,16 +26,21 @@ import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testin
 import { SearchResultsComponent } from './search-results.component';
 import { AppConfigService, NotificationService, TranslationService } from '@alfresco/adf-core';
 import { Store } from '@ngrx/store';
-import { NavigateToFolder } from '@alfresco/aca-shared/store';
+import { NavigateToFolder, SnackbarErrorAction, SnackbarInfoAction } from '@alfresco/aca-shared/store';
 import { Pagination, SearchRequest } from '@alfresco/js-api';
-import { SearchQueryBuilderService } from '@alfresco/adf-content-services';
+import { SavedSearchesService, SearchQueryBuilderService } from '@alfresco/adf-content-services';
 import { ActivatedRoute, Router } from '@angular/router';
-import { BehaviorSubject, of, Subject } from 'rxjs';
+import { BehaviorSubject, of, Subject, throwError } from 'rxjs';
 import { AppTestingModule } from '../../../testing/app-testing.module';
 import { AppService } from '@alfresco/aca-shared';
 import { MatSnackBarModule } from '@angular/material/snack-bar';
 import { Buffer } from 'buffer';
 import { testHeader } from '../../../testing/document-base-page-utils';
+import { HarnessLoader } from '@angular/cdk/testing';
+import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import { MatMenuModule } from '@angular/material/menu';
+import { MatMenuHarness } from '@angular/material/menu/testing';
 
 describe('SearchComponent', () => {
   let component: SearchResultsComponent;
@@ -49,6 +54,10 @@ describe('SearchComponent', () => {
   const searchRequest = {} as SearchRequest;
   let params: BehaviorSubject<any>;
   let showErrorSpy: jasmine.Spy;
+  let loader: HarnessLoader;
+
+  const editSavedSearchesSpy = jasmine.createSpy('editSavedSearch');
+  const getSavedSearchButton = () => fixture.nativeElement.querySelector('.aca-content__save-search-action');
 
   const encodeQuery = (query: any): string => {
     return Buffer.from(JSON.stringify(query)).toString('base64');
@@ -57,7 +66,7 @@ describe('SearchComponent', () => {
   beforeEach(() => {
     params = new BehaviorSubject({ q: 'TYPE: "cm:folder" AND %28=cm: name: email OR cm: name: budget%29' });
     TestBed.configureTestingModule({
-      imports: [AppTestingModule, SearchResultsComponent, MatSnackBarModule],
+      imports: [AppTestingModule, SearchResultsComponent, MatSnackBarModule, MatMenuModule, NoopAnimationsModule],
       providers: [
         {
           provide: AppService,
@@ -68,11 +77,21 @@ describe('SearchComponent', () => {
           }
         },
         {
+          provide: SavedSearchesService,
+          useValue: {
+            getSavedSearches: jasmine.createSpy('getSavedSearches').and.returnValue(of([{ name: 'test', encodedUrl: 'test', order: 0 }])),
+            editSavedSearch: editSavedSearchesSpy
+          }
+        },
+        {
           provide: ActivatedRoute,
           useValue: {
             snapshot: {
               data: {
                 sortingPreferenceKey: ''
+              },
+              queryParams: {
+                q: 'test'
               }
             },
             params: params.asObservable()
@@ -102,6 +121,7 @@ describe('SearchComponent', () => {
     spyOn(queryBuilder, 'update').and.stub();
 
     fixture.detectChanges();
+    loader = TestbedHarnessEnvironment.loader(fixture);
   });
 
   afterEach(() => {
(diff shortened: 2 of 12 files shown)


projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -140,18 +147,21 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
   isLoading = false;
   totalResults: number;
   isTagsEnabled = false;
+  initialSavedSearch = undefined;
```
nikita-web-ua, line 150: can we specify type here?

projects/aca-content/assets/i18n/en.json
```diff
@@ -211,12 +211,15 @@
...
                 "ACTION_BUTTON": "Save Search",
+                "SAVE_CHANGES": "Save changes",
+                "SAVE_AS_NEW": "Save as new",
                 "MODAL_HEADER": "Save this search",
                 "NAME_LABEL": "Name",
                 "NAME_REQUIRED_ERROR": "This field is required",
                 "DESCRIPTION_LABEL": "Description",
                 "SAVE_SUCCESS": "Search Saved",
                 "SAVE_ERROR": "Error occurred. Search could not be saved.",
+                "SEARCH_NAME_NOT_UNIQUE_ERROR": "Saved Search with {{ name }} name already exists.",
```
nikita-web-ua, line 222: it might be worth enclosing the `{{ name }}` in single parentheses

g-jaskowski (approved):

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -164,7 +174,7 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
 
     this.queryBuilder.configUpdated
       .asObservable()
-      .pipe(takeUntil(this.onDestroy$))
+      .pipe(takeUntilDestroyed(this.destroyRef))
```
dominikiwanekhyland, line 177: I think if used in component, takeUntilDestroyed() is enough, without destroyRef as parameter

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -202,7 +212,14 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
     this.columns = this.extensions.documentListPresets.searchResults || [];
 
     if (this.route) {
-      this.route.queryParams.pipe(takeUntil(this.onDestroy$)).subscribe((params: Params) => {
+      this.savedSearchesService
```
dominikiwanekhyland, line 215: do we want to do this only once with first()? What if we switch on navbar to different save search?

dominikiwanekhyland (approved):

projects/aca-content/src/lib/components/search/search-save/dialog/edit/saved-search-edit-dialog.component.ts
```diff
@@ -42,7 +43,11 @@ import { FormControl, FormGroup, Validators } from '@angular/forms';
 })
 export class SavedSearchEditDialogComponent {
   form = new FormGroup({
-    name: new FormControl('', [Validators.required, forbidOnlySpaces]),
+    name: new FormControl('', {
+      validators: [Validators.required, forbidOnlySpaces],
+      asyncValidators: [this.uniqueSearchNameValidator.validate.bind(this.uniqueSearchNameValidator)],
```
dominikiwanekhyland, line 48: https://hyland.atlassian.net/browse/ACS-8621
about the name:
"the name  does NOT need to be unique "
has something changed here?

dominikiwanekhyland (changes requested):

projects/aca-content/src/lib/components/search/search-save/dialog/unique-search-name-validator.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Injectable } from '@angular/core';
+import { AbstractControl, AsyncValidator, ValidationErrors } from '@angular/forms';
+import { catchError, map, Observable, of } from 'rxjs';
+
+@Injectable({ providedIn: 'root' })
+export class UniqueSearchNameValidator implements AsyncValidator {
+  constructor(private savedSearchesService: SavedSearchesService) {}
+
+  validate(control: AbstractControl): Observable<ValidationErrors | null> {
+    return this.savedSearchesService.getSavedSearches().pipe(
```
dominikiwanekhyland, line 35: I think we can get those save searches from here: 
this.savedSearchesService.savedSearches$.asObservable() as getSavedSearches() fetches the data and reads it from the file

projects/aca-content/assets/i18n/en.json
MichalKinas: Why? This is not a number and I think we don't enclose names in parentheses in other cases

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
MichalKinas: I'm trying to think which solution will be best here

projects/aca-content/src/lib/components/search/search-save/dialog/edit/saved-search-edit-dialog.component.ts
MichalKinas: I was discussing this with Arabella and we agreed it feels weird for the name not to be unique so I added a change for that, will update the ticket

projects/aca-content/src/lib/components/search/search-save/dialog/unique-search-name-validator.ts
MichalKinas: I tried that but for some reason it doesn't work, after changing the input value the result is not updated

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
MichalKinas: Added, thanks

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
MichalKinas: Good catch , added

projects/aca-content/assets/i18n/en.json
nikita-web-ua: I can see somewhere we do and in other places we don't, so it is up to you. 
Screenshot 2024-11-07 at 13 58 22

nikita-web-ua (approved):

dominikiwanekhyland (approved):

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
```diff
@@ -49,6 +54,10 @@ describe('SearchComponent', () => {
   const searchRequest = {} as SearchRequest;
   let params: BehaviorSubject<any>;
   let showErrorSpy: jasmine.Spy;
+  let loader: HarnessLoader;
+
+  const editSavedSearchesSpy = jasmine.createSpy('editSavedSearch');
+  const getSavedSearchButton = () => fixture.nativeElement.querySelector('.aca-content__save-search-action');
```
AleksanderSklorz, line 60: can you set type of return for this function?

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
MichalKinas: I found a fix for it, you can check out the last commit

projects/aca-content/src/lib/components/search/search-save/dialog/unique-search-name-validator.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  });
+
+  describe('Save searches returns results', () => {
+    beforeEach(() => {
+      TestBed.overrideProvider(SavedSearchesService, { useValue: { getSavedSearches: () => of([{ name: 'test' }]) } });
+      validator = TestBed.inject(UniqueSearchNameValidator);
+    });
+
+    it('should return null when name is unique', (done) => {
+      validator.validate({ value: 'unique' } as any).subscribe((result) => {
```
AleksanderSklorz, line 46: Maybe instead of as any we can use as AbstractControl? The same for rest tests which are below

AleksanderSklorz:
I found strange issue that sometimes arrow to expand navbar is disappeared when navbar is collapsed so I can't show navbar again

https://github.com/user-attachments/assets/69606d4e-5ee8-4193-a5bb-95de590b6c52

AleksanderSklorz:
Is that expected behavior? When I edit some search and save it as new and then I want to edit this new search then I have no 2 options like "Save changes" and "Save as new". I have only single option. I can imagine somebody may want to have option to save it as new or to modify existing one but actually there is no option  

https://github.com/user-attachments/assets/65533d5e-6999-4882-8a18-3d1c08412845

AleksanderSklorz:
I see multiple issues in case when you edit saved search with duplicated configuration. See below video: 

https://github.com/user-attachments/assets/159e63dd-52d2-4438-bbed-03e64fc9ff5b

There I have two saved searches which search for "c". When I select one of them then both are highlighted (what is caused by the same url but it's ok). The problems are following: 
- if you edit one of them then when you for example refresh page then changes are not saved and you can see state before editing 
- initially both are highlighted after selecting one of them because of the same configuration but I guess when the configuration is edited then only one should be highlighted as they don't have the same config anymore? 
- I'm editing only one of them so theoretically I can expect that only one will be edited but initially before refresh you can see that both searches have new edited config instead of only that one which I clicked

AleksanderSklorz:
When I create new saved search then save button is disabled until I do something. When I edit saved search - button is always enabled even if I didn't do any changes. I can imagine case that somebody may want to have option to "save as new" without doing changes (for example somebody want's to copy saved search) but there is probably no reason to allow to save search without changes. Maybe it has sense to display only one option (save as new) when there is no changes?

https://github.com/user-attachments/assets/c66830ce-7a84-4cc5-9110-babba9afca01

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
MichalKinas: Added

projects/aca-content/src/lib/components/search/search-save/dialog/unique-search-name-validator.spec.ts
MichalKinas: Added regular FormControl there

MichalKinas:
> I found strange issue that sometimes arrow to expand navbar is disappeared when navbar is collapsed so I can't show navbar again
> 
>  Screen.Recording.2024-11-07.at.15.25.12.mov

It's fixed now

MichalKinas:
> Is that expected behavior? When I edit some search and save it as new and then I want to edit this new search then I have no 2 options like "Save changes" and "Save as new". I have only single option. I can imagine somebody may want to have option to save it as new or to modify existing one but actually there is no option
> 
>  Screen.Recording.2024-11-07.at.15.27.27.mov

Right now you will still have 2 options to either save as new or to save changes but in the original saved search since you didn't navigate to new one which is expected

MichalKinas:
> I see multiple issues in case when you edit saved search with duplicated configuration. See below video:
> 
>  Screen.Recording.2024-11-07.at.15.35.57.mov 
> There I have two saved searches which search for "c". When I select one of them then both are highlighted (what is caused by the same url but it's ok). The problems are following:
> 
> * if you edit one of them then when you for example refresh page then changes are not saved and you can see state before editing
> * initially both are highlighted after selecting one of them because of the same configuration but I guess when the configuration is edited then only one should be highlighted as they don't have the same config anymore?
> * I'm editing only one of them so theoretically I can expect that only one will be edited but initially before refresh you can see that both searches have new edited config instead of only that one which I clicked

As far as I remember during the x-in-a-box we assumed that nobody will save exactly the same search so this is not covered right now, we can talk to Arabella if we want to cover this case and plan it as improvement in next RT if so

MichalKinas:
> When I create new saved search then save button is disabled until I do something. When I edit saved search - button is always enabled even if I didn't do any changes. I can imagine case that somebody may want to have option to "save as new" without doing changes (for example somebody want's to copy saved search) but there is probably no reason to allow to save search without changes. Maybe it has sense to display only one option (save as new) when there is no changes?
> 
>  Screen.Recording.2024-11-07.at.15.42.50.mov

This is expected since the only condition to be able to save is valid encoded query, that's why you need to perform any search before you'll be able to save any changes

AleksanderSklorz:
> > I found strange issue that sometimes arrow to expand navbar is disappeared when navbar is collapsed so I can't show navbar again
> > Screen.Recording.2024-11-07.at.15.25.12.mov
> 
> It's fixed now

this issue still exists but as we discussed on slack - it will be fixed in separated task as this issue is not related with changes for Saved Searches

AleksanderSklorz (approved):

PR #4238, opened 2024-11-15
[ACS-8991] Fix sidenav issue when clicking on manage saved searches

MichalKinas:
Fixes https://hyland.atlassian.net/browse/ACS-8991, depends on https://github.com/Alfresco/alfresco-ng2-components/pull/10393


swapnil-verma-gl:
Not sure if this is the write PR to post this or not, but I noticed a behaviour issue in regards to search. After doing a search, If i open the search input layover, and select the file/folder checkbox, then there is no loading bar shown while fetching the results. There is a search query fired in the background as the search results change accordingly after some time. This behaviour is not present on ACADEV. Also, this behaviour only applies to file/folder checkbox. The libraries checkbox works as expected. See video

https://github.com/user-attachments/assets/7c73008b-fcc3-4c9c-b08d-fc8e35198f8f

AleksanderSklorz (approved):

nikita-web-ua (approved):

PR #4243, opened 2024-11-18
[ACS-9008] Proper form initialization for Saved Searches

MichalKinas:
Form might be initialized https://hyland.atlassian.net/browse/ACS-9008, after the fix form is always initialized when the validator is injected.

### DIFF
--- a/projects/aca-content/src/lib/components/search/search-save/dialog/edit/saved-search-edit-dialog.component.ts
+++ b/projects/aca-content/src/lib/components/search/search-save/dialog/edit/saved-search-edit-dialog.component.ts
@@ -42,15 +42,7 @@ import { UniqueSearchNameValidator } from '../unique-search-name-validator';
   host: { class: 'aca-saved-search-edit-dialog' }
 })
 export class SavedSearchEditDialogComponent {
-  form = new FormGroup({
-    name: new FormControl('', {
-      validators: [Validators.required, forbidOnlySpaces],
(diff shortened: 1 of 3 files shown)


nikita-web-ua (approved):

swapnil-verma-gl (approved):

projects/aca-content/src/lib/components/search/search-save/dialog/edit/saved-search-edit-dialog.component.ts
```diff
@@ -42,15 +42,7 @@ import { UniqueSearchNameValidator } from '../unique-search-name-validator';
...
-  form = new FormGroup({
-    name: new FormControl('', {
-      validators: [Validators.required, forbidOnlySpaces],
-      asyncValidators: [this.uniqueSearchNameValidator.validate.bind(this.uniqueSearchNameValidator)],
-      updateOn: 'blur'
-    }),
-    description: new FormControl('')
-  });
-
+  form: FormGroup;
```
AleksanderSklorz, line 45: could you please set type of that generic FormGroup type? Actually it takes any so it will cause any for values etc. if we will try to read value from that form: 
Screenshot 2024-11-18 at 12 39 04

MichalKinas: Done, I've also applied the same fix for the other dialog for saved searches

PR #4251, opened 2024-11-21
[ACS-9010] Loading spinner is not shown on search results page when filtering for files or folders

dominikiwanekhyland:
https://hyland.atlassian.net/browse/ACS-9010

### DIFF
--- a/projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
+++ b/projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
@@ -23,11 +23,11 @@
  */
 
 import { NO_ERRORS_SCHEMA } from '@angular/core';
-import { TestBed, ComponentFixture } from '@angular/core/testing';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { SearchInputComponent } from './search-input.component';
 import { AppTestingModule } from '../../../testing/app-testing.module';
 import { Actions, ofType } from '@ngrx/effects';
-import { SearchByTermAction, SearchActionTypes } from '@alfresco/aca-shared/store';
+import { SearchActionTypes, SearchByTermAction, SearchOptionIds } from '@alfresco/aca-shared/store';
 import { AppHookService, AppService } from '@alfresco/aca-shared';
 import { map } from 'rxjs/operators';
 import { SearchQueryBuilderService } from '@alfresco/adf-content-services';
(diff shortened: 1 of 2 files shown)


projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -239,4 +239,205 @@ describe('SearchInputComponent', () => {
     });
     fixture.detectChanges();
   });
+
+  describe('searchByOption()', () => {
```
AleksanderSklorz, line 243: the name of describe should be just 'searchByOption' without ()

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -239,4 +239,205 @@ describe('SearchInputComponent', () => {
...
+    let isLibrariesCheckedSpy: jasmine.Spy;
```
AleksanderSklorz, line 244: jasmine.Spy is generic type so please set type of function which you mock in place of <>

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -239,4 +239,205 @@ describe('SearchInputComponent', () => {
...
+      expect(storeDispatchSpy).toHaveBeenCalledWith(new SearchByTermAction('test', component.searchOptions));
+      expect(queryLibrariesBuilderUpdateSpy).not.toHaveBeenCalled();
+    });
+
+    it('should add folder filter when only folders are checked', () => {
+      isLibrariesCheckedSpy.and.returnValue(false);
+      isFoldersCheckedSpy.and.returnValue(true);
+      isFilesCheckedSpy.and.returnValue(false);
+
+      spyOn(component['queryBuilder'], 'addFilterQuery').and.callThrough();
```
AleksanderSklorz, line 337: it's a bit hacking to get private field using ['queryBuilder']. Private field is private so should not be accessible. Instead you should get this field using injection (please use TestBed.inject(SearchQueryBuilderService). 

The same issue is in other tests so please correct them as well.

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -239,4 +239,205 @@ describe('SearchInputComponent', () => {
...
+    beforeEach(() => {
+      isLibrariesCheckedSpy = spyOn(component, 'isLibrariesChecked').and.returnValue(false);
+      isFoldersCheckedSpy = spyOn(component, 'isFoldersChecked').and.returnValue(false);
+      isFilesCheckedSpy = spyOn(component, 'isFilesChecked').and.returnValue(false);
+      evaluateLibrariesConstraintSpy = spyOn(component, 'evaluateLibrariesConstraint').and.returnValue(false);
+      syncInputValuesSpy = spyOn(component, 'syncInputValues').and.callThrough();
+      isSameSearchTermSpy = spyOn(component, 'isSameSearchTerm').and.returnValue(false);
+
+      queryBuilderUpdateSpy = spyOn(component['queryBuilder'], 'update');
+      queryLibrariesBuilderUpdateSpy = spyOn(component['queryLibrariesBuilder'], 'update');
```
AleksanderSklorz, line 263: Please use the same approach like suggested here for queryBuilder: 
https://github.com/Alfresco/alfresco-content-app/pull/4251/files#r1858523721

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -239,4 +239,205 @@ describe('SearchInputComponent', () => {
...
+      storeDispatchSpy = spyOn(component['store'], 'dispatch');
```
AleksanderSklorz, line 264: Please use the same approach like suggested here for queryBuilder:
https://github.com/Alfresco/alfresco-content-app/pull/4251/files#r1858523721

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -239,4 +239,205 @@ describe('SearchInputComponent', () => {
...
+    let isFilesCheckedSpy: jasmine.Spy;
+    let evaluateLibrariesConstraintSpy: jasmine.Spy;
+    let syncInputValuesSpy: jasmine.Spy;
+    let isSameSearchTermSpy: jasmine.Spy;
+    let queryBuilderUpdateSpy: jasmine.Spy;
+    let queryLibrariesBuilderUpdateSpy: jasmine.Spy;
+    let storeDispatchSpy: jasmine.Spy;
+
+    beforeEach(() => {
+      isLibrariesCheckedSpy = spyOn(component, 'isLibrariesChecked').and.returnValue(false);
```
AleksanderSklorz, line 255: We shouldn't mock functions of class which we actually test. So here you test SearchInputComponent and you can't mock that component's functions. We can mock only functions which come outside of component like for example services functions. 
If we mock out functions of class which we test then these tests don't test anything because they have mocked results so we don't know if component really works like expected. Please try to refactor it a bit to avoid mocking functions of class which you test and use that approach for each function which comes from that component.

PR #4265, opened 2024-11-28
[ACS-9012] Truncate text in saved searches description

MichalKinas:
Add text truncation in saved searches description column, added missing docs in `extension.schema.json` and fixed one small bug in unique saved search name validator. 
https://hyland.atlassian.net/browse/ACS-9012

### DIFF
--- a/extension.schema.json
+++ b/extension.schema.json
@@ -590,6 +590,14 @@
           "description": "Toggles resizable state of the column",
           "type": "boolean"
         },
+        "draggable": {
+          "description": "Toggles draggable state of the column",
+          "type": "boolean"
+        },
+        "isHidden": {
+          "description": "Toggles hidden state of the column",
+          "type": "boolean"
+        },
         "template": {
           "description": "Column template id",
           "type": "string"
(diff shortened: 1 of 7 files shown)


dominikiwanekhyland (approved):

nikita-web-ua (approved):

AleksanderSklorz:
@MichalKinas I see that actually only description column is truncated but if you check task's title it says about "name" column. Should we truncate name column as well?
image

AleksanderSklorz:
Should we truncate entry in sidebar as well? Actually it doesn't look good because text goes outside of available space: 
image

MichalKinas:
> @MichalKinas I see that actually only description column is truncated but if you check task's title it says about "name" column. Should we truncate name column as well? image

Yes, I think this was misunderstood since I think the original idea was to truncate description column but I think for name it also makes a lot of sense

MichalKinas:
> Should we truncate entry in sidebar as well? Actually it doesn't look good because text goes outside of available space: image

Good point, fixed

AleksanderSklorz (approved):

PR #4341, opened 2025-01-17
[ACS-9154] Fix unstable tests in ACA after migration to local ACS

datguychen:
https://hyland.atlassian.net/browse/ACS-9154

### DIFF
--- a/.github/actions/run-e2e-playwright/action.yml
+++ b/.github/actions/run-e2e-playwright/action.yml
@@ -31,3 +31,11 @@ runs:
 
         echo "Running playwright tests with options ${{ inputs.options }}"
         npx nx run ${{ inputs.options }}-e2e:e2e
+
+    - name: Upload E2Es results
+      if: always()
+      uses: actions/upload-artifact@v4
+      with:
+        name: ${{ inputs.artifact-name }}
+        path: |
+          test-results/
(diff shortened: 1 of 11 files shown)


datguychen:
Waiting for config.json file to be deleted from saved search to rebase the branch and see how the tests perform then.

AleksanderSklorz:
I have one question - there is still excluded one test in e2e/playwright/search/exclude.tests.json: C277736, what with this one?

datguychen:
> I have one question - there is still excluded one test in e2e/playwright/search/exclude.tests.json: C277736, what with this one? @AleksanderSklorz 

This test checks the sorting of the dataTable by Modified date. Currently, the files uploaded during the test run have the same Modified date because they are uploaded simultaneously, resulting in inconsistent test outcomes. To re-enable this test, we need to consider a different approach.

At present, files are uploaded when the environment is created, which could be used for the Modified date comparison. However, we plan to switch the tests from the acadev environment to locally built environments with local ACS. This transition could be challenging due to the time difference between environment build (when environment files are created) and test run (when test files are created).

We need more time to evaluate how the tests perform with the locally built environments to determine if this approach would work for this e2e test.

I will create a ticket for this e2e test and update it in the exclude file in my next PR.

AleksanderSklorz (approved):

MichalKinas (approved):

PR #4140, opened 2024-09-25
[ACS-8694] Cleanup of visibility rules for extensions in ACA

swapnil-verma-gl:
**JIRA ticket link or changeset's description**

https://hyland.atlassian.net/browse/ACS-8694

Updated ACA extensions config to use array based visibility rules instead of string based. Also removed various unused rules

### **_This is a breaking change_**

PR #4351, opened 2025-01-22
[ACS-9158] Remove 'View Details' button from node Details page

nikita-web-ua:
https://hyland.atlassian.net/browse/ACS-9158

The "View Details" button has been removed from the Details page because the button's logic causes the page to redirect on refresh and the link cannot be accessed directly. At the same time "Reduce panel" button is retained and performs the same action.

Screenshot 2025-01-22 at 09 41 16

### DIFF
--- a/projects/aca-content/src/lib/components/details/details.component.ts
+++ b/projects/aca-content/src/lib/components/details/details.component.ts
@@ -23,10 +23,10 @@
  */
 
 import { Component, OnDestroy, OnInit, ViewEncapsulation } from '@angular/core';
-import { ActivatedRoute, NavigationStart } from '@angular/router';
+import { ActivatedRoute } from '@angular/router';
 import { ContentApiService, PageComponent, PageLayoutComponent, ToolbarComponent } from '@alfresco/aca-shared';
 import { NavigateToFolder, NavigateToPreviousPage, SetSelectedNodesAction } from '@alfresco/aca-shared/store';
-import { merge, Subject } from 'rxjs';
+import { Subject } from 'rxjs';
 import { BreadcrumbComponent, ContentService, NodesApiService, PermissionListComponent } from '@alfresco/adf-content-services';
 import { CommonModule } from '@angular/common';
 import { TranslateModule } from '@ngx-translate/core';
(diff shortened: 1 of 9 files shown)


projects/aca-content/src/lib/components/details/details.component.ts
```diff
@@ -110,20 +110,6 @@ export class DetailsComponent extends PageComponent implements OnInit, OnDestroy
       .subscribe((aspectActions) => {
         this.aspectActions = aspectActions;
       });
-    this.infoDrawerOpened$
-      .pipe(
-        first((opened) => !opened),
-        takeUntil(
-          merge(
-            this.onDestroy$,
```
AleksanderSklorz, line 118: onDestroy$ can be removed as now it's unused

nikita-web-ua: removed, ty

AleksanderSklorz (approved):

dominikiwanekhyland (approved):

projects/aca-shared/rules/src/app.rules.ts
```diff
@@ -461,7 +461,7 @@ export const canToggleSharedLink = (context: RuleContext): boolean =>
  * @param context Rule execution context
  */
 export const canShowInfoDrawer = (context: RuleContext): boolean =>
-  [hasSelection(context), navigation.isNotLibraries(context), navigation.isNotTrashcan(context)].every(Boolean);
+  [hasSelection(context), navigation.isNotLibraries(context), navigation.isNotTrashcan(context), navigation.isNotDetails(context)].every(Boolean);
```
swapnil-verma-gl, line 464: In the context of the [visibility rules PR](https://github.com/Alfresco/alfresco-content-app/pull/4140), the canShowInfoDrawer rule and method will be removed in favour of individual array based rules. So the reference of this rule in app.extensions.json at line num [498](https://github.com/Alfresco/alfresco-content-app/pull/4140/files#diff-51260302282a0c533f07b4cfbd078addc3e03e536a50f7532298d479a2bb10afL498)  and [1057](https://github.com/Alfresco/alfresco-content-app/pull/4140/files#diff-51260302282a0c533f07b4cfbd078addc3e03e536a50f7532298d479a2bb10afL1057) will no longer be present once that PR is merged, and would be replaced with separate arrays as can be seen in that PR. Can you please update the code here so that instead of adding the `navigation.isNotDetails(context)` call here, its added as an element of the visible array in the json file itself. I see that this rule is already registered in [aca-content.module.ts](https://github.com/Alfresco/alfresco-content-app/pull/4351/files#diff-dd40e09f69d4680d3ac6532ad0a670b76f59110637e8c2c69509d11ab033bec5R226) in this PR, so you only need to modify the rule in the json.  The change should be something similar to 
```
"visible": [ "canShowInfoDrawer", "app.navigation.isNotDetails" ]
```

MichalKinas (approved):

projects/aca-shared/rules/src/app.rules.ts
nikita-web-ua: done

PR #4359, opened 2025-01-29
[ACS-9083][ADW] Comment creation is available after document is retained (declared as record)

dominikiwanekhyland:
https://hyland.atlassian.net/browse/ACS-9083

### DIFF
--- a/projects/aca-content/src/lib/components/info-drawer/comments-tab/comments-tab.component.ts
+++ b/projects/aca-content/src/lib/components/info-drawer/comments-tab/comments-tab.component.ts
@@ -47,8 +47,8 @@ export class CommentsTabComponent implements OnInit {
   canUpdateNode = false;
 
   constructor(
-    private permission: NodePermissionService,
-    @Optional() private externalPermissionNodeService: ExternalNodePermissionCommentsTabService
+    private readonly permission: NodePermissionService,
+    @Optional() private readonly externalPermissionNodeService: ExternalNodePermissionCommentsTabService
   ) {}
 
   ngOnInit(): void {


AleksanderSklorz:
@dominikiwanekhyland I found one more problem - when you make file as record in ADW then you can't add comment. But if you open ACA then in ACA you can still add comment for that file which was made as record in ADW which should not be possible I believe? cc: @MichalKinas what do you think?

AleksanderSklorz:
@dominikiwanekhyland please fix sonar issues: 
image

projects/aca-content/src/lib/components/info-drawer/comments-tab/comments-tab.component.spec.ts
```diff
@@ -107,4 +113,30 @@ describe('CommentsTabComponent', () => {
       expect(checked).toContain('update');
     });
   });
+
+  describe('ExternalNodePermissionCommentsTabService', () => {
+    it('should return false if node permissions are correct and external service is not allowing it', async () => {
```
AleksanderSklorz, line 118: can you rename title of test a bit because in expect you don't check directly result of service, instead you check canUpdateNode so please change a bit title to describe what is checked?

projects/aca-content/src/lib/components/info-drawer/comments-tab/comments-tab.component.spec.ts
```diff
@@ -107,4 +113,30 @@ describe('CommentsTabComponent', () => {
...
+        id: 'test-node-id',
+        isFile: true,
+        isFolder: false
+      } as Node;
+      fixture.detectChanges();
+      await fixture.whenStable();
+      expect(component.canUpdateNode).toBe(false);
+    });
+
+    it('should return true if node permissions are correct and external service is allowing it', async () => {
```
AleksanderSklorz, line 130: can you rename title of test a bit because in expect you don't check directly result of service, instead you check canUpdateNode so please change a bit title to describe what is checked?

AleksanderSklorz (approved):

PR #4372, opened 2025-02-04
[ACS-9211] Update license headers

nikita-web-ua:
https://hyland.atlassian.net/browse/ACS-9211

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -59,7 +59,7 @@
           "error",
           [
             "/*!",
-            " * Copyright © 2005-2024 Hyland Software, Inc. and its affiliates. All rights reserved.",
+            " * Copyright © 2005-2025 Hyland Software, Inc. and its affiliates. All rights reserved.",
             " *",
             " * Alfresco Example Content Application",
             " *",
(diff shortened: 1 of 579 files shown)


AleksanderSklorz:
@nikita-web-ua on ACA login screen I can see 2024 year. Can we fix that too?
image

nikita-web-ua:
> @nikita-web-ua on ACA login screen I can see 2024 year. Can we fix that too? image

sure, there is a separate ticket for it https://hyland.atlassian.net/browse/ACS-9213

AleksanderSklorz:
> > @nikita-web-ua on ACA login screen I can see 2024 year. Can we fix that too? image
> 
> sure, there is a separate ticket for it https://hyland.atlassian.net/browse/ACS-9213

@nikita-web-ua I think the ticket which you attached is when we run it on remote env. But when run locally we need to update all json files for translations because there is hardcoded year: 
image

When I updated them then I can see it's updated on login page too: 
image

nikita-web-ua:
> 
> @nikita-web-ua I think the ticket which you attached is when we run it on remote env. But when run locally we need to update all json files for translations because there is hardcoded year: image
> 
> When I updated them then I can see it's updated on login page too: image

For ADW it is already up to date, for ACA we need to update it as you wrote in e18n file because it uses other type of auth as part of https://hyland.atlassian.net/browse/ACS-9213 ticket.

AleksanderSklorz (approved):

PR #4383, opened 2025-02-07
[ACS-9229] a11y testing: Search Page / Elements must only use supported ARIA attributes

dominikiwanekhyland:
https://hyland.atlassian.net/browse/ACS-9229


nikita-web-ua:
I think this change might populate another issue, can you check this one please
Screenshot 2025-02-07 at 11 48 55

dominikiwanekhyland:
> I think this change might populate another issue, can you check this one please Screenshot 2025-02-07 at 11 48 55


@nikita-web-ua yes, but it's just how this component is designed, this nesting is not possible to be avoided I guess. We already have there interactive elements nesting and just because it's div element it's not highlighted by this axe devtools. So I guess it's better to solve at least one problem with that which is clickable div element

nikita-web-ua (approved):

PR #4385, opened 2025-02-07
AAE-30744 Removed temporary fix for dialog padding

wojd0:

https://hyland.atlassian.net/browse/AAE-30744

In Angular 17 issue with dialog paddings seems to be resolved, this modification can be removed to prevent extra padding being added.

### DIFF
--- a/projects/aca-content/src/lib/ui/theme.scss
+++ b/projects/aca-content/src/lib/ui/theme.scss
@@ -145,15 +145,11 @@ mat-slide-toggle {
     --mdc-dialog-supporting-text-color: var(--theme-primary-text);
   }
 
-  #{$mat-dialog-surface} {
-    padding: 24px;
+  #{$mat-button} {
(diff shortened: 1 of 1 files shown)


dominikiwanekhyland (approved):

nikita-web-ua:
Is this change still required? I guess it will break some of our dialogs:
Screenshot 2025-02-17 at 13 51 41

wojd0:
@nikita-web-ua Angular v17 changed something that caused all the dialogs to have double padding inside, I'll check this dialog for issues, but I think issues like this should be resolved per-dialog, not by changing global styles for all

wojd0:
after review this change will cause too much disruption for all dialogs in the app, I'll investigate if this issue can be resolved from the [alfresco/hxp-frontend-apps](https://github.com/alfresco/hxp-frontend-apps) side

PR #4394, opened 2025-02-13
[ACS-9266] Make notification and user menu accessible with keyboard

nikita-web-ua:
https://hyland.atlassian.net/browse/ACS-9266
https://hyland.atlassian.net/browse/ACS-9234


AleksanderSklorz:
@nikita-web-ua I found one difference in behavior for notifications menu and settings menu - when you open settings menu then we can close it also using ESC key but for notifications menu we can't close it using ESC, it can be closed only when hit Enter etc. Can we allow to use ESC too because it's often used key for closing?

https://github.com/user-attachments/assets/f3dbee43-b23c-43c6-b2dd-6d504098620e

projects/aca-content/src/lib/components/sidenav/sidenav.component.scss
```diff
@@ -140,5 +140,9 @@
         border-bottom: 0;
       }
     }
+
+    .app-expand-menu .aca-expansion-panel {
```
AleksanderSklorz, line 144: we have aca-expansion-panel already in 46th line, can it be reused and styles can be applied there?

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    fixture.detectChanges();
+  });
+
+  it('should merge nested menu items to MatMenu in ngAfterViewInit', () => {
+    spyOn(component.menu, 'ngAfterContentInit').and.callThrough();
+    testingUtils.getByDirective(MatMenuTrigger).nativeElement.click();
+    fixture.detectChanges();
+    component.ngAfterViewInit();
+
+    expect(component.menu._allItems.length).toBe(2);
```
AleksanderSklorz, line 58: the same comment like in ADF - instead of referring to private field can we somehow use harness class? https://github.com/Alfresco/alfresco-ng2-components/pull/10647#discussion_r1956039856

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.ts
```diff
@@ -50,9 +50,28 @@ export class UserMenuComponent implements OnInit {
...
+
   ngOnInit() {
     if (this.data?.items) {
       this.data.items.sort((a, b) => a.order - b.order);
     }
   }
+
+  ngAfterViewInit() {
+    const menuItems: MatMenuItem[] = [];
+    this.toolbarMenuItems.forEach((toolbarMenuItem: ToolbarMenuItemComponent) => {
```
AleksanderSklorz, line 67: same comment like in ADF - we can replace forEach with filter which will cause a shorter code: 
https://github.com/Alfresco/alfresco-ng2-components/pull/10647#discussion_r1956054188

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.ts
```diff
@@ -50,9 +50,28 @@ export class UserMenuComponent implements OnInit {
...
+      if (toolbarMenuItem.menuItem !== undefined) {
+        menuItems.push(toolbarMenuItem.menuItem);
+      }
+    });
+    const menuItemsQueryList: QueryList<MatMenuItem> = new QueryList<MatMenuItem>();
```
AleksanderSklorz, line 72: no need to set type at the left side when we assigning inline

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.ts
nikita-web-ua: done

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.ts
nikita-web-ua: done

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.spec.ts
nikita-web-ua: I do want to check this private property as part of this unit test, with harness I suppose we can only check the number of matItems, which was correct before the fix as well.

nikita-web-ua:
> @nikita-web-ua I found one difference in behavior for notifications menu and settings menu - when you open settings menu then we can close it also using ESC key but for notifications menu we can't close it using ESC, it can be closed only when hit Enter etc. Can we allow to use ESC too because it's often used key for closing?
> 
>  Screen.Recording.2025-02-14.at.12.56.08.mov

should be fixed now

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.ts
```diff
@@ -50,9 +50,26 @@ export class UserMenuComponent implements OnInit {
...
+  toolbarMenuItems: QueryList<ToolbarMenuItemComponent>;
+
   ngOnInit() {
     if (this.data?.items) {
       this.data.items.sort((a, b) => a.order - b.order);
     }
   }
+
+  ngAfterViewInit() {
+    const menuItems: MatMenuItem[] = this.toolbarMenuItems
```
AleksanderSklorz, line 66: no need to set type for menuItems (at left side) because it's set automatically based on filter function

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.ts
```diff
@@ -50,9 +50,26 @@ export class UserMenuComponent implements OnInit {
...
+      .map((toolbarMenuItem: ToolbarMenuItemComponent) => toolbarMenuItem.menuItem)
```
AleksanderSklorz, line 67: no need to set type for toolbarMenuItem because it's set automatically based on tollbarMenuItems type

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.ts
```diff
@@ -50,9 +50,26 @@ export class UserMenuComponent implements OnInit {
...
+      .filter((menuItem: MatMenuItem) => menuItem !== undefined);
```
AleksanderSklorz, line 68: no need to set type for menuItem because it's set automatically based on previous statement

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.ts
AleksanderSklorz, line 67: Do we need to use map? I guess we can use only filter like: 
.filter((toolbarMenuItem) => toolbarMenuItem.menuItem !== undefined);

When using both we need to iterate toolbarMenuItems twice, but when we use only filter then we will iterate once

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.ts
nikita-web-ua: removed

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.ts
nikita-web-ua: removed

AleksanderSklorz (approved):

projects/aca-content/src/lib/components/sidenav/user-menu/user-menu.component.ts
nikita-web-ua: removed

PR #4448, opened 2025-03-14
Angular 18 upgrade

dominikiwanekhyland:


package.json
```diff
@@ -30,64 +30,66 @@
...
+    "@angular/router": "18.2.9",
     "@fontsource/open-sans": "^5.1.0",
-    "@mat-datetimepicker/core": "13.0.2",
-    "@ngrx/effects": "17.0.1",
-    "@ngrx/router-store": "17.0.1",
-    "@ngrx/store": "17.0.1",
-    "@ngrx/store-devtools": "17.0.1",
+    "@mat-datetimepicker/core": "14.0.0",
+    "@ngrx/effects": "18.1.1",
+    "@ngrx/operators": "^18.0.0",
```
AleksanderSklorz, line 52: Rest of ngrx versions has not ^ so should we remove ^ here as well to have the same?

migrations.json
DenysVuika, line 1: this file should be removed after migrations, it should not be committed

AleksanderSklorz (approved):

AleksanderSklorz:
@dominikiwanekhyland we noticed that we need to update also node version in Dockerfile and I see this PR updates node version in .nvmrc file so could you please also update node in Dockerfile like in this PR?
https://github.com/Alfresco/alfresco-content-app/pull/4532

PR #4461, opened 2025-03-18
[ACS-6412] remove excluded test for template issue

akashrathod28:
https://hyland.atlassian.net/browse/ACS-6412


projects/aca-playwright-shared/src/page-objects/components/dialogs/create-from-template-dialog-component.ts
```diff
@@ -27,16 +27,16 @@ import { BaseComponent } from '../base.component';
...
 
   constructor(page: Page) {
     super(page, CreateFromTemplateDialogComponent.rootElement);
   }
 
   cancelButton = this.getChild('[data-automation-id="create-from-template-dialog-cancel-button"]');
   createButton = this.getChild('[data-automation-id="create-from-template-dialog-create-button"]');
-  getDialogTitle = (text: string) => this.getChild('.mat-dialog-title', { hasText: text });
-  getDialogLabel = (text: string) => this.getChild('label', { hasText: text });
+  getDialogTitle = (text: string) => this.getChild('.mat-mdc-dialog-title', { hasText: text });
```
swapnil-verma-gl, line 38: I'm curious. We had gone through a lot of effort trying to reduce/remove occurrences of mat selectors from our codebase. Should we think about doing the same for E2Es as well? It would prevent E2E breakage if in the future, angular material decides to do the same again. Also I think this change was needed because of that reason only

@MichalKinas what do you think?

akashrathod28: I added automation id to few elements here.

swapnil-verma-gl: Yes, this looks much better. Thanks!

swapnil-verma-gl (approved):

AleksanderSklorz (approved):

PR #4556, opened 2025-05-08
[ACS-9536] fix failing unit tests

g-jaskowski:
### Ticket 
[ACS-9536 Unit tests are randomly failing sometimes on ACA](https://hyland.atlassian.net/browse/ACS-9536)

### Changes

- Reset overridden selectors affecting global store state
- Use correct testing module in other suite

### Testing

- Run tests locally multiple times
image

- Run tests on PR workflow multiple times
[5 attempts of Test All step skipping cache](https://github.com/Alfresco/alfresco-content-app/actions/runs/14907556727)

### DIFF
--- a/projects/aca-content/viewer/src/lib/components/preview/preview.component.spec.ts
+++ b/projects/aca-content/viewer/src/lib/components/preview/preview.component.spec.ts
@@ -24,22 +24,15 @@
 
 import { Router, ActivatedRoute } from '@angular/router';
 import { TestBed, ComponentFixture, fakeAsync, tick } from '@angular/core/testing';
-import { AuthenticationService } from '@alfresco/adf-core';
-import { UploadService, NodesApiService, DiscoveryApiService } from '@alfresco/adf-content-services';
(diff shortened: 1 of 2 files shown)


projects/aca-content/viewer/src/lib/components/preview/preview.component.spec.ts
```diff
@@ -56,12 +49,7 @@ describe('PreviewComponent', () => {
 
   beforeEach(() => {
     TestBed.configureTestingModule({
-      imports: [LibTestingModule, AcaViewerModule],
-      providers: [
-        { provide: DocumentBasePageService, useValue: DocumentBasePageServiceMock },
```
AleksanderSklorz, line 61: I see one problem - if you check AppTestingModule then it contains some mock for rest services but for DocumentBasePageService it has: 
```ts
{ provide: DocumentBasePageService, useExisting: ContentManagementService }
```
which means it uses real (not mocked) service for this one service after you removed mock for that from these tests. We should not use real services for tests, we should to use mocks.

AleksanderSklorz (approved):

PR #4577, opened 2025-05-20
[ACS-8746] use notification service instead of store, yellow warnings, warning on no library permission

g-jaskowski:
### Tickets:
https://hyland.atlassian.net/browse/ACS-8746
https://hyland.atlassian.net/browse/ACS-9344 - second point

### Changes
- Use AFD NotificationService instead of store actions, remove SnackbarActions
- Change warning notification color to yellow (according to feedback from Satori DS team)
- Use warning notification on no library permission instead of error
- Few eslint errors fix


projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
```diff
@@ -54,6 +54,7 @@ describe('SearchComponent', () => {
   const searchRequest = {} as SearchRequest;
   let params: BehaviorSubject<any>;
   let showErrorSpy: jasmine.Spy;
+  let showInfoSpy: jasmine.Spy;
```
AleksanderSklorz, line 57: I know that above it's written like that but Spy is generic type so ideally you should set types for mocked function inside ```<>```. You can search for examples in other spec files how to set this

projects/aca-content/src/lib/store/effects/library.effects.ts
```diff
@@ -120,18 +120,16 @@ export class LibraryEffects {
                   this.store.dispatch(new NavigateRouteAction([route, id]));
                 },
                 (error: HttpErrorResponse) => {
-                  let errorTranslationKey: string;
                   switch (error.status) {
                     case 403:
-                      errorTranslationKey = 'APP.BROWSE.LIBRARIES.ERRORS.LIBRARY_NO_PERMISSIONS';
+                      this.notificationService.showWarning('APP.BROWSE.LIBRARIES.ERRORS.LIBRARY_NO_PERMISSIONS');
```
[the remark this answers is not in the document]
g-jaskowski: It is the notification specified in [ACS-9344](https://hyland.atlassian.net/browse/ACS-9344)

AleksanderSklorz:
there is black snackbar when use move option. Also for 'copy' option the same happens  

https://github.com/user-attachments/assets/c10ad17f-5bd6-482b-bf29-98a487c08f53

AleksanderSklorz:
@g-jaskowski please resolve sonar issues: 
image

projects/aca-content/src/lib/services/content-management.service.spec.ts
```diff
@@ -604,9 +634,9 @@ describe('ContentManagementService', () => {
       subject = new Subject<string>();
       spyOn(nodeActions, 'moveNodes').and.returnValue(subject);
 
-      spyOn(notificationService, 'openSnackMessageAction').and.returnValue({
-        onAction: () => of(null)
-      } as MatSnackBarRef<SimpleSnackBar>);
+      openSnackMessageActionSpy.and.returnValue({
+        onAction: () => of(new Subject<void>())
+      } as unknown as MatSnackBarRef<SimpleSnackBar>);
```
AleksanderSklorz: You don't need unknown. I guess you had issue because there is assigned incorrect value to onAction above. If you check what's type of onAction result in MatSnackBarRef then it expects ```Observable<void>``` but you assign ``` of(new Subject<void>()) ``` which is ```Observable<Subject<void>>``` type so types are incorrect. You can use following instead:

```ts
openSnackMessageActionSpy.and.returnValue({
        onAction: () => of(undefined)
      } as MatSnackBarRef<SimpleSnackBar>);

File: projects/aca-content/src/lib/services/content-management.service.ts [hunk 4]
```diff
@@ -71,6 +65,13 @@ interface RestoredNode {
   statusCode?: number;
 }
 
+interface SnackbarMessageData {
+  key: string;
+  params: any;
```
AleksanderSklorz: I think you can use following type instead of any because it will limit a bit possible values (for example numbers/strings will be disallowed): 
```ts
params: { [key: string]: any }

File: projects/aca-content/src/lib/services/content-management.service.ts [hunk 5]
```diff
@@ -895,27 +908,29 @@ export class ContentManagementService {
...
+    const messageData = this.getRestoreMessageData(status);
 
-    if (message) {
-      if (status.oneSucceeded && !status.someFailed) {
-        const isSite = this.isSite(status.success[0].entry);
-        const path: PathInfo = status.success[0].entry.path;
-        const parent = path.elements[path.elements.length - 1];
-        const route = isSite ? ['/libraries', parent.id] : ['/personal-files', parent.id];
+    if (messageData) {
+      const translatedMessage = this.translation.instant(messageData.key, messageData.params);
```
AleksanderSklorz: could you please set type to string for that const because it's treated like any?image

projects/aca-content/src/lib/services/content-management.service.ts
```diff
@@ -895,27 +908,29 @@ export class ContentManagementService {
...
+      const action = messageData.userActionLabel ? this.translation.instant(messageData.userActionLabel) : '';
```
AleksanderSklorz: same here, please set type: 
image

projects/aca-content/src/lib/services/content-management.service.ts
```diff
@@ -927,39 +942,54 @@ export class ContentManagementService {
...
+      };
     }
 
     if (status.oneFailed && status.fail[0].statusCode) {
       if (status.fail[0].statusCode === 409) {
-        return new SnackbarErrorAction('APP.MESSAGES.ERRORS.TRASH.NODES_RESTORE.NODE_EXISTS', {
-          name: status.fail[0].entry.name
-        });
+        return {
+          key: 'APP.MESSAGES.ERRORS.TRASH.NODES_RESTORE.NODE_EXISTS',
```
AleksanderSklorz: only difference is key so will be better: 
```ts
return {
          key: status.fail[0].statusCode === 409 ? 'APP.MESSAGES.ERRORS.TRASH.NODES_RESTORE.NODE_EXISTS' : 'APP.MESSAGES.ERRORS.TRASH.NODES_RESTORE.GENERIC',
          params: { name: status.fail[0].entry.name },
          type: 'error'
        };
```
It will avoid repeating code.

nikita-web-ua (approved):

PR #4604, opened 2025-06-06
[ACS-9749] [E2E] Automated 5513 and 5516

datguychen:
https://hyland.atlassian.net/browse/ACS-9749


projects/aca-playwright-shared/src/api/nodes-api.ts
```diff
@@ -290,7 +299,7 @@ export class NodesApi {
     });
   }
 
-  async setGranularPermission(nodeId: string, inheritPermissions: boolean = false, username: string, role: string): Promise<NodeEntry | null> {
+  async setGranularPermission(nodeId: string, username: string, role: string, inheritPermissions: boolean = false): Promise<NodeEntry | null> {
```
AleksanderSklorz: I know it was not your change but no need to set type for inheritPermissions because we assign false inline to it.

PR #4610, opened 2025-06-09
[ACS-5503] Folder rules rule update with correct error message when definition is invalid

MichalKinas:
https://hyland.atlassian.net/browse/ACS-5503

### DIFF
--- a/projects/aca-content/folder-rules/src/services/folder-rules.service.spec.ts
+++ b/projects/aca-content/folder-rules/src/services/folder-rules.service.spec.ts
@@ -23,7 +23,7 @@
  */
 
 import { TestBed } from '@angular/core/testing';
-import { CoreTestingModule } from '@alfresco/adf-core';
+import { CoreTestingModule, NotificationService } from '@alfresco/adf-core';
 import { of } from 'rxjs';
 import { FolderRulesService } from './folder-rules.service';
 import {
@@ -42,6 +42,7 @@ import { AlfrescoApiService, AlfrescoApiServiceMock } from '@alfresco/adf-conten
 
 describe('FolderRulesService', () => {
   let folderRulesService: FolderRulesService;
+  let notificationService: NotificationService;
 
   let callApiSpy: jasmine.Spy;
 
@@ -61,6 +62,7 @@ describe('FolderRulesService', () => {
     });
 
     folderRulesService = TestBed.inject<FolderRulesService>(FolderRulesService);
+    notificationService = TestBed.inject<NotificationService>(NotificationService);
 
     callApiSpy = spyOn<any>(folderRulesService, 'callApi');
   });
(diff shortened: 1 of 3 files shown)


projects/aca-content/folder-rules/src/services/folder-rules.service.spec.ts
```diff
@@ -61,6 +62,7 @@ describe('FolderRulesService', () => {
     });
 
     folderRulesService = TestBed.inject<FolderRulesService>(FolderRulesService);
+    notificationService = TestBed.inject<NotificationService>(NotificationService);
```
AleksanderSklorz: no need to set type in ```<>``` as it's detected automatically

projects/aca-content/folder-rules/src/services/folder-rules.service.spec.ts
```diff
@@ -148,6 +150,17 @@ describe('FolderRulesService', () => {
     expect(result).toEqual(mockedRule);
   });
 
+  it('should display error message and revert enabled state when updating rule fails', async () => {
+    callApiSpy
```
AleksanderSklorz, line 154: looks like callApiSpy spies private function which is not good practice. Instead we can spy on for example following function: 
this.apiService.getInstance().contentPrivateClient.callApi(path, httpMethod, ...params)

because callApi from contentPrivateClient is public

projects/aca-content/folder-rules/src/services/folder-rules.service.spec.ts
MichalKinas: Done

projects/aca-content/folder-rules/src/services/folder-rules.service.spec.ts
MichalKinas: I'm afraid this won't work since `apiService` is private for this service and can't be accessed outside of the class

AleksanderSklorz: @MichalKinas not exactly - we can do like for all other tests - we can use TestBed.inject to get reference to this service and then spy public function. In that way we don't need to access private member as we can access reference by injection

MichalKinas: Hmm I can see that this approach is also used in other places, maybe it would be better to create a task for it and change it to proper approach everywhere? Maybe it would be also good to check if folder rules apis are still private, if those are made public we could refactor that too, what do you think?

AleksanderSklorz: sounds good to me

AleksanderSklorz (approved):

projects/aca-content/folder-rules/src/services/folder-rules.service.spec.ts
MichalKinas: Created https://hyland.atlassian.net/browse/ACS-9758

AleksanderSklorz: thank you

PR #4613, opened 2025-06-10
[ACS-9761] ACA run e2e with env

akashrathod28:
https://hyland.atlassian.net/browse/ACS-9761

### DIFF
--- a/.github/workflows/run-e2e-with-env.yml
+++ b/.github/workflows/run-e2e-with-env.yml
@@ -21,14 +21,14 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  BASE_URL: ${{ secrets.PIPELINE_ENV_URL }}
+  BASE_URL: ${{ inputs.url }}
   ADMIN_EMAIL: ${{ secrets.PIPELINE_ADMIN_USERNAME }}
   ADMIN_PASSWORD: ${{ secrets.PIPELINE_ADMIN_PASSWORD }}
   HR_USER: ${{ secrets.HR_USER }}
   HR_USER_PASSWORD: ${{ secrets.HR_USER_PASSWORD }}
   SCREENSHOT_USERNAME: ${{ secrets.SCREENSHOT_USERNAME }}
   SCREENSHOT_PASSWORD: ${{ secrets.SCREENSHOT_PASSWORD}}
-  PLAYWRIGHT_E2E_HOST: ${{ secrets.PLAYWRIGHT_E2E_HOST }}
+  PLAYWRIGHT_E2E_HOST: ${{ inputs.url }}/aca/#/
   GH_BUILD_NUMBER: ${{ github.run_id }}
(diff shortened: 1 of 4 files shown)


.github/workflows/run-e2e-with-env.yml
```diff
@@ -0,0 +1,175 @@
...
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Branch to test'
+        required: true
+        default: 'develop'
+      url:
+        description: 'Application URL'
+        required: true
+        default: 'https://acadev.envalfresco.com'
```
MichalKinas: This leaks info about our testing env, the default should be something else or just empty

.github/workflows/run-e2e-with-env.yml
```diff
@@ -0,0 +1,175 @@
+name: "Run e2e with Env"
+description: "Run e2e tests with environment variables"
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
```
MichalKinas: This seems unused why do we need this?

.github/workflows/run-e2e-with-env.yml
```diff
@@ -0,0 +1,175 @@
...
+        run: npx playwright install chromium
+
+      - uses: ./.github/actions/run-e2e-playwright
+        with:
+          options: "${{ matrix.e2e-suites.name }}"
+          artifact-name: ${{ matrix.e2e-suites.name }}
+          test-runner: playwright
+      - uses: ./.github/actions/after-e2e
+
+  finalize:
```
MichalKinas: I don't think this step is needed for E2Es, it only make sense for pull requests

.github/workflows/run-e2e-with-env.yml
akashrathod28: Sure will remove this.

.github/workflows/run-e2e-with-env.yml
akashrathod28: Plan is to run test on custom branch with older tag. ex. if we run ACS 23.5 with ADW 5.3
let me know if this will will be correct approach.

.github/workflows/run-e2e-with-env.yml
akashrathod28: Yes, I will remove Finalise step.

.github/workflows/run-e2e-with-env.yml
akashrathod28: I am removing this step for now. will update workflow once we get such scenario to run.

MichalKinas: Yeah so it's not how it works, adding the branch input will not do the trick here, the idea is that you instead trigger the workflow on specific branch, so for example if you have a branch with ACA 5.3 you have to migrate this code that you've added there and then trigger the workflow on that branch

MichalKinas (approved):

PR #4632, opened 2025-06-23
[ACS-9777] [E2E] Added e2es for 5522 and 5523

datguychen:
https://hyland.atlassian.net/browse/ACS-9777

### DIFF
--- a/e2e/playwright/info-drawer/src/tests/comments.e2e.ts
+++ b/e2e/playwright/info-drawer/src/tests/comments.e2e.ts
@@ -32,7 +32,8 @@ import {
   NodesApi,
   FileActionsApi,
   SitesApi,
-  MyLibrariesPage
+  MyLibrariesPage,
+  timeouts
 } from '@alfresco/aca-playwright-shared';
 import { Site } from '@alfresco/js-api';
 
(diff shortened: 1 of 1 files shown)


e2e/playwright/info-drawer/src/tests/comments.e2e.ts
```diff
@@ -140,3 +151,77 @@ test.describe('Info Drawer - Comments', () => {
...
+  });
+
+  test('[XAT-5522] Comments: Consumer user does not see the multiline field and Add button', async ({ myLibrariesPage }) => {
+    await navigateToFolderAndOpenCommentsTab(myLibrariesPage, siteEntry5522.guid, folderName5522);
+    await expect(myLibrariesPage.infoDrawer.commentsHeader).toBeVisible();
+    await expect(myLibrariesPage.infoDrawer.commentInputField).toBeHidden();
+    await expect(myLibrariesPage.infoDrawer.addCommentButton).toBeHidden();
+  });
+
+  test('[XAT-5523] Contributor can NOT add comments', async ({ myLibrariesPage }) => {
```
AleksanderSklorz: Do we want this test? I see there is raised following bug jira: https://hyland.atlassian.net/browse/ACA-4697 which says that contributor should be able to add comment. 

We don't need that test OR bug task is invalid and should  be closed. What do you think @datguychen?

adamswiderskipw: I changed the test case from CAN to CAN'T because the bug is 4 years old and its not planned to be fixed in the future. On a second thought, I think I should change it back to CAN and maybe exclude the test I've created. Let me know what you think.

adamswiderskipw: https://hyland.atlassian.net/browse/ACA-1770

AleksanderSklorz: To avoid confusion I think the best would be - write test where user CAN do it as "CAN" is really expected behavior. Then exclude that test and link jira's ticket to exclusion. So then if anyone will fix it in future we can just enable test and everything will be working as expected. Thanks that we won't need to refactor test in future + it will be consistent with expected behavior mentioned in jira.

AleksanderSklorz (approved):

PR #4662, opened 2025-07-09
ACS-9789 Fixing Issues on metadata sidebar

Isoumyajit:
https://hyland.atlassian.net/browse/ACS-9789

### DIFF
--- a/app/src/app/app.module.ts
+++ b/app/src/app/app.module.ts
@@ -57,6 +57,7 @@ import { SHELL_APP_SERVICE, SHELL_AUTH_TOKEN, ShellModule } from '@alfresco/adf-
 import { StoreDevtoolsModule } from '@ngrx/store-devtools';
 import { APP_ROUTES } from './app.routes';
 import { MatIconRegistry } from '@angular/material/icon';
+import { MAT_FORM_FIELD_DEFAULT_OPTIONS } from '@angular/material/form-field';
 
 registerLocaleData(localeFr);
 registerLocaleData(localeDe);
@@ -111,6 +112,10 @@ registerLocaleData(localeSv);
         name: 'app',
         source: 'assets'
       }
+    },
+    {
+      provide: MAT_FORM_FIELD_DEFAULT_OPTIONS,
+      useValue: { appearance: 'fill', floatLabel: 'always' }
     }
   ],
   declarations: [AppComponent],
--- a/app/src/app/app.module.ts
+++ b/app/src/app/app.module.ts
@@ -57,7 +57,6 @@ import { SHELL_APP_SERVICE, SHELL_AUTH_TOKEN, ShellModule } from '@alfresco/adf-
 import { StoreDevtoolsModule } from '@ngrx/store-devtools';
 import { APP_ROUTES } from './app.routes';
 import { MatIconRegistry } from '@angular/material/icon';
-import { MAT_FORM_FIELD_DEFAULT_OPTIONS } from '@angular/material/form-field';
 
 registerLocaleData(localeFr);
 registerLocaleData(localeDe);
@@ -112,10 +111,6 @@ registerLocaleData(localeSv);
         name: 'app',
         source: 'assets'
       }
-    },
-    {
-      provide: MAT_FORM_FIELD_DEFAULT_OPTIONS,
-      useValue: { appearance: 'fill', floatLabel: 'always' }
     }
   ],
   declarations: [AppComponent],
--- a/projects/aca-content/src/lib/ui/application.scss
+++ b/projects/aca-content/src/lib/ui/application.scss
@@ -1,5 +1,6 @@
 /* stylelint-disable scss/no-global-function-names */
 @use '@angular/material' as mat;
+@use './mat-selectors.scss' as ms;
 @import './mixins';
 @import './theme';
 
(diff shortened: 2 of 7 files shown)


AleksanderSklorz:
@Isoumyajit I see one issue. Actually selectbox looks like: 

image


but earlier it looked: 
image


As you can see - earlier only selectbox was greyed out but label had not grayed out background. For selectbox please change only that background, rest looks good for selectbox

app/src/app/app.module.ts
```diff
@@ -111,6 +112,10 @@ registerLocaleData(localeSv);
         name: 'app',
         source: 'assets'
       }
+    },
+    {
+      provide: MAT_FORM_FIELD_DEFAULT_OPTIONS,
```
AleksanderSklorz: I think that change will affect whole application so you will change behavior of form fields in whole application instead just in metadata sidebar

projects/aca-content/src/lib/ui/application.scss
```diff
@@ -47,6 +48,13 @@ ng-component {
   }
 }
 
+.adf-property-list .adf-property-value {
```
AleksanderSklorz: why that change is required here in global application.scss file instead of just adding it to styles specific for metadata sidebar ?

projects/aca-content/src/lib/ui/mat-selectors.scss
```diff
@@ -99,3 +99,5 @@ $mat-select-disabled: '.mat-mdc-select-disabled';
 $mat-select-value: '.mat-mdc-select-value';
 $mat-checkbox-native-control: '.mdc-checkbox__native-control';
 $mat-card-header: '.mat-mdc-card-header';
+$mat-text-field-input: '.mdc-text-field__input';
+$mat-mdc-form-field-infix: '.mat-mdc-form-field-infix';
```
AleksanderSklorz: that selector is already added in 32nd line

app/src/app/app.module.ts
Isoumyajit: Yes I think all the form-fields looks similar and behave consistently so I applied it globally in the app.module.ts but it can be separately applied to specific form-fields as well. @AleksanderSklorz

projects/aca-content/src/lib/ui/application.scss
Isoumyajit: I assume all the form-fields must look same and behave consistently in all the places. so applied that in a global level @AleksanderSklorz.

projects/aca-content/src/lib/ui/mat-selectors.scss
Isoumyajit: Removed

app/src/app/app.module.ts
AleksanderSklorz: We can't apply it globally because we don't use the same inputs everywhere in application. It must be applied only for places which has been broken after HxP change. We can't use it globally because it breaks our application, for example I checked now place not related with metadata sidebar) and it's broken because now it looks like: 
image

but earlier we had not floating labels: 
image

Please revert that change and apply only to metadata sidebar (to places which were broken)

projects/aca-content/src/lib/ui/application.scss
```diff
@@ -54,6 +55,23 @@ ng-component {
   margin-bottom: 65px;
   border-top: 1px solid var(--theme-metadata-property-panel-border-color);
   border-bottom: 1px solid var(--theme-metadata-property-panel-border-color);
+
+  .adf-card-view-selectitem #{ms.$mat-form-field-infix} {
```
AleksanderSklorz: Similarly like https://github.com/Alfresco/alfresco-content-app/pull/4662/files#r2196860313 please apply these styles to the place related with metadata sidebar. And they should not be even in application.scss file because they are related with specific component, not with whole application

app/src/app/app.module.ts
```diff
@@ -96,6 +98,18 @@ registerLocaleData(localeSv);
       provide: SHELL_AUTH_TOKEN,
       useValue: AuthGuard
     },
+    {
+      provide: TRANSLATION_PROVIDER,
```
AleksanderSklorz: why that changes is added? I can't see relationship between styles and translation

app/src/app/app.module.ts
Isoumyajit: Updated

projects/aca-content/src/lib/ui/application.scss
Isoumyajit: Updated

app/src/app/app.module.ts
Isoumyajit: removed

projects/aca-shared/src/lib/components/info-drawer/info-drawer.component.scss
```diff
@@ -0,0 +1,20 @@
+@use '../../styles/mat-selectors' as ms;
+
+.aca-info-drawer {
```
AleksanderSklorz: We use Encapsulation.None so we have rule to wrap our styles in component's selector, so please wrap all these styles inside: 

aca-info-drawer {

}

selector 

You can find that selector here: 
image

AleksanderSklorz:
@dominikiwanekhyland please review that PR and also https://github.com/Alfresco/alfresco-ng2-components/pull/10982 to check if it works with your changes for that task and let here us know if that works

AleksanderSklorz (approved):
I'm approving it because it looks good for me but please wait for @dominikiwanekhyland approval too to ensure it works with his changes

PR #4666, opened 2025-07-10
MNT-25070: ADW - Search request is altered when using quote marks

rmnvch:
https://hyland.atlassian.net/browse/MNT-25070

### DIFF
--- a/e2e/playwright/search/exclude.tests.json
+++ b/e2e/playwright/search/exclude.tests.json
@@ -3,5 +3,9 @@
   "XAT-5600": "https://hyland.atlassian.net/browse/ACS-6928",
   "XAT-17697": "https://hyland.atlassian.net/browse/ACS-7464",
   "XAT-17121": "https://hyland.atlassian.net/browse/ACS-9795",
-  "XAT-17702": "https://hyland.atlassian.net/browse/ACS-9795"
+  "XAT-17702": "https://hyland.atlassian.net/browse/ACS-9795",
+  "XAT-17701": "https://hyland.atlassian.net/browse/ACS-9860",
+  "XAT-17700": "https://hyland.atlassian.net/browse/ACS-9860",
+  "XAT-5581": "https://hyland.atlassian.net/browse/ACS-9860",
+  "XAT-5589": "https://hyland.atlassian.net/browse/ACS-9860"
 }
--- a/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
+++ b/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
@@ -29,7 +29,7 @@ import { Store } from '@ngrx/store';
 import { NavigateToFolder } from '@alfresco/aca-shared/store';
 import { Pagination, SearchRequest } from '@alfresco/js-api';
 import { SavedSearchesService, SearchQueryBuilderService } from '@alfresco/adf-content-services';
-import { ActivatedRoute, NavigationStart, Router } from '@angular/router';
+import { ActivatedRoute, Event, NavigationStart, Params, Router } from '@angular/router';
 import { BehaviorSubject, of, Subject, throwError } from 'rxjs';
 import { AppTestingModule } from '../../../testing/app-testing.module';
 import { AppService } from '@alfresco/aca-shared';
@@ -52,9 +52,9 @@ describe('SearchComponent', () => {
   let router: Router;
   let route: ActivatedRoute;
   const searchRequest = {} as SearchRequest;
-  let params: BehaviorSubject<any>;
-  let queryParams: Subject<any>;
-  let routerEvents: Subject<any>;
+  let params: BehaviorSubject<Params>;
+  let queryParams: Subject<Params>;
+  let routerEvents: Subject<Event>;
   let showErrorSpy: jasmine.Spy<(message: string, action?: string, interpolateArgs?: any, showAction?: boolean) => MatSnackBarRef<any>>;
   let showInfoSpy: jasmine.Spy<(message: string, action?: string, interpolateArgs?: any, showAction?: boolean) => MatSnackBarRef<any>>;
   let loader: HarnessLoader;
@@ -69,9 +69,9 @@ describe('SearchComponent', () => {
   beforeEach(() => {
     params = new BehaviorSubject({ q: 'TYPE: "cm:folder" AND %28=cm: name: email OR cm: name: budget%29' });
     queryParams = new Subject();
-    routerEvents = new Subject<any>();
+    routerEvents = new Subject();
 
-    const routerMock = jasmine.createSpyObj('Router', ['navigate'], {
+    const routerMock = jasmine.createSpyObj<Router>('Router', ['navigate'], {
       url: '/mock-search-url',
       events: routerEvents
     });
--- a/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
+++ b/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
@@ -29,7 +29,7 @@ import { Store } from '@ngrx/store';
 import { NavigateToFolder } from '@alfresco/aca-shared/store';
 import { Pagination, SearchRequest } from '@alfresco/js-api';
 import { SavedSearchesService, SearchQueryBuilderService } from '@alfresco/adf-content-services';
-import { ActivatedRoute, Router } from '@angular/router';
+import { ActivatedRoute, NavigationStart, Router } from '@angular/router';
 import { BehaviorSubject, of, Subject, throwError } from 'rxjs';
 import { AppTestingModule } from '../../../testing/app-testing.module';
 import { AppService } from '@alfresco/aca-shared';
@@ -53,6 +53,8 @@ describe('SearchComponent', () => {
   let route: ActivatedRoute;
   const searchRequest = {} as SearchRequest;
   let params: BehaviorSubject<any>;
+  let queryParams: Subject<any>;
+  let routerEvents: Subject<any>;
   let showErrorSpy: jasmine.Spy<(message: string, action?: string, interpolateArgs?: any, showAction?: boolean) => MatSnackBarRef<any>>;
   let showInfoSpy: jasmine.Spy<(message: string, action?: string, interpolateArgs?: any, showAction?: boolean) => MatSnackBarRef<any>>;
   let loader: HarnessLoader;
@@ -66,6 +68,14 @@ describe('SearchComponent', () => {
 
   beforeEach(() => {
     params = new BehaviorSubject({ q: 'TYPE: "cm:folder" AND %28=cm: name: email OR cm: name: budget%29' });
+    queryParams = new Subject();
+    routerEvents = new Subject<any>();
+
+    const routerMock = jasmine.createSpyObj('Router', ['navigate'], {
+      url: '/mock-search-url',
+      events: routerEvents
+    });
+
     TestBed.configureTestingModule({
       imports: [AppTestingModule, SearchResultsComponent, MatSnackBarModule, MatMenuModule, NoopAnimationsModule],
       providers: [
@@ -94,9 +104,11 @@ describe('SearchComponent', () => {
                 sortingPreferenceKey: ''
               }
             },
-            params: params.asObservable()
+            params: params.asObservable(),
+            queryParams: queryParams.asObservable()
           }
-        }
+        },
+        { provide: Router, useValue: routerMock }
       ]
     });
 
@@ -106,7 +118,6 @@ describe('SearchComponent', () => {
     translate = TestBed.inject(TranslationService);
     router = TestBed.inject(Router);
     route = TestBed.inject(ActivatedRoute);
-    route.queryParams = of({});
 
     const notificationService = TestBed.inject(NotificationService);
     showErrorSpy = spyOn(notificationService, 'showError');
@@ -303,6 +314,25 @@ describe('SearchComponent', () => {
     component.editSavedSearch({ name: 'test', encodedUrl: 'test', order: 0 });
     tick();
     expect(showErrorSpy).toHaveBeenCalledWith('APP.BROWSE.SEARCH.SAVE_SEARCH.EDIT_DIALOG.ERROR_MESSAGE');
+  }));
+
+  it('should call execute once on page reload', fakeAsync(() => {
+    spyOn(queryBuilder, 'execute');
+    queryParams.next({ q: encodeQuery({ userQuery: 'cm:name:"test*"' }) });
+
+    tick();
+
+    expect(queryBuilder.execute).toHaveBeenCalledTimes(1);
+  }));
+
+  it('should NOT call execute on navigation to search page', fakeAsync(() => {
+    spyOn(queryBuilder, 'execute');
+    routerEvents.next(new NavigationStart(1, '/mock-search-url', 'imperative'));
+    queryParams.next({ q: encodeQuery({ userQuery: 'cm:name:"test*"' }) });
+
+    tick();
+
+    expect(queryBuilder.execute).not.toHaveBeenCalled();
   }));
 
   testHeader(SearchResultsComponent, false);
--- a/projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
+++ b/projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
@@ -64,7 +64,7 @@ import {
   ToolbarComponent
 } from '@alfresco/aca-shared';
 import { SearchSortingDefinition } from '@alfresco/adf-content-services/lib/search/models/search-sorting-definition.interface';
-import { filter, startWith, take, takeUntil } from 'rxjs/operators';
+import { filter, startWith, switchMap, take, takeUntil } from 'rxjs/operators';
 import { CommonModule } from '@angular/common';
 import { TranslateModule } from '@ngx-translate/core';
 import { SearchInputComponent } from '../search-input/search-input.component';
@@ -209,21 +209,15 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
         this.router.events.pipe(
           filter((e): e is NavigationStart => e instanceof NavigationStart),
           startWith(null)
-        )
+        ),
+        this.route.queryParams.pipe(switchMap(() => this.savedSearchesService.getSavedSearches()))
       ])
         .pipe(takeUntilDestroyed(this.destroyRef))
-        .subscribe(([params, navigationStartEvent]) => {
+        .subscribe(([params, navigationStartEvent, savedSearches]) => {
           const shouldExecuteQuery = this.shouldExecuteQuery(navigationStartEvent, params);
 
-          this.savedSearchesService
-            .getSavedSearches()
-            .pipe(takeUntilDestroyed(this.destroyRef))
-            .subscribe((savedSearches) => {
-              const savedSearchFound = savedSearches.find(
-                (savedSearch) => savedSearch.encodedUrl === encodeURIComponent(params[this.queryParamName])
-              );
-              this.initialSavedSearch = savedSearchFound !== undefined ? savedSearchFound : this.initialSavedSearch;
-            });
+          this.initialSavedSearch = savedSearches.find((savedSearch) => savedSearch.encodedUrl === encodeURIComponent(params[this.queryParamName]));
+
           if (params[this.queryParamName]) {
             this.isLoading = true;
           }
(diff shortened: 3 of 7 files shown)


projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -207,42 +204,55 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
...
+              const savedSearchFound = savedSearches.find(
+                (savedSearch) => savedSearch.encodedUrl === encodeURIComponent(params[this.queryParamName])
+              );
+              this.initialSavedSearch = savedSearchFound !== undefined ? savedSearchFound : this.initialSavedSearch;
+            });
+          if (params[this.queryParamName]) {
+            this.isLoading = true;
+          }
+          this.loadedFilters$.next();
+          this.encodedQuery = params[this.queryParamName] || null;
```
AleksanderSklorz: Probably || null can be skipped because if it finds nothing in params then it will give undefined and undefined similarly like null is falsy so should work similarly

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -207,42 +204,55 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
...
+          const shouldExecuteQuery = this.shouldExecuteQuery(navigationStartEvent, params);
+
+          this.savedSearchesService
+            .getSavedSearches()
+            .pipe(takeUntilDestroyed(this.destroyRef))
+            .subscribe((savedSearches) => {
+              const savedSearchFound = savedSearches.find(
+                (savedSearch) => savedSearch.encodedUrl === encodeURIComponent(params[this.queryParamName])
+              );
+              this.initialSavedSearch = savedSearchFound !== undefined ? savedSearchFound : this.initialSavedSearch;
```
AleksanderSklorz: That code can be simplified to:
this.initialSavedSearch = savedSearchFound ?? this.initialSavedSearch;

But do we really need to assign initialSavedSearch to initialSavedSearch? I see you assign that field here and earlier in declaration of that field it's just undefined, so it should work to have just: 

this.initialSavedSearch = savedSearchFound; 

and that's it. Please try that approach

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -207,42 +204,55 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
...
+            });
+          if (params[this.queryParamName]) {
+            this.isLoading = true;
+          }
+          this.loadedFilters$.next();
+          this.encodedQuery = params[this.queryParamName] || null;
+          this.searchedWord = extractSearchedWordFromEncodedQuery(this.encodedQuery);
+          this.updateUserQuery();
+          const filtersFromEncodedQuery = extractFiltersFromEncodedQuery(this.encodedQuery);
+          if (filtersFromEncodedQuery !== null) {
```
AleksanderSklorz: can be just: 
if (filtersFromEncodeQuery) {

}

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -207,42 +204,55 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
...
+            const filtersToLoad = this.queryBuilder.categories.length;
+            let loadedFilters = this.searchedWord === '' ? 0 : 1;
+            this.queryBuilder.filterLoaded.pipe(takeUntilDestroyed(this.destroyRef), takeUntil(this.loadedFilters$)).subscribe(() => {
               loadedFilters++;
               if (filtersToLoad === loadedFilters) {
                 this.loadedFilters$.next();
-                this.queryBuilder.execute(false);
+                if (shouldExecuteQuery) {
+                  this.queryBuilder.execute(false);
```
AleksanderSklorz: this line gives warning: 
Screenshot 2025-07-10 at 08 45 09

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -207,42 +204,55 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
...
-          this.queryBuilder.populateFilters.next({});
-          this.queryBuilder.execute(false);
-        }
-        this.queryBuilder.userQuery = extractUserQueryFromEncodedQuery(this.encodedQuery);
-      });
+            this.queryBuilder.populateFilters.next(filtersFromEncodedQuery);
+          } else {
+            this.queryBuilder.populateFilters.next({});
+            if (shouldExecuteQuery) {
+              this.queryBuilder.execute(false);
```
AleksanderSklorz: similar warning like https://github.com/Alfresco/alfresco-content-app/pull/4666/files#r2196752122 for this line

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -207,42 +204,55 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
...
+        )
+      ])
+        .pipe(takeUntilDestroyed(this.destroyRef))
+        .subscribe(([params, navigationStartEvent]) => {
+          const shouldExecuteQuery = this.shouldExecuteQuery(navigationStartEvent, params);
+
+          this.savedSearchesService
+            .getSavedSearches()
+            .pipe(takeUntilDestroyed(this.destroyRef))
+            .subscribe((savedSearches) => {
```
AleksanderSklorz: this subscribe is inside subscribe from 215th line - we should to avoid subscribe inside other subscribe, can you check if you can use any rxjs operator to avoid nested subscriptions?

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -207,42 +204,55 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
...
+          }
+          this.loadedFilters$.next();
+          this.encodedQuery = params[this.queryParamName] || null;
+          this.searchedWord = extractSearchedWordFromEncodedQuery(this.encodedQuery);
+          this.updateUserQuery();
+          const filtersFromEncodedQuery = extractFiltersFromEncodedQuery(this.encodedQuery);
+          if (filtersFromEncodedQuery !== null) {
+            const filtersToLoad = this.queryBuilder.categories.length;
+            let loadedFilters = this.searchedWord === '' ? 0 : 1;
+            this.queryBuilder.filterLoaded.pipe(takeUntilDestroyed(this.destroyRef), takeUntil(this.loadedFilters$)).subscribe(() => {
```
AleksanderSklorz: this subscribe is inside subscribe from 215th line - we should to avoid subscribe inside other subscribe, can you check if you can use any rxjs operator to avoid nested subscriptions?

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -338,4 +348,16 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
     const updatedUserQuery = formatSearchTerm(this.searchedWord, this.searchConfig['app:fields']);
     this.queryBuilder.userQuery = updatedUserQuery;
   }
+
+  private shouldExecuteQuery(navigationStartEvent: NavigationStart | null, params: Params): boolean {
+    if (navigationStartEvent === null) {
```
AleksanderSklorz: can be just 
if (!navigationStartEvent) {

}

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
```diff
@@ -338,4 +348,16 @@ export class SearchResultsComponent extends PageComponent implements OnInit {
...
+      return true;
+    } else if (navigationStartEvent.navigationTrigger === 'imperative') {
+      return false;
+    } else if (navigationStartEvent.navigationTrigger === 'popstate' || navigationStartEvent.navigationTrigger === 'hashchange') {
+      return true;
```
AleksanderSklorz: the same return is in 354th line (first if) so to avoid repeating return please combine both ifs into single if. So to do that you can just add conditions for navigationTrigger to first if using ||. Then true will be returned when navigationStartEnd is null or navigationTrigger is popstate or navigationTrigger is hashchange

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
AleksanderSklorz, line 1: please write unit tests for changes in that file

projects/aca-content/src/lib/utils/aca-search-utils.ts
```diff
@@ -118,7 +118,13 @@ export function extractSearchedWordFromEncodedQuery(encodedQuery: string): strin
           .split('AND')
           .map((searchCondition) => {
             const searchTerm = searchCondition.split('"')[1];
-            return searchTerm === '*' ? searchTerm : searchTerm.slice(0, -1);
+            if (searchTerm === '*') {
```
AleksanderSklorz: can be simplified to: 
```ts
return searchTerm?.endsWith('*') && searchTerm !== '*' ? searchTerm.slice(0, -1) : searchTerm;

File: projects/aca-content/src/lib/utils/aca-search-utils.ts [hunk 11]
```diff
@@ -139,3 +145,16 @@ export function extractFiltersFromEncodedQuery(encodedQuery: string): any {
...
+
+/**
+ * Trims one set of parentheses from parsed user query.
+ *
+ * @param userQuery user query parsed from encoded query
+ * @returns string
+ */
+function trimUserQuery(userQuery: string): string {
+  if (userQuery) {
+    return userQuery.replace(/^\(|\)$/g, '');
```
AleksanderSklorz: can be just: 
```ts
return userQuery?.replace(/^\(|\)$/g, '') ?? '';
```

No need for if (it's replaced with ? operator) and no need for two returns (it's replaced with single return with ??)

AleksanderSklorz:
@rmnvch PR's description template is incorrect in that PR. For ACA we don't use following template: 
image


For ACA we use simpler one, which looks like: 
image

and it requires only task link

AleksanderSklorz:
Please fix issues found by Sonar: 
image

AleksanderSklorz:
@rmnvch is that fix working? I checked it and it seems not working still (if I checked it correctly): 

https://github.com/user-attachments/assets/4b32a694-bd7b-40f6-bb05-2ad54422a386

rmnvch:
> @rmnvch is that fix working? I checked it and it seems not working still (if I checked it correctly):
> 
>  Screen.Recording.2025-07-10.at.09.21.21.mov

hm.. interesting. I used queries like 'cm:name:"test.docx" AND cm:title:"test" AND cm:description:"test"' and 'cm:content:"test"' for reference and those work as intended. I will check additionally. It seems that '"' doesn't get properly escaped in this query 'cm:name:”test.docx”'. Thank's for pointing that out

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
rmnvch: @AleksanderSklorz , in this case I would leave an Eslint warning. putting "void" satisfies EsLint but isn't acceptable from SonarCloud's perspective. And I think warning looks better than noop .catch() block since original promise handled mostly in a proper way.

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
rmnvch: @AleksanderSklorz , I've refactored that part. I've checked inner implementation of .getSavedSearches() and it turned out that there weren't actually a http observable inside and we get hanging subscriptions.

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
AleksanderSklorz: @rmnvch this one seems not addressed, can you check?

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
AleksanderSklorz: @rmnvch this is not addressed too, can you have a look? Maybe it can be resolved similarly like previous subscribe which you changed

projects/aca-content/src/lib/components/search/search-results/search-results.component.ts
AleksanderSklorz: @rmnvch not addressed, please have a look

rmnvch: @AleksanderSklorz , done

rmnvch:
> > @rmnvch is that fix working? I checked it and it seems not working still (if I checked it correctly):
> > Screen.Recording.2025-07-10.at.09.21.21.mov
> 
> hm.. interesting. I used queries like 'cm:name:"test.docx" AND cm:title:"test" AND cm:description:"test"' and 'cm:content:"test"' for reference and those work as intended. I will check additionally. It seems that '"' doesn't get properly escaped in this query 'cm:name:”test.docx”'. Thank's for pointing that out

UPD:
'cm:name:”test.docx”' double quotes are typographical which are not being escaped thus breaking our request query leading to 500 error
and in this:
'cm:name:"test.docx"'
 they're standard ASCII double quote character (U+0022) which are escaped properly in JSON. 
so the fix works as expected. 

The fix works as expected

e2e/playwright/search/exclude.tests.json
```diff
@@ -3,5 +3,9 @@
   "XAT-5600": "https://hyland.atlassian.net/browse/ACS-6928",
   "XAT-17697": "https://hyland.atlassian.net/browse/ACS-7464",
   "XAT-17121": "https://hyland.atlassian.net/browse/ACS-9795",
-  "XAT-17702": "https://hyland.atlassian.net/browse/ACS-9795"
+  "XAT-17702": "https://hyland.atlassian.net/browse/ACS-9795",
+  "XAT-17701": "https://hyland.atlassian.net/browse/ACS-9860",
```
AleksanderSklorz, line 7: @rmnvch I have one question - did you check if these failing e2es are already existing and are not caused by changes in that PR? If they are caused because of changes in PR then they should be fixed here

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
```diff
@@ -53,6 +53,8 @@ describe('SearchComponent', () => {
   let route: ActivatedRoute;
   const searchRequest = {} as SearchRequest;
   let params: BehaviorSubject<any>;
+  let queryParams: Subject<any>;
```
AleksanderSklorz: please avoid any - instead of any we can use Params here

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
```diff
@@ -53,6 +53,8 @@ describe('SearchComponent', () => {
...
+  let routerEvents: Subject<any>;
```
AleksanderSklorz: please avoid any - instead of any we can use Event type from ```@angular/router```

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
```diff
@@ -66,6 +68,14 @@ describe('SearchComponent', () => {
 
   beforeEach(() => {
     params = new BehaviorSubject({ q: 'TYPE: "cm:folder" AND %28=cm: name: email OR cm: name: budget%29' });
+    queryParams = new Subject();
+    routerEvents = new Subject<any>();
```
AleksanderSklorz: as mentioned here https://github.com/Alfresco/alfresco-content-app/pull/4666/files#r2204050781 - any should be avoided. But you can just remove ```<any>``` completely here, as it's enough to specify generic type in declaration and here we can have just new Subject()

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
```diff
@@ -66,6 +68,14 @@ describe('SearchComponent', () => {
...
+
+    const routerMock = jasmine.createSpyObj('Router', ['navigate'], {
```
AleksanderSklorz: It's any actually: 
image

Please use ```jasmine.createSpyObj<Router>(...);``` to have correct type

AleksanderSklorz:
@rmnvch I see that cm:name:”test.docx” gives still 500. Looking at network and sent request I can see that when we use cm:name:"test.docx" AND cm:title:"test" AND cm:description:"test" 
then it doesn't give 500 because request looks like: 
image

Please notice \ before ". But when we use cm:name:”test.docx” then we have missing \. So maybe adding \ will resolve problem?
Here is what is sent: 
query: "(cm:name:”test.docx”)"

you see, \ is missing

AleksanderSklorz:
I found some issue. If I use 
cm:name:"test.docx" AND cm:title:"test" AND cm:description:"test"

First time it displays "test.docx test test" in input (it skips extra words like cm:name etc. in input and it only displays searched values). But when I use the same second time then it displays whole: 
"cm:name:"test.docx" AND cm:title:"test" AND cm:description:"test""

See video. Can we make it consistent and always display the same (so independently of numbers of running search it should display the same format)? Request seems correct each time but formatted text in input is different for each time

https://github.com/user-attachments/assets/0281c879-1d86-4152-8879-492124d8ddc7

rmnvch:
> Please notice \ before ". But when we use cm:name:”test.docx” then we have missing . So maybe adding \ will resolve problem? Here is what is sent: query: "(cm:name:”test.docx”)"
> 
> you see, \ is missing

@AleksanderSklorz ,
yeah, but it's JSON.stringify() who escapes regular double quotes using backslashes doing it internally. The problem is that alfresco API doesn't recognize typographical "quotes". I'will probably just replace any typo quote with regular one when building query. This will fix that in this case. But generally I believe it should be handled properly by BE (sanitize and etc.)

AleksanderSklorz:
> > Please notice \ before ". But when we use cm:name:”test.docx” then we have missing . So maybe adding \ will resolve problem? Here is what is sent: query: "(cm:name:”test.docx”)"
> > you see, \ is missing
> 
> @AleksanderSklorz , yeah, but it's JSON.stringify() who escapes regular double quotes using backslashes doing it internally. The problem is that alfresco API doesn't recognize typographical "quotes". I'will probably just replace any typo quote with regular one when building query. This will fix that in this case. But generally I believe it should be handled properly by BE (sanitize and etc.)

@rmnvch Understood. My point was that it's strange that it's escaped properly for that longer search term but for that shorter it's not escaped. We need to resolve it on frontend or on backend. Can you ask Damian Ujma or Piotr Zurek (they are our backend people) if they are able to fix that on backend side? If not then we need to fix that on frontend side as we need to give some solution to customer. When you get information from them if they will fix that on backend or not then please let here know. If they won't fix that then please fix that on frontend side, if they can fix that then let them fix that

rmnvch:
> > > Please notice \ before ". But when we use cm:name:”test.docx” then we have missing . So maybe adding \ will resolve problem? Here is what is sent: query: "(cm:name:”test.docx”)"
> > > you see, \ is missing
> > 
> > 
> > @AleksanderSklorz , yeah, but it's JSON.stringify() who escapes regular double quotes using backslashes doing it internally. The problem is that alfresco API doesn't recognize typographical "quotes". I'will probably just replace any typo quote with regular one when building query. This will fix that in this case. But generally I believe it should be handled properly by BE (sanitize and etc.)
> 
> @rmnvch Understood. My point was that it's strange that it's escaped properly for that longer search term but for that shorter it's not escaped. We need to resolve it on frontend or on backend. Can you ask Damian Ujma or Piotr Zurek (they are our backend people) if they are able to fix that on backend side? If not then we need to fix that on frontend side as we need to give some solution to customer. When you get information from them if they will fix that on backend or not then please let here know. If they won't fix that then please fix that on frontend side, if they can fix that then let them fix that

@AleksanderSklorz ,
issue should be handled by BE team. Here's Jira ticket: 
https://hyland.atlassian.net/browse/ACS-9878

and discussion:
https://teams.microsoft.com/l/message/19:0bfd2ea7634b41f6a6832e9fe29ef251@thread.tacv2/1752491556939?tenantId=8ca5db88-a5ab-48f7-a5e0-4ce50935f807&groupId=18f33871-30a8-4ab9-ab81-ea9ac6212ae8&parentMessageId=1752491556939&teamName=Guilds&channelName=Back%20End%20Guild&createdTime=1752491556939

rmnvch:
> I found some issue. If I use cm:name:"test.docx" AND cm:title:"test" AND cm:description:"test"
> 
> First time it displays "test.docx test test" in input (it skips extra words like cm:name etc. in input and it only displays searched values). But when I use the same second time then it displays whole: "cm:name:"test.docx" AND cm:title:"test" AND cm:description:"test""
> 
> See video. Can we make it consistent and always display the same (so independently of numbers of running search it should display the same format)? Request seems correct each time but formatted text in input is different for each time
> 
>  Screen.Recording.2025-07-14.at.09.43.16.mov

@AleksanderSklorz , 
The issue resolved. The issue was that Angular usually doesn't fire queryParams change for same queries.

PR #4683, opened 2025-07-16
[ACS-9374][ACS-9590] Prevent search popup flicker on Enter key press and infinite loading animation

Shivangi917:
https://hyland.atlassian.net/browse/ACS-9374
https://hyland.atlassian.net/browse/ACS-9590

### DIFF
--- a/projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.html
+++ b/projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.html
@@ -5,8 +5,8 @@
       mat-icon-button
       matPrefix
       class="app-search-button"
-      (click)="openDropdown()"
-      (keydown.enter)="openDropdown()"
+      (click)="searchSubmit()"
+      (keydown.enter)="searchSubmit()"
       [title]="'SEARCH.BUTTON.TOOLTIP' | translate"
     >
       <mat-icon [attr.aria-label]="'SEARCH.BUTTON.ARIA-LABEL' | translate">search</mat-icon>
--- a/projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
+++ b/projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
@@ -123,10 +123,6 @@ export class SearchInputControlComponent implements OnInit, OnChanges {
     }
   }
 
-  openDropdown() {
-    this.searchInput.nativeElement.focus();
-  }
-
   searchSubmit() {
     this.searchFieldFormControl.markAsTouched();
 


projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.html
```diff
@@ -5,7 +5,8 @@
       mat-icon-button
       matPrefix
       class="app-search-button"
-      (click)="searchSubmit()"
+      (click)="openDropdown()"
+      (keydown)="onSearchButtonKeyDown($event)"
```
AleksanderSklorz: Instead of (keydown) please use (keydown.enter) so then you won't need to check what's key using if statement in ts code.

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.html
```diff
@@ -18,7 +19,9 @@
       [type]="inputType"
       id="app-control-input"
       [formControl]="searchFieldFormControl"
-      (keyup.enter)="searchSubmit()"
+      (keydown)="onInputKeyDown($event)"
```
AleksanderSklorz: same like above - please use keydown.enter

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.spec.ts
```diff
@@ -42,14 +42,24 @@ describe('SearchInputControlComponent', () => {
     fixture.detectChanges();
   });
 
-  it('should emit submit event on searchSubmit', () => {
-    component.searchTerm = 'mock-search-term';
+  it('should emit submit event if form is valid', () => {
+    component.searchTerm = 'valid';
+
+    let submittedTerm = '';
+    component.submit.subscribe((term) => (submittedTerm = term));
```
AleksanderSklorz: instead subscribe - can you please spyOn emit function on submit and check using expect if emit function is called with proper parameter?

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.spec.ts
```diff
@@ -42,14 +42,24 @@ describe('SearchInputControlComponent', () => {
...
+    expect(submittedTerm).toBe('valid');
+  });
+
+  it('should not emit submit event if form is invalid', () => {
+    component.searchTerm = '';
 
-    let submittedSearchTerm = '';
-    component.submit.subscribe((searchTerm) => (submittedSearchTerm = searchTerm));
+    let submitted = false;
+    component.submit.subscribe(() => (submitted = true));
```
AleksanderSklorz: same like: https://github.com/Alfresco/alfresco-content-app/pull/4683/files#r2210303486 but check if function is not called

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.spec.ts
```diff
@@ -87,4 +97,57 @@ describe('SearchInputControlComponent', () => {
...
+  });
+
+  it('should handle Enter key on input and submit if valid', () => {
+    const event = new KeyboardEvent('keydown', { key: 'Enter' });
+    spyOn(event, 'preventDefault');
+    spyOn(event, 'stopPropagation');
+    component.searchTerm = 'abc';
+
+    let emitted = '';
+    component.submit.subscribe((val) => (emitted = val));
```
AleksanderSklorz: similarly like https://github.com/Alfresco/alfresco-content-app/pull/4683/files#r2210303486 please check if function is called instead of using subscribe for event

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.spec.ts
```diff
@@ -87,4 +97,57 @@ describe('SearchInputControlComponent', () => {
...
+
+    component.onSearchButtonKeyDown(event);
+    expect(event.preventDefault).toHaveBeenCalled();
+    expect(component.isSearchBarActive).toBeTrue();
+  });
+
+  it('should not submit on non-Enter key press', () => {
+    const event = new KeyboardEvent('keydown', { key: 'Escape' });
+    let emitted = false;
+    component.submit.subscribe(() => (emitted = true));
```
AleksanderSklorz: the same comment like https://github.com/Alfresco/alfresco-content-app/pull/4683/files#r2210305296

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.html
```diff
@@ -18,7 +19,9 @@
...
+      (focus)="onInputFocus()"
```
AleksanderSklorz: no need for onInputFocus function - instead you can do following 
```html 
(focus)="isSearchBarActive = true"

File: projects/aca-content/src/lib/components/search/search-input/search-input.component.html [hunk 8]
```diff
@@ -36,7 +36,14 @@
         (submit)="onSearchSubmit($event)"
         (searchChange)="onSearchChange($event)"
       />
-      <mat-hint *ngIf="hasLibrariesConstraint" class="app-search-hint">{{ 'SEARCH.INPUT.HINT' | translate }}</mat-hint>
+      <div class="app-search-feedback">
+        <mat-hint *ngIf="hasLibrariesConstraint" class="app-search-hint">
+          {{ 'SEARCH.INPUT.HINT' | translate }}
+        </mat-hint>
+        <mat-error *ngIf="searchInputControl.searchFieldFormControl.hasError('required') && searchInputControl.searchFieldFormControl.touched" class="app-search-error">
```
AleksanderSklorz: Regarding to hasError function - we should to avoid calling functions in html (there are exceptions like using functions for events or inside loop but for most cases we can avoid that). Here instead of calling hasError function you can do: 
searchInputControl.searchFieldFormControl.errors.required

Thanks that function is not called. Functions should be avoided because they may affect performance when used in html

projects/aca-content/src/lib/components/search/search-input/search-input.component.html
AleksanderSklorz, line 1: please write unit tests for changes in that file

AleksanderSklorz:
@Shivangi917 I found issue - when I type only spaces then it doesn't show error for me but input is empty (it has only spaces). Also when I type space for library then it shows me that hint with number of characters but it should not count characters if there are only spaces or spaces are at the beginning/end of text. Please have a look: 

https://github.com/user-attachments/assets/4f9fff87-63b2-4f08-9db5-7fa2fd59d902

AleksanderSklorz:
@Shivangi917 I found one issue - when I check checbox and have empty input then there is no search executed but when I uncheck checkbox then search is executed but input is empty - can you have a look? 


https://github.com/user-attachments/assets/3ea8e8d4-9e7b-41b7-b31a-7766f5bf486b

projects/aca-content/src/lib/components/search/search-input/search-input.component.html
Shivangi917: okay

projects/aca-content/src/lib/components/search/search-input/search-input.component.html
AleksanderSklorz: @Shivangi917 seems not addressed

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -1,62 +1,116 @@
...
     }).compileComponents();
 
     fixture = TestBed.createComponent(SearchInputComponent);
     component = fixture.componentInstance;
+    store = TestBed.inject(Store) as jasmine.SpyObj<Store<AppStore>>;
+    store.pipe.and.returnValue(of([]));
+    fixture.detectChanges();
+  });
+
+  function getFirstError(): string {
```
AleksanderSklorz: function should be moved before beforeEach

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -1,62 +1,116 @@
...
+    const error = fixture.debugElement.query(By.directive(MatError));
+    return error?.nativeElement.textContent.trim();
+  }
 
-    // (component as any).searchInputControl = mockControl;
+  function openSearchContainer(): void {
```
AleksanderSklorz: function should be moved before beforeEach

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -1,62 +1,116 @@
...
 
     fixture = TestBed.createComponent(SearchInputComponent);
     component = fixture.componentInstance;
+    store = TestBed.inject(Store) as jasmine.SpyObj<Store<AppStore>>;
+    store.pipe.and.returnValue(of([]));
+    fixture.detectChanges();
+  });
+
+  function getFirstError(): string {
+    const error = fixture.debugElement.query(By.directive(MatError));
```
AleksanderSklorz: instead of using directly debugElement, please try to use UnitTestingUtils helper

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -1,62 +1,116 @@
...
+    return error?.nativeElement.textContent.trim();
+  }
 
-    // (component as any).searchInputControl = mockControl;
+  function openSearchContainer(): void {
+    const menuButton = fixture.debugElement.query(By.css('.app-search-container'));
```
AleksanderSklorz: instead of using directly debugElement, please try to use UnitTestingUtils helper

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -1,62 +1,116 @@
...
+    openSearchContainer();
+    component.searchInputControl.searchFieldFormControl.setValue('');
+    component.searchInputControl.searchFieldFormControl.markAsUntouched();
+    fixture.detectChanges();
+    const error = fixture.debugElement.query(By.directive(MatError));
+    expect(error).toBeNull();
+  });
+
+  it('should dispatch SearchByTermAction when libraries are checked and term is new', () => {
+    spyOn(component as any, 'isLibrariesChecked').and.returnValue(true);
```
AleksanderSklorz: Don't spy on component's functions - you can spy on only things which are outside component like services. We can't spyOn them because it will change behavior of those functions. Test component based on those functions without spying them

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -1,62 +1,116 @@
...
+    spyOn(component as any, 'isFilesChecked').and.returnValue(false);
+    component.searchedWord = 'test';
+    component.onSearchSubmit('Enter');
+    expect(store.dispatch).toHaveBeenCalled();
+  });
+
+  it('should not dispatch SearchByTermAction when no checkboxes are selected and term is empty', () => {
+    store.dispatch.calls.reset();
+
+    spyOn(component as any, 'isLibrariesChecked').and.returnValue(false);
```
AleksanderSklorz: the same like https://github.com/Alfresco/alfresco-content-app/pull/4683/commits/b587f5d1bf7509cd9f1a94485013a6ba8945a098#r2221681874

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -1,62 +1,116 @@
...
-
-    //   const error = fixture.debugElement.query(By.directive(MatError));
-    //   expect(error).toBeNull();
-    // });
+  it('should not show error when field has value', () => {
+    openSearchContainer();
+    component.searchInputControl.searchFieldFormControl.setValue('test');
+    component.searchInputControl.searchFieldFormControl.markAsTouched();
+    fixture.detectChanges();
+    const error = fixture.debugElement.query(By.directive(MatError));
```
AleksanderSklorz: instead of using directly debugElement, please try to use UnitTestingUtils helper

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -1,62 +1,116 @@
...
+    expect(error).toBeNull();
+  });
+
+  it('should not show error when field is untouched', () => {
+    openSearchContainer();
+    component.searchInputControl.searchFieldFormControl.setValue('');
+    component.searchInputControl.searchFieldFormControl.markAsUntouched();
+    fixture.detectChanges();
+    const error = fixture.debugElement.query(By.directive(MatError));
```
AleksanderSklorz: instead of using directly debugElement, please try to use UnitTestingUtils helper

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
```diff
@@ -82,9 +82,18 @@ export class SearchInputControlComponent implements OnInit {
...
+    }, 0);
+  }
+
   searchSubmit() {
-    if (!this.searchFieldFormControl.errors) {
-      this.submit.emit(this.searchTerm);
+    this.searchFieldFormControl.markAsTouched();
+
+    const trimmedTerm = this.searchTerm?.trim();
+    if (this.searchFieldFormControl.valid && trimmedTerm?.length > 0) {
```
AleksanderSklorz: this if can be simplified to: 
```ts
    if (this.searchFieldFormControl.valid && trimmedTerm) {

File: projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts [hunk 18]
```diff
@@ -82,9 +82,18 @@ export class SearchInputControlComponent implements OnInit {
     });
   }
 
+  openDropdown() {
+    setTimeout(() => {
```
AleksanderSklorz: Just question - do we need timeout here?

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
```diff
@@ -82,9 +82,18 @@ export class SearchInputControlComponent implements OnInit {
...
+      this.searchInput.nativeElement.focus();
+    }, 0);
```
AleksanderSklorz: No need to specify 0 - by default 0 is set so you can skip that

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  let component: SearchInputComponent;
+  let store: jasmine.SpyObj<Store<AppStore>>;
+
+  beforeEach(async () => {
+    const storeSpy = jasmine.createSpyObj<Store<AppStore>>('Store', ['dispatch', 'pipe']);
+
+    await TestBed.configureTestingModule({
+      imports: [AppTestingModule, ReactiveFormsModule, SearchInputComponent, SearchInputControlComponent],
+      providers: [{ provide: Store, useValue: storeSpy }],
+      schemas: [NO_ERRORS_SCHEMA]
```
AleksanderSklorz: NO_ERRORS_SCHEMA seems redundant, please remove it

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+describe('SearchInputComponent', () => {
+  let fixture: ComponentFixture<SearchInputComponent>;
+  let component: SearchInputComponent;
+  let store: jasmine.SpyObj<Store<AppStore>>;
+
+  beforeEach(async () => {
+    const storeSpy = jasmine.createSpyObj<Store<AppStore>>('Store', ['dispatch', 'pipe']);
+
+    await TestBed.configureTestingModule({
+      imports: [AppTestingModule, ReactiveFormsModule, SearchInputComponent, SearchInputControlComponent],
```
AleksanderSklorz: ReactiveFormsModule and SearchInputControlComponent are redundant - please remove them

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.html
```diff
@@ -18,14 +19,17 @@
       [type]="inputType"
       id="app-control-input"
       [formControl]="searchFieldFormControl"
-      (keyup.enter)="searchSubmit()"
+      (keydown.enter)="searchSubmit()"
+
```
AleksanderSklorz: redundant empty line, please fix

AleksanderSklorz:
@Shivangi917 message for length for libraries as still as hint (not error). Please use red color for that as we discussed. Also I noticed that when we hit enter for length less than 2 characters then search is fired - should be fixed too as it says we can't use text less than 2 characters: 

image

AleksanderSklorz:
@Shivangi917 adding only spaces cause that required error and min length error are hidden. Spaces should be counted only when they are between other characters. If you have only spaces or you have space at the beginning/end of text then it should not count

https://github.com/user-attachments/assets/d0c3b079-8170-4cd7-8427-07c70584b5a1

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
Shivangi917: The input is always present in the DOM and not conditionally rendered, we don’t need the setTimeout() — we can call .focus() directly. I will remove that. Thanks

projects/aca-content/src/lib/components/search/search-input/search-input.component.html
```diff
@@ -36,7 +36,16 @@
         (submit)="onSearchSubmit($event)"
         (searchChange)="onSearchChange($event)"
       />
-      <mat-hint *ngIf="hasLibrariesConstraint" class="app-search-hint">{{ 'SEARCH.INPUT.HINT' | translate }}</mat-hint>
+      <mat-error *ngIf="hasLibrariesConstraint" class="app-search-error">
```
AleksanderSklorz: The mat-error is repeated 3 times. Here should be only one mat-error and message inside mat-error should be different based on condition

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+  let store: jasmine.SpyObj<Store<AppStore>>;
+  let unitTestingUtils: UnitTestingUtils;
+  let loader: HarnessLoader;
+
+  function getFirstError(): string {
+    const error = unitTestingUtils.getByDirective(MatError);
+    return error?.nativeElement.textContent.trim();
+  }
+
+  async function openMenu() {
```
AleksanderSklorz: missing returned type

projects/aca-content/src/lib/components/search/search-input/search-input.component.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    const menu = await loader.getHarness(MatMenuHarness);
+    await menu.open();
+    return menu;
+  }
+
+  async function getCheckbox(id: string) {
```
AleksanderSklorz: missing returned type

AleksanderSklorz: no need for async here as this function doesn't contain await

projects/aca-content/folder-rules/src/rule-details/validators/no-whitespace.validator.ts
AleksanderSklorz, line 1: This validator should not be here. This is place for folder-rules feature and that validator is not related with this feature. Instead I would put it in one of those places: 
- in ADF project inside core/src/lib I would create separate folder for validators and I would put it there 
- OR in ACA project I would create separate folder in aca-shared/src/lib for validators and I would put it there

projects/aca-content/folder-rules/src/rule-details/validators/no-whitespace.validator.spec.ts
AleksanderSklorz, line 1: This validator should not be here. This is place for folder-rules feature and that validator is not related with this feature. Instead I would put it in one of those places: 
- in ADF project inside core/src/lib I would create separate folder for validators and I would put it there 
- OR in ACA project I would create separate folder in aca-shared/src/lib for validators and I would put it there

projects/aca-content/folder-rules/src/rule-details/validators/no-whitespace.validator.ts
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * from Hyland Software. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { AbstractControl, ValidationErrors, ValidatorFn } from '@angular/forms';
+
+export const noWhitespaceValidator = (): ValidatorFn => {
+  return (control: AbstractControl): ValidationErrors | null => {
+    const rawValue = control.value;
+    if (rawValue === null || rawValue === '') {
```
AleksanderSklorz: can be just 
if (!rawValue) {

}

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
```diff
@@ -107,6 +106,6 @@ export class SearchInputControlComponent implements OnInit {
   }
 
   isTermTooShort() {
-    return !!(this.searchTerm && this.searchTerm.length < 2);
+    return !!(this.searchTerm.trim() && this.searchTerm.trim().length < 2);
```
AleksanderSklorz: can be shorter: 
return this.searchTerm.trim()?.length < 2;

projects/aca-content/src/lib/components/search/search-input/search-input.component.html
```diff
@@ -36,8 +36,21 @@
         (submit)="onSearchSubmit($event)"
         (searchChange)="onSearchChange($event)"
       />
-      <mat-hint *ngIf="hasLibrariesConstraint" class="app-search-hint">{{ 'SEARCH.INPUT.HINT' | translate }}</mat-hint>
-
+      <mat-error *ngIf="searchInputControl.searchFieldFormControl.touched" class="app-search-error">
+        <ng-container *ngIf="hasLibrariesConstraint; else checkRequired">
```
AleksanderSklorz: Now mat-errors are not repeated but instead you have many ng-container/ng-template. 

Let me explain step by step how this should be resolved. 

You need following changes: 

For SearchInputControlComponent: 
1. Add following field
```ts
@Output()
error = new EventEmitter<string>();
```

2. In ngOnInit add
```ts
ngOnInit() {
  ...
this.searchFieldFormControl.statusChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((status) => {
      // here put logic to check if you have some errors and if yes then which error
      this.error.emit(errorWhichYouGot);
    });
}
```
Changes for SearchInputComponent
1. Handle error event from search input control in html
```html
(error)="error = $event"
```
2. Display that error in mat-error: 
```html
<mat-error *ngIf="searchInputControl.searchFieldFormControl.touched" class="app-search-error">
  {{ error }}
</mat-error>
```

Thanks that as you can see we will have only one mat-error element which just displays needed error

Shivangi917: okay

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
```diff
@@ -29,8 +29,9 @@ import { MatButtonModule } from '@angular/material/button';
 import { MatIconModule } from '@angular/material/icon';
 import { MatFormFieldModule } from '@angular/material/form-field';
 import { MatInputModule } from '@angular/material/input';
-import { FormControl, FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { FormControl, FormsModule, ReactiveFormsModule, Validators } from '@angular/forms';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { noWhitespaceValidator } from 'projects/aca-shared/src/lib/validators/no-whitespace.validator';
```
AleksanderSklorz: incorrect import, import should look like: 
import { noWhitespaceValidator } from '@alfresco/aca-shared';

If you can't import from '@alfresco/aca-shared'; then it means that you have probably some missing export for noWhitespaceValidator (please check how other things are exported from aca-shared

projects/aca-content/src/lib/components/search/search-input/search-input.component.html
```diff
@@ -36,8 +36,21 @@
         (submit)="onSearchSubmit($event)"
         (searchChange)="onSearchChange($event)"
       />
-      <mat-hint *ngIf="hasLibrariesConstraint" class="app-search-hint">{{ 'SEARCH.INPUT.HINT' | translate }}</mat-hint>
-
+      <mat-error *ngIf="searchInputControl.searchFieldFormControl.touched" class="app-search-error">
```
AleksanderSklorz: If you will follow with that approach which I described here: 
https://github.com/Alfresco/alfresco-content-app/pull/4683/files#r2227689989

then it's possible that you will be able to move this condition for touched also to ngOnInit to subscribe. Not sure if you will be able to do it in statusChanges but if not then instead of statusChanges use events field (so this.searchFieldFormControl.events.pipe......). 

It can be implemented like: emit errors from search input control only when input is touched.

Please check if something like that will be working

projects/aca-content/assets/i18n/en.json
```diff
@@ -574,7 +574,9 @@
           "FILES": "Files",
           "FOLDERS": "Folders",
           "LIBRARIES": "Libraries",
-          "HINT": "Search input must have at least 2 alphanumeric characters."
+          "HINT": "Search input must have at least 2 alphanumeric characters.",
```
AleksanderSklorz: please rename that as now you don't use that as hint, it can be named: MIN_LENGTH

AleksanderSklorz:
@Shivangi917 I found small issue: 
when I type space, character, space, character then it still says I need at least 2 characters but I have already 2 characters. When I add for example one more space after that then error is gone, see video: 

https://github.com/user-attachments/assets/30610895-85e4-434a-9e9e-29417fdec918

AleksanderSklorz:
@Shivangi917 one more small issue: 
when I have empty input for libraries then it gives me error about number of characters instead of error about required field: 

https://github.com/user-attachments/assets/9b4c2a2e-27f7-4537-8b41-4db98432c7c0

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.spec.ts
```diff
@@ -30,59 +30,60 @@ import { NO_ERRORS_SCHEMA } from '@angular/core';
...
 
   it('should clear searchTerm', () => {
     component.searchTerm = 'c';
     fixture.detectChanges();
-
     component.clear();
     expect(component.searchTerm).toBe('');
   });
 
   it('should check if searchTerm has a length less than 2', () => {
```
AleksanderSklorz, line 82: This unit test fails now: 
image

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
```diff
@@ -80,11 +118,36 @@ export class SearchInputControlComponent implements OnInit {
...
+        } else {
+          this.validationError.emit('');
+        }
+      }
+    });
+  }
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes['hasLibrariesConstraint'] && !changes['hasLibrariesConstraint'].firstChange) {
+      if (this.searchFieldFormControl.touched) {
```
AleksanderSklorz: no need for nested if, just use && to combine both ifs and that this condition as third condition of top if. Also this causes sonar issue: 
image

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
```diff
@@ -41,9 +54,22 @@ import { noWhitespaceValidator } from '@alfresco/aca-shared';
   encapsulation: ViewEncapsulation.None,
   host: { class: 'app-search-control' }
 })
-export class SearchInputControlComponent implements OnInit {
+export class SearchInputControlComponent implements OnInit, OnChanges {
   private readonly destroyRef = inject(DestroyRef);
 
+  private validateInput(): void {
```
AleksanderSklorz: private functions should be moved below public functions

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
```diff
@@ -40,13 +54,33 @@ import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
...
```
AleksanderSklorz: can you rename that function as it is confusing because this function doesn't validate input - input is already validated and errors are already detected. You can name that function as emitValidationError or something like that

AleksanderSklorz:
@Shivangi917 I found following issue - if I type 1 character and then I will select libraries then search is triggered but I should have at least 2 characters and error is saying about that. Search should not be triggered in that situation: 

https://github.com/user-attachments/assets/e6d770cc-7852-4761-a6d5-fb4d46fbb772

AleksanderSklorz:
@Shivangi917 last thing - checking checkbox clears errors so user doesn't know why search is not happen. For example if I type 1 character and check libraries then search doesn't happen and doesn't show error why it doesn't happen.
Also opposite - if I have libraries checked and I type nothing or spaces and then I uncheck libraries then search is not triggered but there is no error to say that I have only spaces/field is required. Can you fix that? Input is already touched so error should be visible:

https://github.com/user-attachments/assets/5d046e8d-a0f4-4052-acb0-52b63f5948d7

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
```diff
@@ -40,13 +54,20 @@ import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
...
   /** Type of the input field to render, e.g. "search" or "text" (default). */
   @Input()
   inputType = 'text';
 
+  /**
+   * Indicates whether the search is constrained by libraries.
+   * If true, specific error messaging or validation behavior may be triggered.
+   */
+  @Input()
+  hasLibrariesConstraint: boolean;
```
nikita-web-ua: can we provide default value for this input?

dominikiwanekhyland (approved):

projects/aca-content/src/lib/components/search/search-input-control/search-input-control.component.ts
Shivangi917: yes we can

PR #4701, opened 2025-07-24
GH Auto: Upstream dependencies ADF:9.1.0-16492059202 JS-API:10.1.0-16492059202 using Tag:9.1.0-16492059202

alfresco-build:
Automatic PR


AleksanderSklorz:
@Shivangi917 last thing - checking checkbox clears errors so user doesn't know why search is not happen. For example if I type 1 character and check libraries then search doesn't happen and doesn't show error why it doesn't happen. 
Also opposite - if I have libraries checked and I type nothing or spaces and then I uncheck libraries then search is not triggered but there is no error to say that I have only spaces/field is required. Can you fix that? Input is already touched so error should be visible: 

https://github.com/user-attachments/assets/ad42efa0-e3bd-4e90-a71a-dc448b9782a1

AleksanderSklorz (approved):

PR #4750, opened 2025-08-22
[ACS-10035] Ensure ADW handles no PUT for preferences API method in ACS below 25.x v. [PoC]

rmnvch:
https://hyland.atlassian.net/browse/ACS-10035

25.08 UPD:
- utilizing existing rules/extensions mechanism for managing supported version;
- adds new evaluator with helper functions
- adds pipe as alternative approach for hiding components
- extends AppExtensionService to support manual evaluator call

introduces:
1. AcsVersionManagerService: checks minimal version with current
2. IsFeatureAvailablePipe
4. configuration-based approach adding "acs-version-manger" prop in app.config

### DIFF
--- a/docs/extending/rules-list.md
+++ b/docs/extending/rules-list.md
@@ -76,4 +76,10 @@ or not.
 | 1.7.0   | app.navigation.isPreview        | Current page is **Preview**.                                     |
 | 5.1.1   | app.navigation.isDetails        | User is currently on the **Folder Details** page.                |
 
+#### ACS Versions compatibility Rules/Evaluators
 
+Rules/Evaluators created for specific features in ADW to be checked if supported in current ACS version. Evaluators are created using  **createVersionRule** helper function locking specific version number into the rule.
+
+| Version | Key                             | Description                                                              |
+|---------|---------------------------------|--------------------------------------------------------------------------|
+| 8.1.0   | isSavedSearchAvailable          | Checks whether current ACS version supports PUT method in Preferences API|
--- a/docs/extending/rules-list.md
+++ b/docs/extending/rules-list.md
@@ -80,6 +80,6 @@ or not.
 
 Rules/Evaluators created for specific features in ADW to be checked if supported in current ACS version. Evaluators are created using  **createVersionRule** helper function locking specific version number into the rule.
 
-| Version | Key                             | Description                                                              |
-|---------|---------------------------------|--------------------------------------------------------------------------|
-| 8.1.0   | isSavedSearchAvailable          | Checks whether current ACS version supports PUT method in Preferences API|
+| Version | Key                             | Description                                                               |
+|---------|---------------------------------|---------------------------------------------------------------------------|
+| 8.1.0   | isSavedSearchAvailable          | Checks whether current ACS version supports PUT method in Preferences API |
--- a/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
+++ b/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
@@ -24,6 +24,7 @@
 
 import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { SearchResultsComponent } from './search-results.component';
+import { Pipe, PipeTransform } from '@angular/core';
 import { AppConfigService, NotificationService, TranslationService } from '@alfresco/adf-core';
 import { Store } from '@ngrx/store';
 import { NavigateToFolder } from '@alfresco/aca-shared/store';
@@ -32,7 +33,7 @@ import { SavedSearchesService, SearchQueryBuilderService } from '@alfresco/adf-c
 import { ActivatedRoute, Event, NavigationStart, Params, Router } from '@angular/router';
 import { BehaviorSubject, of, Subject, throwError } from 'rxjs';
 import { AppTestingModule } from '../../../testing/app-testing.module';
-import { AppExtensionService, AppService } from '@alfresco/aca-shared';
+import { AppService } from '@alfresco/aca-shared';
 import { MatSnackBarModule, MatSnackBarRef } from '@angular/material/snack-bar';
 import { Buffer } from 'buffer';
 import { testHeader } from '../../../testing/document-base-page-utils';
@@ -41,6 +42,13 @@ import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { MatMenuModule } from '@angular/material/menu';
 import { MatMenuHarness } from '@angular/material/menu/testing';
+
+@Pipe({ name: 'isFeatureSupportedInCurrentAcs', standalone: true })
+class MockIsFeatureSupportedInCurrentAcsPipe implements PipeTransform {
+  transform() {
+    return of(true);
+  }
+}
 
 describe('SearchComponent', () => {
   let component: SearchResultsComponent;
@@ -112,6 +120,12 @@ describe('SearchComponent', () => {
       ]
     });
 
+    TestBed.overrideComponent(SearchResultsComponent, {
+      add: {
+        imports: [MockIsFeatureSupportedInCurrentAcsPipe]
+      }
+    });
+
     config = TestBed.inject(AppConfigService);
     store = TestBed.inject(Store);
     queryBuilder = TestBed.inject(SearchQueryBuilderService);
@@ -131,13 +145,6 @@ describe('SearchComponent', () => {
     component = fixture.componentInstance;
 
     spyOn(queryBuilder, 'update').and.stub();
-
-    // Mock AppExtensionService methods
-    spyOn(AppExtensionService.prototype, 'isFeatureSupported').and.returnValue(true);
-    spyOn(AppExtensionService.prototype, 'getCreateActions').and.returnValue(of([]));
-    spyOn(AppExtensionService.prototype, 'getAllowedToolbarActions').and.returnValue(of([]));
-    spyOn(AppExtensionService.prototype, 'getBulkActions').and.returnValue(of([]));
-    spyOn(AppExtensionService.prototype, 'getViewerToolbarActions').and.returnValue(of([]));
 
     fixture.detectChanges();
     loader = TestbedHarnessEnvironment.loader(fixture);
--- a/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
+++ b/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
@@ -31,7 +31,7 @@ import { NavigateToFolder } from '@alfresco/aca-shared/store';
 import { Pagination, SearchRequest } from '@alfresco/js-api';
 import { SavedSearchesService, SearchQueryBuilderService } from '@alfresco/adf-content-services';
 import { ActivatedRoute, Event, NavigationStart, Params, Router } from '@angular/router';
-import { BehaviorSubject, of, Subject, throwError } from 'rxjs';
+import { BehaviorSubject, Observable, of, Subject, throwError } from 'rxjs';
 import { AppTestingModule } from '../../../testing/app-testing.module';
 import { AppService } from '@alfresco/aca-shared';
 import { MatSnackBarModule, MatSnackBarRef } from '@angular/material/snack-bar';
@@ -43,9 +43,9 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { MatMenuModule } from '@angular/material/menu';
 import { MatMenuHarness } from '@angular/material/menu/testing';
 
-@Pipe({ name: 'isFeatureSupportedInCurrentAcs', standalone: true })
+@Pipe({ name: 'isFeatureSupportedInCurrentAcs' })
 class MockIsFeatureSupportedInCurrentAcsPipe implements PipeTransform {
-  transform() {
+  transform(): Observable<boolean> {
     return of(true);
   }
 }
--- a/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
+++ b/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
@@ -32,7 +32,7 @@ import { SavedSearchesService, SearchQueryBuilderService } from '@alfresco/adf-c
 import { ActivatedRoute, Event, NavigationStart, Params, Router } from '@angular/router';
 import { BehaviorSubject, of, Subject, throwError } from 'rxjs';
 import { AppTestingModule } from '../../../testing/app-testing.module';
-import { AppService } from '@alfresco/aca-shared';
+import { AppExtensionService, AppService } from '@alfresco/aca-shared';
 import { MatSnackBarModule, MatSnackBarRef } from '@angular/material/snack-bar';
 import { Buffer } from 'buffer';
 import { testHeader } from '../../../testing/document-base-page-utils';
@@ -131,6 +131,13 @@ describe('SearchComponent', () => {
     component = fixture.componentInstance;
 
     spyOn(queryBuilder, 'update').and.stub();
+
+    // Mock AppExtensionService methods
+    spyOn(AppExtensionService.prototype, 'isFeatureSupported').and.returnValue(true);
+    spyOn(AppExtensionService.prototype, 'getCreateActions').and.returnValue(of([]));
+    spyOn(AppExtensionService.prototype, 'getAllowedToolbarActions').and.returnValue(of([]));
+    spyOn(AppExtensionService.prototype, 'getBulkActions').and.returnValue(of([]));
+    spyOn(AppExtensionService.prototype, 'getViewerToolbarActions').and.returnValue(of([]));
 
     fixture.detectChanges();
     loader = TestbedHarnessEnvironment.loader(fixture);
--- a/projects/aca-content/src/lib/pipes/is-feature-supported.pipe.spec.ts
+++ b/projects/aca-content/src/lib/pipes/is-feature-supported.pipe.spec.ts
@@ -1,0 +1,47 @@
+/*!
+ * Copyright © 2005-2025 Hyland Software, Inc. and its affiliates. All rights reserved.
+ *
+ * Alfresco Example Content Application
+ *
+ * This file is part of the Alfresco Example Content Application.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail. Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * The Alfresco Example Content Application is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * The Alfresco Example Content Application is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * from Hyland Software. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { IsFeatureSupportedInCurrentAcsPipe } from './is-feature-supported.pipe';
+
+class MockAppExtensionService {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  isFeatureSupported(_feature: string) {
+    return true;
+  }
+}
+
+describe('IsFeatureSupportedInCurrentAcsPipe', () => {
+  it('should create an instance', () => {
+    const pipe = new IsFeatureSupportedInCurrentAcsPipe(new MockAppExtensionService() as any);
+    expect(pipe).toBeTruthy();
+  });
+
+  it('should call isFeatureSupported in AppExtensionService', () => {
+    const service = new MockAppExtensionService();
+    spyOn(service, 'isFeatureSupported').and.returnValue(false);
+    const pipe = new IsFeatureSupportedInCurrentAcsPipe(service as any);
+    expect(pipe.transform('someFeature')).toBe(false);
+    expect(service.isFeatureSupported).toHaveBeenCalledWith('someFeature');
+  });
+});
--- a/projects/aca-content/src/lib/pipes/is-feature-supported.pipe.spec.ts
+++ b/projects/aca-content/src/lib/pipes/is-feature-supported.pipe.spec.ts
@@ -22,6 +22,7 @@
  * from Hyland Software. If not, see <http://www.gnu.org/licenses/>.
  */
 
+import { of } from 'rxjs';
 import { IsFeatureSupportedInCurrentAcsPipe } from './is-feature-supported.pipe';
 
 class MockAppExtensionService {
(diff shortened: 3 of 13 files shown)


projects/aca-content/src/lib/pipes/is-feature-available.pipe.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Pipe, PipeTransform } from '@angular/core';
+import { AcsVersionManagerService } from '../services/acs-version-manager.service';
+
+@Pipe({
+  name: 'isFeatureAvailable'
+})
+export class IsFeatureAvailablePipe implements PipeTransform {
+  constructor(private acsVersionManager: AcsVersionManagerService) {}
+
+  transform(feature: string): boolean {
```
dominikiwanekhyland: feature: Features ? to prevent some random strings or typos?

projects/aca-content/src/lib/components/search/search-save/sidenav/save-search-sidenav.component.html
```diff
@@ -1,3 +1,8 @@
-@if (item) {
-<app-expand-menu [item]="item" (actionClicked)="onActionClick($event)" />
+@if('savedSearches' | isFeatureAvailable) {
```
dominikiwanekhyland: with that implementation we still have to start save-search-sidenav component which call in init SavedSearchesService.init() so it still tries to fetch them?

I think what can be done here not to start that component, would be to create a visibility rule (many examples in app.rules.ts) and add it for app.search.navbar in app.extensions.json

rmnvch: will check those rules, thanks

app/src/app.config.json
```diff
@@ -9,6 +9,14 @@
   "mobileAppSwitch": {
     "enabled": true
   },
+  "acs-version-manager": {
```
MichalKinas: I don't think exposing this in `app.config.json` is a good idea, this file is often customized by the users so they can mess with that by mistake, it can be defined in some other place per feature

projects/aca-shared/rules/src/app.rules.ts
```diff
@@ -483,6 +483,48 @@ export function canOpenWithOffice(context: AcaRuleContext): boolean {
...
+ * Checks if user savedSearches are supported by current ACS version.
+ * JSON ref: `isSavedSearchAvailable`
+ */
+export const isSavedSearchAvailable = createVersionRule('25.1.0');
+
+/**
+ * Partially applies minimal version of a feature against a core compatibility evaluation.
+ * @param minimalVersion The minimal version to check against.
+ */
+function createVersionRule(minimalVersion: string): (context: RuleContext) => boolean {
```
MichalKinas: Let's also export those functions so that thye can be used in ADW/ACC too

rmnvch: done

projects/aca-content/src/lib/pipes/is-feature-available.pipe.ts
rmnvch: @dominikiwanekhyland , I've changed implementation a bit and from now on that pipe accepts an id of an evaluator. We have evaluators provided as strings on module level, so maybe I'd stay with strings for now.

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
```diff
@@ -132,6 +132,13 @@ describe('SearchComponent', () => {
 
     spyOn(queryBuilder, 'update').and.stub();
 
+    // Mock AppExtensionService methods
+    spyOn(AppExtensionService.prototype, 'isFeatureSupported').and.returnValue(true);
```
AleksanderSklorz: Why those functions are mocked in that way? We usually mock functions on instances instead of prototypes so it should be something like 
```ts
const appExtensionService = TestBed.inject(AppExtensionService);
spyOn(appExtensionService, 'isFeatureSupported').and.returnValue(true);
```

Eventually if it can't be mocked like that because of some reasons then we mocked it at TestBed.configureTestingModule level, something like: 
```ts
TestBed.configureTestingModule({
  provide: AppExtensionService,
  useValue: {
    // mocked functions 
  }
})

Comment (AleksanderSklorz) [inline projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts]: One more thing regarding to that. In previous comment I showed how we mock services in components but in that case I can see that your service is not used in component but it's used in IsFeatureSupportedInCurrentAcsPipe pipe so there is no sense to run real logic of that pipe because we have separate tests for pipe in separated file. Instead I would mock whole pipe completely and its transform function. You can mock pipe for example by applying mock in TestBed.configureTestingModule. 

Why I suggest to mock whole pipe and its transform function instead of mocking appExtensionService? 
Because each test file should test only its part of code. Here we have test file for component so should to test only that component. We have separated test file for pipe which should to test only pipe. So in that case we should not refer to pipe's logic in component's test and mock service which is used in pipe because we should not even care about content of that pipe. We should to just mock out pipe and transform function and don't care if that pipe use any services. Thanks that our tests will be isolated.

File: projects/aca-content/src/lib/pipes/is-feature-supported.pipe.ts [hunk 6]
```diff
@@ -0,0 +1,43 @@
...
+import { Store } from '@ngrx/store';
+import { map } from 'rxjs';
+
+@Pipe({
+  name: 'isFeatureSupportedInCurrentAcs'
+})
+export class IsFeatureSupportedInCurrentAcsPipe implements PipeTransform {
+  constructor(
+    private readonly appExtensionsService: AppExtensionService,
+    private store: Store<AppStore>
```
AleksanderSklorz: can be readonly too

projects/aca-content/src/lib/pipes/is-feature-supported.pipe.ts
```diff
@@ -0,0 +1,43 @@
...
+  ) {}
+
+  transform(evaluatorId: string) {
```
AleksanderSklorz: missing return type

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
AleksanderSklorz: Please notice that thanks that you will no need to mock out service here so it will resolve that comment too: https://github.com/Alfresco/alfresco-content-app/pull/4750/files#r2309667202

projects/aca-content/src/lib/pipes/is-feature-supported.pipe.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * from Hyland Software. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { of } from 'rxjs';
+import { IsFeatureSupportedInCurrentAcsPipe } from './is-feature-supported.pipe';
+
+class MockAppExtensionService {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
AleksanderSklorz: Instead of disabling eslint I would to just remove _feature parameter. That parameter is unused by mock so no need for that. It won't break anything because your code will use mocked code so nothing will be broken.

projects/aca-content/src/lib/pipes/is-feature-supported.pipe.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  isFeatureSupported(_feature: string) {
```
AleksanderSklorz: missing return type

projects/aca-content/src/lib/pipes/is-feature-supported.pipe.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  let serviceSpy: jasmine.SpyObj<MockAppExtensionService>;
+  let storeSpy: jasmine.SpyObj<any>;
+
+  beforeEach(() => {
+    serviceSpy = jasmine.createSpyObj('MockAppExtensionService', ['isFeatureSupported']);
+    storeSpy = jasmine.createSpyObj('Store', ['dispatch', 'select']);
+  });
+
+  it('should create an instance', () => {
+    const pipe = new IsFeatureSupportedInCurrentAcsPipe(serviceSpy as any, storeSpy);
```
AleksanderSklorz: no need to create pipes manually in tests, we can just use TestBed.inject(IsFeatureSupportedInCurrentAcsPipe). It will be better because thanks that we will avoid using 'as any' and it won't break our tests when parameters of constructor will change. To use approach with TestBed.inject you will need to register pipe in provider array in TestBed.configureTestingModule probably. Something like: 
```ts
TestBed.configureTestingModule({
  providers: [IsFeatureSupportedInCurrentAcsPipe]
});

File: projects/aca-content/src/lib/pipes/is-feature-supported.pipe.spec.ts [hunk 11]
```diff
@@ -0,0 +1,59 @@
...
+class MockAppExtensionService {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  isFeatureSupported(_feature: string) {
+    return true;
+  }
+}
+
+describe('IsFeatureSupportedInCurrentAcsPipe', () => {
+  let serviceSpy: jasmine.SpyObj<MockAppExtensionService>;
+  let storeSpy: jasmine.SpyObj<any>;
```
AleksanderSklorz: instead of any should be 
```ts
jasmine.SpyObj<Store<AppStore>>

File: projects/aca-content/src/lib/pipes/is-feature-supported.pipe.spec.ts [hunk 12]
```diff
@@ -0,0 +1,59 @@
...
+
+  beforeEach(() => {
+    serviceSpy = jasmine.createSpyObj('MockAppExtensionService', ['isFeatureSupported']);
+    storeSpy = jasmine.createSpyObj('Store', ['dispatch', 'select']);
```
AleksanderSklorz, line 39: we can consider using provideMockStore inside providers array in TestBed.configureTestingModule. Thanks that we can refer to store by TestBed.inject(MockStore) and that MockStore type has some helpful functions like overrideSelector which makes testing store easier

projects/aca-content/src/lib/pipes/is-feature-supported.pipe.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  it('should create an instance', () => {
+    const pipe = new IsFeatureSupportedInCurrentAcsPipe(serviceSpy as any, storeSpy);
+    expect(pipe).toBeTruthy();
+  });
+
+  it('should call isFeatureSupported in AppExtensionService', (done) => {
+    serviceSpy.isFeatureSupported.and.returnValue(false);
+    storeSpy.select.and.returnValue(of('7.4.0'));
+    const pipe = new IsFeatureSupportedInCurrentAcsPipe(serviceSpy as any, storeSpy);
```
AleksanderSklorz: same here like in https://github.com/Alfresco/alfresco-content-app/pull/4750/files#r2309719176 - please use approach with TestBed.inject to get pipe. Also please move that pipe to some variable to reuse it instead of setting it twice (if you will remove first test and use pipe only once then you don't need to create variable to reuse it but please still use TestBed.inject to get pipe)

projects/aca-content/src/lib/pipes/is-feature-supported.pipe.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+describe('IsFeatureSupportedInCurrentAcsPipe', () => {
+  let serviceSpy: jasmine.SpyObj<MockAppExtensionService>;
+  let storeSpy: jasmine.SpyObj<any>;
+
+  beforeEach(() => {
+    serviceSpy = jasmine.createSpyObj('MockAppExtensionService', ['isFeatureSupported']);
+    storeSpy = jasmine.createSpyObj('Store', ['dispatch', 'select']);
+  });
+
+  it('should create an instance', () => {
```
AleksanderSklorz: no need to test only pipe creation because you have other test below so if pipe can't be created then the second test will just fail so you can remove that test

projects/aca-content/src/lib/pipes/is-feature-supported.pipe.ts
```diff
@@ -0,0 +1,43 @@
...
+    return this.store.select(getCurrentACSVersion).pipe(map(() => this.appExtensionsService.isFeatureSupported(evaluatorId)));
```
AleksanderSklorz: We use getCurrentACSVersion here but that version is never read from that selector. We should to read that version from getCurrentACSVersion selector and used that

projects/aca-shared/rules/src/app.rules.ts
```diff
@@ -483,6 +483,48 @@ export function canOpenWithOffice(context: AcaRuleContext): boolean {
...
+ */
+export const isSavedSearchAvailable = createVersionRule('25.1.0');
+
+/**
+ * Partially applies minimal version of a feature against a core compatibility evaluation.
+ * @param minimalVersion The minimal version to check against.
+ */
+export function createVersionRule(minimalVersion: string): (context: RuleContext) => boolean {
+  return (context: RuleContext): boolean => {
+    const acsVersion = context.repository.version?.display?.split(' ')[0];
```
AleksanderSklorz, line 498: We should not duplicate logic of getCurrentACSVersion selector. We should to have only one source of version instead of duplicating because then having reusable selector has no sense. I see that in pipe you have something like: 
```ts
this.store.select(getCurrentACSVersion).pipe(map(() => this.appExtensionsService.isFeatureSupported(evaluatorId)))
```

You use getCurrentACSVersion but you don't read that version from that selector so that selector usage is useless. Instead we should use getCurrentACSVersion selector, read version from that selector and pass that version as parameter to function which check if version is correct. I would do that in following way: 

1. Create some object/map with pairs feature -> minimal version, for example: 
```ts
export const FeatureCompatibility = {
  SavedSearch: '25.1.0',
  OtherFeature: '25.2.0'
} as const;

export type FeatureCompatibility = typeof FeatureCompatibility[keyof typeof FeatureCompatibility]
```
2. Now create function which compares actual and needed versions for specific feature: 
```ts
isVersionCompatible(feature: FeatureCompatibility, currentVersion: string): boolean {
  // read required version from FeatureCompatibility object and compare it with currentVersion
}
```
4. Example of usage: 
```ts
// read current version using getCurrentACSVersion selector 
// pass version from getCurrentACSVersion and feature to isVersionCompatible function: 
isVersionCompatible(FeatureCompatibility.SavedSearch, currentVersion);

```

Thanks that approach we will always read version using the same thing (getCurrentACSVersion) without duplicating it in multiple places

projects/aca-content/src/lib/pipes/is-feature-supported.pipe.ts
rmnvch: @AleksanderSklorz , reading store's selector is used as a trigger here. Otherwise there is a race condition - once we reload directly search page (not navigating from personal-files)  without this subscription we simply won't get version as we do get it from existing logic in AppExtensionService
.

projects/aca-shared/rules/src/app.rules.ts
AleksanderSklorz, line 1: Update rules-list.md documentation file with all new/changed rules

projects/aca-shared/rules/src/app.rules.ts
rmnvch: @AleksanderSklorz , there is no duplication and we do have only 1 source of truth - existing AppExtiensionService. In current approach having dedicated selector for version feels a bit redundant as we able to get version from RuleContext using pipe as well. Right now in both approaches - .json config or using pipe we receive version from RuleContext - existing logic. If I've understood correctly your suggestion we would need to pass ACS version additionally, no we do not need to

projects/aca-shared/rules/src/app.rules.ts
```diff
@@ -483,6 +483,48 @@ export function canOpenWithOffice(context: AcaRuleContext): boolean {
...
+ * Checks if user savedSearches are supported by current ACS version.
+ * JSON ref: `isSavedSearchAvailable`
+ */
+export const isSavedSearchAvailable = createVersionRule('25.1.0');
+
+/**
+ * Partially applies minimal version of a feature against a core compatibility evaluation.
+ * @param minimalVersion The minimal version to check against.
+ */
+export function createVersionRule(minimalVersion: string): (context: RuleContext) => boolean {
```
AleksanderSklorz, line 496: no need to export and no need to document functions which are not used outside that file (are not public)

projects/aca-shared/rules/src/app.rules.ts
```diff
@@ -483,6 +483,48 @@ export function canOpenWithOffice(context: AcaRuleContext): boolean {
...
+  return (context: RuleContext): boolean => {
+    const acsVersion = context.repository.version?.display?.split(' ')[0];
+    return isVersionCompatible(acsVersion, minimalVersion);
+  };
+}
+
+export function isVersionCompatible(currentVersion: string, minimalVersion: string): boolean {
```
AleksanderSklorz: no need to export functions which are not used outside that file (are not public). If something is exported only because of access in unit tests then it should be tested indirectly through public functions when that private function is used

projects/aca-shared/rules/src/app.rules.ts
rmnvch: @AleksanderSklorz , there was a comment above:
https://github.com/Alfresco/alfresco-content-app/pull/4750#discussion_r2298162777

projects/aca-shared/rules/src/app.rules.ts
```diff
@@ -483,6 +483,48 @@ export function canOpenWithOffice(context: AcaRuleContext): boolean {
...
+  if (!currentVersion || !minimalVersion) {
+    return false;
+  }
+
+  const currentParts = currentVersion.split('.').map(Number);
+  const minimalParts = minimalVersion.split('.').map(Number);
+  const maxLength = Math.max(currentParts.length, minimalParts.length);
+
+  for (let i = 0; i < maxLength; i++) {
+    const currentSegment = i < currentParts.length ? currentParts[i] : 0;
```
AleksanderSklorz: that line and line below can have simplified condition: 
const currentSegment = currentParts[i] ?? 0;
This will be working because when you refer to element in array which doesn't exist then it will give just undefined so undefined ?? 0 will give 0

projects/aca-shared/rules/src/app.rules.spec.ts
```diff
@@ -1198,6 +1198,72 @@ describe('app.evaluators', () => {
   });
 });
 
+describe('isSavedSearchAvailable', () => {
+  function makeContext(versionDisplay: string | undefined) {
+    return {
+      repository: {
+        version: versionDisplay ? { display: versionDisplay } : undefined
+      }
+    } as any;
```
AleksanderSklorz: use RuleContext type instead of any

projects/aca-shared/rules/src/app.rules.spec.ts
```diff
@@ -1198,6 +1198,72 @@ describe('app.evaluators', () => {
   });
 });
 
+describe('isSavedSearchAvailable', () => {
+  function makeContext(versionDisplay: string | undefined) {
```
AleksanderSklorz: missing return type (should be RuleContext type)

AleksanderSklorz: Instead of string | undefined please mark versionDisplay parameter as optional because it will simplify usages - instead of passing undefined when we don't want to set version then we can just skip that parameter. So it will look like 
```
 function makeContext(versionDisplay?: string) { ... }

Usages: 
makeContext('25.0.0'); 
or 
makeContext();

File: projects/aca-shared/rules/src/app.rules.spec.ts
Comment (AleksanderSklorz) [inline projects/aca-shared/rules/src/app.rules.spec.ts:1]: As I already commented in app.rules.ts file - when you removed redundant exports for functions which are not used outside app.rules.ts file then as I said there - private code should be tested through public code so to test private code you need to call public functions which use that private code, we should not export functions just to make them public for tests

File: projects/aca-shared/src/lib/services/app.extension.service.ts [hunk 22]
```diff
@@ -592,4 +592,8 @@ export class AppExtensionService implements RuleContext {
   bulkActionExecuted(): void {
     this.bulkActionExecuted$.next();
   }
+
+  isFeatureSupported(feature: string) {
```
AleksanderSklorz: please add return type

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
rmnvch: @AleksanderSklorz , the reason for that is when I use just a regular way TestingModule failed to resolve dependency for AppExtensionService because it was also injected in Base component, which result page is extending. That's basically a workaround for that. But I like your suggestion to have a pipe mock for that and will update accordingly, thanks

projects/aca-shared/rules/src/app.rules.ts
AleksanderSklorz: @rmnvch I found for you one example where we pass some additional data to function from app.rules.ts. That example you can find in ADW code in manage-holds.module.ts file. There is code like: 
image
You can see that there I have functions like canManageHolds, canDeleteHoldFolder etc. which are rule functions and there are passed some extra data to them like filePlanRoles or holdsPermissions. So we need to pass feature and current version in that case. I guess your case may look a bit differently but generally should be something similar

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
AleksanderSklorz: Ok, got it. Then please try to follow solution with pipe as it should resolve whole problem completely

projects/aca-shared/rules/src/app.rules.ts
AleksanderSklorz: Sorry, I didn't notice your reply. I will mark @MichalKinas here to make final decision how we should to approach that

projects/aca-shared/rules/src/app.rules.ts
AleksanderSklorz: ok, so if we are planning to use them in ADW/ACC then it's fine, we can keep export here

projects/aca-shared/rules/src/app.rules.ts
MichalKinas: I think both of them should stay, we can internally use the rules based approach and the pipe but the selector could be useful for customers if they would want to add the condition to their extensions in ACA, this should be properly documented in https://hyland.atlassian.net/browse/ACS-9909, regarding the debated code in the PR I think if the race condition happens and context needs some time to get data it would be best to use the version returned by the selector if we're using it either way, so that we won't have to execute `context.repository.version?.display?.split(' ')[0];` twice

rmnvch: The thing is that I've added a particular method in AppExtensionService to embed this pipe's logic into existing flow  therefore it expects RuleContext type all the way down, not exact version string. And the last step in the chain is anyway `context.repository.version?.display?.split(' ')[0];`. To avoid misconception I will just take another trigger probably, to avoid feeling of doubled logic.

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
rmnvch: done

rmnvch: done

projects/aca-content/src/lib/pipes/is-feature-supported.pipe.ts
rmnvch: @AleksanderSklorz , changed the trigger

projects/aca-shared/rules/src/app.rules.ts
rmnvch: now we use different trigger

projects/aca-shared/rules/src/app.rules.ts
rmnvch: done

docs/extending/rules-list.md
```diff
@@ -76,4 +76,10 @@ or not.
 | 1.7.0   | app.navigation.isPreview        | Current page is **Preview**.                                     |
 | 5.1.1   | app.navigation.isDetails        | User is currently on the **Folder Details** page.                |
 
+#### ACS Versions compatibility Rules/Evaluators
 
+Rules/Evaluators created for specific features in ADW to be checked if supported in current ACS version. Evaluators are created using  **createVersionRule** helper function locking specific version number into the rule.
+
+| Version | Key                             | Description                                                              |
+|---------|---------------------------------|--------------------------------------------------------------------------|
+| 8.1.0   | isSavedSearchAvailable          | Checks whether current ACS version supports PUT method in Preferences API|
```
AleksanderSklorz: Intelllij shows warning that this table is not formatted properly - I think it's because of missing one empty space in last column after API word so API word is merged with | character 
image

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
```diff
@@ -42,6 +43,13 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { MatMenuModule } from '@angular/material/menu';
 import { MatMenuHarness } from '@angular/material/menu/testing';
 
+@Pipe({ name: 'isFeatureSupportedInCurrentAcs', standalone: true })
```
AleksanderSklorz: We have Angular 19 so no need for standalone flag because A19 introduced standalone by default

projects/aca-shared/rules/src/app.rules.ts
AleksanderSklorz: I see your new approach @rmnvch. Now you used getRepositoryStatus in pipe instead of getCurrentACSVersion right so we don't get version twice, right? But in that case getCurrentACSVersion selector is useless as it's unused at all so we should to do following: leave that selector for now but when we finish all related tasks then we need to think about that selector if that should be removed completely. But leave it for now because it may be helpful for other related tasks, we can decide about removing it after finishing all tasks

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
```diff
@@ -42,6 +43,13 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
...
+class MockIsFeatureSupportedInCurrentAcsPipe implements PipeTransform {
+  transform() {
```
AleksanderSklorz: missing return type

projects/aca-shared/rules/src/app.rules.spec.ts
```diff
@@ -1198,6 +1198,71 @@ describe('app.evaluators', () => {
...
+  function makeContext(versionDisplay?: string): RuleContext {
+    return {
+      repository: {
+        version: versionDisplay ? { display: versionDisplay } : undefined
+      }
+    } as RuleContext;
+  }
+
+  describe('isSavedSearchAvailable', () => {
+    it('returns true if ACS version is equal to minimal version', () => {
```
nikita-web-ua: tests should start with 'should'

nikita-web-ua (approved):

PR #4751, opened 2025-08-22
[ACS-10076] [ACA] Search input clears when using special search with text:value

dominikiwanekhyland:
https://hyland.atlassian.net/browse/ACS-10076

### DIFF
--- a/projects/aca-content/src/lib/utils/aca-search-utils.ts
+++ b/projects/aca-content/src/lib/utils/aca-search-utils.ts
@@ -117,7 +117,7 @@ export function extractSearchedWordFromEncodedQuery(encodedQuery: string): strin
       ? userQuery
           .split('AND')
           .map((searchCondition) => {
-            const searchTerm = searchCondition.split('"')[1];
+            const searchTerm = searchCondition.includes('"') ? searchCondition.split('"')[1] : searchCondition.trim();
             return searchTerm?.endsWith('*') && searchTerm !== '*' ? searchTerm.slice(0, -1) : searchTerm;
           })
           .join(' ')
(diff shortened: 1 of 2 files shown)


projects/aca-content/src/lib/utils/aca-search-utils.ts
```diff
@@ -117,7 +117,7 @@ export function extractSearchedWordFromEncodedQuery(encodedQuery: string): strin
       ? userQuery
           .split('AND')
           .map((searchCondition) => {
-            const searchTerm = searchCondition.split('"')[1];
+            const searchTerm = searchCondition.includes('"') ? searchCondition.split('"')[1] : searchCondition.trim();
```
AleksanderSklorz, line 120: please add unit tests to test it

DenysVuika: I think that needs some sort of regex here, for instance

```
const userQuery = `title:"Hello World" AND status:active AND "complex phrase"`;

const searchTerms = userQuery
  .split(/\s+AND\s+/i) // split safely, case-insensitive, allow spaces
  .map((condition) => {
    const match = condition.match(/"([^"]+)"|(\S+)/); 
    return match ? match[1] ?? match[2] : '';
  })
  .filter(Boolean); // remove empty strings

console.log(searchTerms);
// ["Hello World", "status:active", "complex phrase"]
```

AleksanderSklorz (approved):

projects/aca-content/src/lib/utils/aca-search-utils.ts
dominikiwanekhyland: @DenysVuika I tried a full regex solution, but it quickly became very complex due to all the edge cases in the unit tests — quoted vs unquoted terms, optional field prefixes, multiple AND conditions, trailing asterisks, etc. Handling everything correctly would require many additional if statements and special cases anyway, so I think keeping the current approach with minimal regex for parsing quoted vs unquoted terms is safer, easier to read, and more maintainable.

PR #4759, opened 2025-08-26
Post release version bump

dominikiwanekhyland:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "alfresco-content-app",
-  "version": "7.1.0",
+  "version": "7.2.0",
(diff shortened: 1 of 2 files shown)


package.json
AleksanderSklorz, line 1: I think you need to update rest of package.json files too, see for example: https://github.com/Alfresco/alfresco-content-app/pull/4678/files

AleksanderSklorz: It's the same comment like Michal's one https://github.com/Alfresco/alfresco-content-app/pull/4759#pullrequestreview-3155462002 (I added in the same time)

AleksanderSklorz (approved):

PR #4760, opened 2025-08-26
Proxy: remove dependency on ADF cli, cleanup

DenysVuika:
- remove the need for ADF CLI and `require`
- remove unnecessary workarounds

### DIFF
--- a/app/proxy.conf.js
+++ b/app/proxy.conf.js
@@ -1,26 +1,18 @@
-require('@alfresco/adf-cli/tooling').dotenvConfig({ path: process.env.ENV_FILE });
-
 const { BASE_URL } = process.env;
+console.log('Using backend URL: ' + BASE_URL || 'unknown');
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


app/proxy.conf.js
```diff
@@ -1,26 +1,18 @@
...
-        "pathRewrite": {
-            "^/alfresco/alfresco": ""
-        },
-        "changeOrigin": true,
-        'logLevel': 'debug',
-        onProxyReq: function(request) {
-          if(request["method"] !== "GET")
-          request.setHeader("origin", BASE_URL);
-        },
-        // workaround for REPO-2260
```
BSekula, line 18: // workaround for REPO-2260

What is REPO-2260? 

do we still need it?  

it was added [here](https://hyland.atlassian.net/browse/AAE-7145)

DenysVuika: no, not needed for years already :)

PR #4784, opened 2025-09-08
[ACS-10116] [ACA] Search page renders HTML from the Description,Title field instead of showing it as plain text

dominikiwanekhyland:
https://hyland.atlassian.net/browse/ACS-10116

### DIFF
--- a/projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
+++ b/projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
@@ -86,6 +86,11 @@ describe('SearchResultsRowComponent', () => {
     utils = new UnitTestingUtils(fixture.debugElement);
   });
 
+  const getNameEl = () => utils.getByCSS('.aca-link.aca-crop-text').nativeElement as HTMLSpanElement;
+  const getTitleEl = () => utils.getByDataAutomationId('search-results-entry-title').nativeElement as HTMLSpanElement;
+  const getDescriptionEl = () => utils.getByDataAutomationId('search-results-entry-description').nativeElement as HTMLDivElement;
+  const getContentEl = () => utils.getByCSS('.aca-result-content.aca-crop-text').nativeElement as HTMLDivElement;
+
   it('should show the current node', () => {
     component.context = { row: { node: nodeEntry } };
     fixture.detectChanges();
@@ -98,23 +103,17 @@ describe('SearchResultsRowComponent', () => {
     component.content$.pipe(first()).subscribe(() => {
       fixture.detectChanges();
 
-      expect(utils.getByCSS('.aca-link.aca-crop-text').nativeElement.innerHTML).toBe('<span class="aca-highlight">Random</span>');
-      expect(utils.getByCSS('.aca-link.aca-crop-text').nativeElement.title).toBe('Random');
+      expect(getNameEl().innerHTML).toBe('<span class="aca-highlight">Random</span>');
+      expect(getNameEl().title).toBe('Random');
 
-      expect(utils.getByCSS('[data-automation-id="search-results-entry-title"]').nativeElement.innerHTML).toBe(
-        ' ( <span class="aca-highlight">Random</span> title )'
-      );
-      expect(utils.getByCSS('[data-automation-id="search-results-entry-title"]').nativeElement.title).toBe('Random title');
+      expect(getTitleEl().innerHTML).toBe(' ( <span class="aca-highlight">Random</span> title )');
+      expect(getTitleEl().title).toBe('Random title');
 
-      expect(utils.getByCSS('[data-automation-id="search-results-entry-description"]').nativeElement.innerHTML).toBe(
-        'some <span class="aca-highlight">random</span> description'
-      );
-      expect(utils.getByCSS('[data-automation-id="search-results-entry-description"]').nativeElement.title).toBe('some random description');
+      expect(getDescriptionEl().innerHTML).toBe('some <span class="aca-highlight">random</span> description');
+      expect(getDescriptionEl().title).toBe('some random description');
 
-      expect(utils.getByCSS('.aca-result-content.aca-crop-text').nativeElement.innerHTML).toBe(
-        '...Interesting <span class="aca-highlight">random</span> content...'
-      );
-      expect(utils.getByCSS('.aca-result-content.aca-crop-text').nativeElement.title).toBe('...Interesting random content...');
+      expect(getContentEl().innerHTML).toBe('...Interesting <span class="aca-highlight">random</span> content...');
+      expect(getContentEl().title).toBe('...Interesting random content...');
       done();
     });
     fixture.detectChanges();
@@ -138,9 +137,8 @@ describe('SearchResultsRowComponent', () => {
     component.name$.pipe(first()).subscribe(() => {
       fixture.detectChanges();
 
-      const nameEl = utils.getByCSS('.aca-link.aca-crop-text').nativeElement;
-      expect(nameEl.innerHTML).toBe('2 &lt; 5 &gt; 3');
-      expect(nameEl.textContent).toBe('2 < 5 > 3');
+      expect(getNameEl().innerHTML).toBe('2 &lt; 5 &gt; 3');
+      expect(getNameEl().textContent).toBe('2 < 5 > 3');
       done();
     });
     fixture.detectChanges();
@@ -155,9 +153,8 @@ describe('SearchResultsRowComponent', () => {
     component.name$.pipe(first()).subscribe(() => {
       fixture.detectChanges();
 
-      const nameEl = utils.getByCSS('.aca-link.aca-crop-text').nativeElement;
-      expect(nameEl.innerHTML).toContain('&lt;script&gt;alert("xss")&lt;/script&gt;');
-      expect(nameEl.textContent).toBe('<script>alert("xss")</script>');
+      expect(getNameEl().innerHTML).toContain('&lt;script&gt;alert("xss")&lt;/script&gt;');
+      expect(getNameEl().textContent).toBe('<script>alert("xss")</script>');
       done();
     });
     fixture.detectChanges();
@@ -172,9 +169,8 @@ describe('SearchResultsRowComponent', () => {
     component.name$.pipe(first()).subscribe(() => {
       fixture.detectChanges();
 
-      const nameEl = utils.getByCSS('.aca-link.aca-crop-text').nativeElement;
-      expect(nameEl.innerHTML).toBe('&lt;b&gt;<span class="aca-highlight">BoldHighlight</span>&lt;/b&gt;');
-      expect(nameEl.textContent).toBe('<b>BoldHighlight</b>');
+      expect(getNameEl().innerHTML).toBe('&lt;b&gt;<span class="aca-highlight">BoldHighlight</span>&lt;/b&gt;');
+      expect(getNameEl().textContent).toBe('<b>BoldHighlight</b>');
       done();
     });
     fixture.detectChanges();
--- a/projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
+++ b/projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
@@ -86,10 +86,11 @@ describe('SearchResultsRowComponent', () => {
     utils = new UnitTestingUtils(fixture.debugElement);
   });
 
-  const getNameEl = () => utils.getByCSS('.aca-link.aca-crop-text').nativeElement as HTMLSpanElement;
-  const getTitleEl = () => utils.getByDataAutomationId('search-results-entry-title').nativeElement as HTMLSpanElement;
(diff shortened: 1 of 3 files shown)


AleksanderSklorz (changes requested):
@dominikiwanekhyland seems like this is still not working properly because for example
 ```html
<h2>test</h2> 
```
tags in title cause that those text is rendered as html instead of just text

image

File: projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts [hunk 1]
```diff
@@ -137,4 +137,121 @@ describe('SearchResultsRowComponent', () => {
...
+        .pipe(first())
+        .subscribe((sanitizedName) => {
+          expect(sanitizedName).not.toContain('<script>');
+          done();
+        });
+      fixture.detectChanges();
+    });
+  });
+
+  function createMaliciousEntry(overrides: any): ResultSetRowEntry {
```
AleksanderSklorz, line 244: please use correct type instead of any, for example
```ts
Partial<ResultNode>
```

AleksanderSklorz: Please use correct type instead of any, for example:
```ts
Partial<ResultNode>

File: projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts [hunk 2]
```diff
@@ -47,6 +48,7 @@ import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 })
 export class SearchResultsRowComponent implements OnInit {
   private settings = inject(AppSettingsService);
+  private readonly sanitizer = inject(DomSanitizer);
```
AleksanderSklorz: private readonly sanitizer = inject(DomSanitizer);

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.ts
```diff
@@ -122,15 +124,17 @@ export class SearchResultsRowComponent implements OnInit {
...
+    this.description$.next(this.sanitizer.sanitize(SecurityContext.HTML, description));
+    this.content$.next(this.sanitizer.sanitize(SecurityContext.HTML, content));
 
     this.nameStripped = this.stripHighlighting(name);
     this.descriptionStripped = this.stripHighlighting(description);
     this.contentStripped = this.stripHighlighting(content);
 
     if (title !== name) {
+      const sanitizedTitle = this.sanitizer.sanitize(SecurityContext.HTML, title);
+      this.title$.next(sanitizedTitle ? ` ( ${sanitizedTitle} )` : '');
```
AleksanderSklorz: Why we duplicate calling next for titlte$? Below is the same call but with unsanitized title.

dominikiwanekhyland:
> @dominikiwanekhyland seems like this is still not working properly because for example
> 
> ```
> <h2>test</h2> 
> ```
> 
> tags in title cause that those text is rendered as html instead of just text
> 
> image

So the code looks like it's done deliberately to allow html there so I think better solution will be to allow **sanitized** html there without any malicious script and it's been agreed with Akash

AleksanderSklorz:
> > @dominikiwanekhyland seems like this is still not working properly because for example
> > ```
> > <h2>test</h2> 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > tags in title cause that those text is rendered as html instead of just text
> > image
> 
> So the code looks like it's done deliberately to allow html there so I think better solution will be to allow **sanitized** html there without any malicious script and it's been agreed with Akash

Hmm, I think html tags are allowed because of need to highlight searched text. I think highlighting was implemented by Michał:
image

But in jira's scenario it breaks our styles because the only "exceptional" case was highlighting. I don't know if we want user to affect our look. Should we wait for @MichalKinas to confirm if we want that behavior @dominikiwanekhyland, @akashrathod28  because only known case was for highlighting? I would say that it should just display tags as normal text without rendering them. Something like: 
image

MichalKinas:
> > > @dominikiwanekhyland seems like this is still not working properly because for example
> > > ```
> > > <h2>test</h2> 
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > tags in title cause that those text is rendered as html instead of just text
> > > image
> > 
> > 
> > So the code looks like it's done deliberately to allow html there so I think better solution will be to allow **sanitized** html there without any malicious script and it's been agreed with Akash
> 
> Hmm, I think html tags are allowed because of need to highlight searched text. I think highlighting was implemented by Michał: image
> 
> But in jira's scenario it breaks our styles because the only "exceptional" case was highlighting. I don't know if we want user to affect our look. Should we wait for @MichalKinas to confirm if we want that behavior @dominikiwanekhyland, @akashrathod28 because only known case was for highlighting? I would say that it should just display tags as normal text without rendering them. Something like: image

Hey guys, yeah so as @AleksanderSklorz mentioned this was introduced to enable highlighting in search results, regarding the other tags I'm not quite sure that this would be valid use case and I sort of doubt someone would want to use this so I would say we should display them as normal text, however we can also ask Arabella for her opinion on this

dominikiwanekhyland:
> > > > @dominikiwanekhyland seems like this is still not working properly because for example
> > > > ```
> > > > <h2>test</h2> 
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > tags in title cause that those text is rendered as html instead of just text
> > > > image
> > > 
> > > 
> > > So the code looks like it's done deliberately to allow html there so I think better solution will be to allow **sanitized** html there without any malicious script and it's been agreed with Akash
> > 
> > 
> > Hmm, I think html tags are allowed because of need to highlight searched text. I think highlighting was implemented by Michał: image
> > But in jira's scenario it breaks our styles because the only "exceptional" case was highlighting. I don't know if we want user to affect our look. Should we wait for @MichalKinas to confirm if we want that behavior @dominikiwanekhyland, @akashrathod28 because only known case was for highlighting? I would say that it should just display tags as normal text without rendering them. Something like: image
> 
> Hey guys, yeah so as @AleksanderSklorz mentioned this was introduced to enable highlighting in search results, regarding the other tags I'm not quite sure that this would be valid use case and I sort of doubt someone would want to use this so I would say we should display them as normal text, however we can also ask Arabella for her opinion on this

@AleksanderSklorz @MichalKinas I've applied those changes, so now any html script is treated as text, except highlihts

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
```diff
@@ -133,8 +133,82 @@ describe('SearchResultsRowComponent', () => {
...
+    } as ResultSetRowEntry;
+
+    component.context = { row: { node: customEntry } };
+    component.name$
+      .asObservable()
+      .pipe(first())
+      .subscribe(() => {
+        fixture.detectChanges();
+
+        const nameElement: HTMLSpanElement = fixture.debugElement.query(By.css('.aca-link.aca-crop-text')).nativeElement;
```
AleksanderSklorz: In new code we use UnitTestingUtils so please try to use this instead of directly By.css. Also please create reusable function because '.aca-link.aca-crop-text' selector is repeated

AleksanderSklorz:
@dominikiwanekhyland on that PR there is introduced new bug when comparing with remote env. 
On branch: 
image

On remote env: 
image

There is empty () on branch when title is not set. On remote env there is no () at all when title is unset.

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
```diff
@@ -91,50 +83,100 @@ describe('SearchResultsRowComponent', () => {
...
-
-        const contentElement: HTMLDivElement = fixture.debugElement.query(By.css('.aca-result-content.aca-crop-text')).nativeElement;
-        expect(contentElement.innerHTML).toBe('...Interesting <span class="aca-highlight">random</span> content...');
-        expect(contentElement.title).toBe('...Interesting random content...');
-        done();
-      });
+    component.content$.pipe(first()).subscribe(() => {
+      fixture.detectChanges();
+
+      expect(utils.getByCSS('.aca-link.aca-crop-text').nativeElement.innerHTML).toBe('<span class="aca-highlight">Random</span>');
```
AleksanderSklorz: That selector .aca-link.aca-crop-text is repeated multiple times so can be wrapped into reusable function

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
```diff
@@ -91,50 +83,100 @@ describe('SearchResultsRowComponent', () => {
...
+      expect(utils.getByCSS('.aca-link.aca-crop-text').nativeElement.title).toBe('Random');
+
+      expect(utils.getByCSS('[data-automation-id="search-results-entry-title"]').nativeElement.innerHTML).toBe(
```
AleksanderSklorz: same here, that selector is used multiple times so can be wrapped into function

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
```diff
@@ -91,50 +83,100 @@ describe('SearchResultsRowComponent', () => {
...
+        ' ( <span class="aca-highlight">Random</span> title )'
+      );
+      expect(utils.getByCSS('[data-automation-id="search-results-entry-title"]').nativeElement.title).toBe('Random title');
+
+      expect(utils.getByCSS('[data-automation-id="search-results-entry-description"]').nativeElement.innerHTML).toBe(
```
AleksanderSklorz: same here, that selector is used multiple times so can be wrapped into function

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
```diff
@@ -91,50 +83,100 @@ describe('SearchResultsRowComponent', () => {
...
+        'some <span class="aca-highlight">random</span> description'
+      );
+      expect(utils.getByCSS('[data-automation-id="search-results-entry-description"]').nativeElement.title).toBe('some random description');
+
+      expect(utils.getByCSS('.aca-result-content.aca-crop-text').nativeElement.innerHTML).toBe(
```
AleksanderSklorz: same here, that selector is used multiple times so can be wrapped into function

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
```diff
@@ -91,50 +83,100 @@ describe('SearchResultsRowComponent', () => {
...
+    component.content$.pipe(first()).subscribe(() => {
+      fixture.detectChanges();
+
+      expect(utils.getByCSS('.aca-link.aca-crop-text').nativeElement.innerHTML).toBe('<span class="aca-highlight">Random</span>');
+      expect(utils.getByCSS('.aca-link.aca-crop-text').nativeElement.title).toBe('Random');
+
+      expect(utils.getByCSS('[data-automation-id="search-results-entry-title"]').nativeElement.innerHTML).toBe(
+        ' ( <span class="aca-highlight">Random</span> title )'
+      );
+      expect(utils.getByCSS('[data-automation-id="search-results-entry-title"]').nativeElement.title).toBe('Random title');
```
AleksanderSklorz: For data-automation-id selectors please use getByDataAutomationId function instead of getByCss

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
```diff
@@ -91,50 +83,96 @@ describe('SearchResultsRowComponent', () => {
 
     fixture = TestBed.createComponent(SearchResultsRowComponent);
     component = fixture.componentInstance;
+    utils = new UnitTestingUtils(fixture.debugElement);
   });
 
+  const getNameEl = () => utils.getByCSS('.aca-link.aca-crop-text').nativeElement as HTMLSpanElement;
```
AleksanderSklorz: Please add return type to those 4 functions

projects/aca-content/src/lib/components/search/search-results-row/search-results-row.component.spec.ts
```diff
@@ -91,50 +83,97 @@ describe('SearchResultsRowComponent', () => {
 
     fixture = TestBed.createComponent(SearchResultsRowComponent);
     component = fixture.componentInstance;
+    utils = new UnitTestingUtils(fixture.debugElement);
   });
 
+  const getNameEl: () => HTMLSpanElement = () => utils.getByCSS('.aca-link.aca-crop-text').nativeElement as HTMLSpanElement;
```
AleksanderSklorz: I meant something like:
```ts
  const getNameEl = (): HTMLSpanElement => utils.getByCSS('.aca-link.aca-crop-text').nativeElement as HTMLSpanElement;
```
We don't set type to const when it can be detected but we set types for "return for functions" :)

AleksanderSklorz (approved):

PR #4795, opened 2025-09-12
[ACS-10100][Security] Remove npx usage from CI/CD and shell scripts to mitigate supply chain risks

dominikiwanekhyland:
https://hyland.atlassian.net/browse/ACS-10100


.github/workflows/pull-request.yml
```diff
@@ -80,7 +80,7 @@ jobs:
         with:
           name: npm-logs
           path: /home/runner/.npm/_logs/
-      - run: npx nx build aca-playwright-shared
+      - run: ./node_modules/.bin/nx build aca-playwright-shared
```
AleksanderSklorz: Based on jira's description: 
image


Maybe it would be good idea to prepare some npm scripts for similar/the same commands (and run them by npm run) to avoid repeating whole commands x times? It would be good at least for commands which are repeated multiple times (or with different parameters if possible to pass parameters). For example this command is repeated also in  .github/workflows/run-e2e-with-env.yml

I won't add this comment everywhere but could you please go through your changes and apply suggestion where possible if you agree that it sounds like good idea?

package.json
```diff
@@ -22,8 +22,21 @@
     "print-affected:build": "nx print-affected --target=build --select=tasks.target.project",
     "print-affected:test": "nx print-affected --target=test --select=tasks.target.project",
     "print-affected:lint": "nx print-affected --target=lint --select=tasks.target.project",
-    "lintfix": "nx affected:lint --fix"
-  },
+    "lintfix": "nx affected:lint --fix",
+    "ci:lint": "nx run-many --target=lint",
+    "ci:lint:affected": "nx affected:lint",
```
AleksanderSklorz: looks like duplicated script and we can use affected:lint script instead

e2e/playwright/authentication/project.json
```diff
@@ -7,7 +7,7 @@
     "e2e": {
       "executor": "nx:run-commands",
       "options": {
-        "commands": ["npx playwright test --config=e2e/playwright/authentication/playwright.config.ts"]
+        "commands": ["playwright test --config=e2e/playwright/authentication/playwright.config.ts"]
```
AleksanderSklorz, line 10: You can treat that comment as optional but if you think that may be helpful then we can probably simplify that command and reuse it too thanks to that we have similar folders structure for each e2e: 
1. You can add following script to package.json: 
"e2e": "playwright test --config=e2e/playwright/$npm_config_project/playwright.config.ts"

2. Then you can use that script in following way: 
npm run e2e --project=viewer

Thanks that it will inject "project" name into path. You can name script and parameter as any name you prefer

g-jaskowski (approved):

PR #4805, opened 2025-09-19
[MNT-25332] ADW does not show End Editing menu option

g-jaskowski:
https://hyland.atlassian.net/browse/MNT-25332


AleksanderSklorz:
@g-jaskowski I see "Edit in Word online" option only in ADW. There is no option like that in ACA so are you sure this change should be in ACA and not in ADW? See screenshots: 
ADW: 
image
ACA: 
image

g-jaskowski:
> @g-jaskowski I see "Edit in Word online" option only in ADW. There is no option like that in ACA so are you sure this change should be in ACA and not in ADW? See screenshots: ADW: image ACA: image

You're right, there's a similar case in MS Online Module in ADW with another rule, I'll close this PR and open a new one

PR #4817, opened 2025-09-30
[ACS-10083] Filter not behaving correctly in file and site

rmnvch:
https://hyland.atlassian.net/browse/ACS-10083

relates to:
ADF PR:
https://github.com/Alfresco/alfresco-ng2-components/pull/11237

blocks ADW PR:
https://github.com/Alfresco/alfresco-applications/pull/1258

### DIFF
--- a/projects/aca-content/src/lib/components/files/files.component.spec.ts
+++ b/projects/aca-content/src/lib/components/files/files.component.spec.ts
@@ -90,7 +90,7 @@ describe('FilesComponent', () => {
           useValue: {
             snapshot: { data: { preferencePrefix: 'prefix' }, paramMap: convertToParamMap({ folderId: undefined }) },
             params: of({ folderId: 'someId' }),
-            queryParamMap: of({})
+            queryParamMap: of(convertToParamMap({}))
           }
         },
         AppExtensionService,
--- a/projects/aca-content/src/lib/components/files/files.component.ts
+++ b/projects/aca-content/src/lib/components/files/files.component.ts
@@ -30,7 +30,7 @@ import {
   ShowHeaderMode
 } from '@alfresco/adf-core';
 import { Component, OnDestroy, OnInit, ViewEncapsulation } from '@angular/core';
-import { ActivatedRoute, ParamMap, Params } from '@angular/router';
+import { ActivatedRoute, Params } from '@angular/router';
 import { Node, NodeEntry, PathElement } from '@alfresco/js-api';
 import { NodeActionsService } from '../../services/node-actions.service';
 import {
@@ -121,7 +121,7 @@ export class FilesComponent extends PageComponent implements OnInit, OnDestroy {
 
     this.title = data.title;
 
-    this.route.queryParamMap.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((queryMap: ParamMap) => {
+    this.route.queryParamMap.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((queryMap) => {
       this.queryParams = extractFiltersFromEncodedQuery(queryMap?.get('q'));
     });
     this.route.params.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(({ folderId }: Params) => {


projects/aca-content/src/lib/components/files/files.component.ts
```diff
@@ -61,6 +61,7 @@ import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
 import { SearchAiInputContainerComponent } from '../knowledge-retrieval/search-ai/search-ai-input-container/search-ai-input-container.component';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { HttpErrorResponse } from '@angular/common/http';
+import { extractFiltersFromEncodedQuery } from '../../utils/aca-search-utils';
```
AleksanderSklorz, line 64: Import path can be shorter, Intellij's hint: 
image

projects/aca-content/src/lib/components/files/files.component.spec.ts
```diff
@@ -190,8 +195,26 @@ describe('FilesComponent', () => {
       expect(router.navigate['calls'].argsFor(0)[0]).toEqual(['/personal-files', 'parent-id']);
     });
 
+    it('should set decoded query as queryParams', () => {
+      const initialQuery = { checkList: 'TYPE:"cm:folder"' };
+
+      const mockParamMap = getEncodedParamMap(initialQuery);
+
+      Object.defineProperty(component['route'], 'queryParamMap', {
```
AleksanderSklorz: We don't refer to private things by []. Please try to use TestBed.inject(ActivatedRoute) instead - you have already route variable in 53rd line which has assigned value from TestBed.inject

projects/aca-content/src/lib/components/files/files.component.spec.ts
```diff
@@ -190,8 +195,26 @@ describe('FilesComponent', () => {
...
+      fixture.detectChanges();
+
+      expect(component.queryParams).toEqual(initialQuery);
+    });
+
     it('should check isFilterHeaderActive to be true when filters are present in queryParamMap', () => {
-      Object.defineProperty(route, 'queryParamMap', { value: of({ params: { $thumbnail: 'TYPE:"cm:folder"' } }) });
+      const mockParamMap = getEncodedParamMap();
+
+      Object.defineProperty(component['route'], 'queryParamMap', {
```
AleksanderSklorz: We don't refer to private things by []. Please try to use TestBed.inject(ActivatedRoute) instead - you have already route variable in 53rd line which has assigned value from TestBed.inject

projects/aca-content/src/lib/components/files/files.component.ts
rmnvch: @AleksanderSklorz ,
that creates a circular and self-referential import path that breaks module resolution and bundling (refers own package from the inside)

AleksanderSklorz: ah ok, then leave it as it is now

AleksanderSklorz (approved):

rmnvch:
@AleksanderSklorz ,
for some reason your initial comment has gone. The issue was connected to lag after page reload with queryParams - filters are applied after initial API call:

https://github.com/user-attachments/assets/15a93a3a-49e0-43de-bc39-bd17623113f1

So the reason for that is that we have 2 APIs to work with folders - Search API and Nodes API ('/nodes/{nodeId}/children' endpoint). I've examined several options:
- Migrate to Nodes API fully. It can be used with folders/files filtering, but Nodes API has limits in terms of filtering be text queries - it supports only EQUALS operator in queries ('where' parameter) and cannot be used to filter according text queries - 'LIKE' or 'CONTAINS' are not supported as confirmed with BE team. 
-  Extend loader indicator. Filters are initiated after initial data is loaded, it means that table should be showed to make additional request. 

I suggest create dedicated ticket for exploring option with additional loader indicator. It would require some refactoring 

cc @MichalKinas

AleksanderSklorz:
> @AleksanderSklorz , for some reason your initial comment has gone. The issue was connected to lag after page reload with queryParams - filters are applied after initial API call:
> 
>  Screen.Recording.2025-10-06.at.13.07.47.mov 
> So the reason for that is that we have 2 APIs to work with folders - Search API and Nodes API ('/nodes/{nodeId}/children' endpoint). I've examined several options:
> 
> * Migrate to Nodes API fully. It can be used with folders/files filtering, but Nodes API has limits in terms of filtering be text queries - it supports only EQUALS operator in queries ('where' parameter) and cannot be used to filter according text queries - 'LIKE' or 'CONTAINS' are not supported as confirmed with BE team.
> * Extend loader indicator. Filters are initiated after initial data is loaded, it means that table should be showed to make additional request.
> 
> I suggest create dedicated ticket for exploring option with additional loader indicator. It would require some refactoring
> 
> cc @MichalKinas

Hi @rmnvch. My comment is still but not here. It's in alfresco-applications repository: 
https://github.com/Alfresco/alfresco-applications/pull/1258#issuecomment-3355208593

Regarding to solution: by "extend loader indicator" you mean adding something like "loading spinner" there, right? Yes, it makes sense for me. If you need more time you can do it as part of that PR, it's fine or you can create separate ticket if you prefer do it as separated one. In case you want to do it as separated task then please create jira task and add link to that task as comment here in PR

AleksanderSklorz:
@rmnvch I see now those filters don't work properly, for example - I select folder filter but I can't see anything. Folder should be displayed in that case: 

https://github.com/user-attachments/assets/fde7bc29-2a08-4b1f-b25e-f7bdb68c9183

projects/aca-content/src/lib/components/files/files.component.ts
```diff
@@ -120,8 +121,8 @@ export class FilesComponent extends PageComponent implements OnInit, OnDestroy {
 
     this.title = data.title;
 
-    this.route.queryParamMap.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((queryMap: Params) => {
-      this.queryParams = queryMap.params;
+    this.route.queryParamMap.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((queryMap: ParamMap) => {
```
AleksanderSklorz: we don't need a type here at all. In this case type can be detected automatically so let's remove that ParamMap type: 
image
As screenshot shows - it's detected automatically without setting type

rmnvch: fixed

rmnvch:
> @rmnvch I see now those filters don't work properly, for example - I select folder filter but I can't see anything. Folder should be displayed in that case:
> 
>  Screen.Recording.2025-10-13.at.11.45.19.mov

@AleksanderSklorz , hey. I believe it's temporary issue due to API's 500 error. I've encountered that last week too

Uploading sample error.mov…

rmnvch:
> > @AleksanderSklorz , for some reason your initial comment has gone. The issue was connected to lag after page reload with queryParams - filters are applied after initial API call:
> > Screen.Recording.2025-10-06.at.13.07.47.mov
> > So the reason for that is that we have 2 APIs to work with folders - Search API and Nodes API ('/nodes/{nodeId}/children' endpoint). I've examined several options:
> > 
> > * Migrate to Nodes API fully. It can be used with folders/files filtering, but Nodes API has limits in terms of filtering be text queries - it supports only EQUALS operator in queries ('where' parameter) and cannot be used to filter according text queries - 'LIKE' or 'CONTAINS' are not supported as confirmed with BE team.
> > * Extend loader indicator. Filters are initiated after initial data is loaded, it means that table should be showed to make additional request.
> > 
> > I suggest create dedicated ticket for exploring option with additional loader indicator. It would require some refactoring
> > cc @MichalKinas
> 
> Hi @rmnvch. My comment is still but not here. It's in alfresco-applications repository: [Alfresco/alfresco-applications#1258 (comment)](https://github.com/Alfresco/alfresco-applications/pull/1258#issuecomment-3355208593)
> 
> Regarding to solution: by "extend loader indicator" you mean adding something like "loading spinner" there, right? Yes, it makes sense for me. If you need more time you can do it as part of that PR, it's fine or you can create separate ticket if you prefer do it as separated one. In case you want to do it as separated task then please create jira task and add link to that task as comment here in PR

@AleksanderSklorz , I've finished updates. the changes are mainly in ADF's PR. https://github.com/Alfresco/alfresco-ng2-components/pull/11237

AleksanderSklorz:
@rmnvch it works better now but found one more problem. When you select both checkboxes and refresh page then you can't see anything. When you change page after reloading for example to libraries and you will return to personal files then page is broken as well: 

https://github.com/user-attachments/assets/d84f79c4-a5fa-4e8d-a7ca-7d8b33a481aa

rmnvch:
> @rmnvch please fix new found issue: [#4817 (comment)](https://github.com/Alfresco/alfresco-content-app/pull/4817#issuecomment-3405324755)

@AleksanderSklorz , thanks for pointing that out, my bad. fixed in ADF's PR.

PR #4824, opened 2025-10-02
[ACS-10409] Fix permission checking logic in library navigation

nikita-web-ua:
https://hyland.atlassian.net/browse/ACS-10409

### DIFF
--- a/projects/aca-content/src/lib/store/effects/library.effects.ts
+++ b/projects/aca-content/src/lib/store/effects/library.effects.ts
@@ -31,16 +31,18 @@ import {
   LibraryActionTypes,
   NavigateLibraryAction,
   NavigateRouteAction,
-  UpdateLibraryAction
+  UpdateLibraryAction,
+  isAdmin
 } from '@alfresco/aca-shared/store';
 import { inject, Injectable } from '@angular/core';
 import { Actions, createEffect, ofType } from '@ngrx/effects';
 import { Store } from '@ngrx/store';
-import { map, mergeMap, take } from 'rxjs/operators';
+import { catchError, map, mergeMap, take } from 'rxjs/operators';
 import { ContentApiService } from '@alfresco/aca-shared';
 import { ContentManagementService } from '../../services/content-management.service';
 import { NotificationService } from '@alfresco/adf-core';
 import { HttpErrorResponse } from '@angular/common/http';
+import { Observable, of } from 'rxjs';
 
 @Injectable()
 export class LibraryEffects {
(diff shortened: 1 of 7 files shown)


projects/aca-content/src/lib/store/effects/library.effects.ts
```diff
@@ -99,45 +101,59 @@ export class LibraryEffects {
...
         })
       ),
     { dispatch: false }
   );
 
+  private navigateToLibraryById(libraryId: string, route?: string): Observable<null> {
+    return this.contentApi.getNode(libraryId, { relativePath: '/documentLibrary' }).pipe(
+      map((node) => node.entry.id),
+      mergeMap((id) => {
+        const routePath = route ? route : 'libraries';
```
AleksanderSklorz: can be just 
const routePath = route || 'libraries';

AleksanderSklorz: or even better - maybe we can set libraries as default for route parameter - then condition is not needed at all 
private navigateToLibraryById(libraryId: string, route = 'libraries')

projects/aca-content/src/lib/store/effects/library.effects.ts
```diff
@@ -99,45 +101,59 @@ export class LibraryEffects {
...
-                  }
+        mergeMap((action) => {
+          const payload = action.payload;
+          if (payload && 'guid' in payload) {
+            return this.store.select(isAdmin).pipe(
+              take(1),
+              mergeMap((isUserAdmin) => {
+                if (!isUserAdmin && payload.visibility !== 'PUBLIC' && !payload.role) {
+                  this.notificationService.showError('APP.BROWSE.LIBRARIES.LIBRARY_NO_PERMISSIONS_WARNING');
+                  return of(null);
```
AleksanderSklorz: I see you have many return of(null) - here and also 127, 139 and 152 lines. I think you always return ```Observable<null>```. If there is no value and we always return null then can we just return nothing? Just have it as "void"? Similarly like above leaveLibrary action where we don't return anything.

AleksanderSklorz:
@nikita-web-ua just one question - I understand that you handle that case on frontend side? Wasn't there a possibility to handle that by backend because ideally backend should to check if user has permission to something and then return proper status code (403 ideally). Then only change for frontend should be handle that 403 code and display proper message. 

It would be ideal solution because I can imagine situation that some user can have custom application and can have custom logic for libraries - in that case user will get the same problem with our backend as we had.

nikita-web-ua:
> @nikita-web-ua just one question - I understand that you handle that case on frontend side? Wasn't there a possibility to handle that by backend because ideally backend should to check if user has permission to something and then return proper status code (403 ideally). Then only change for frontend should be handle that 403 code and display proper message.
> 
> It would be ideal solution because I can imagine situation that some user can have custom application and can have custom logic for libraries - in that case user will get the same problem with our backend as we had.

This was my thought as well, but Damian says it works as expected. For some reason there is a parent node for the site, and first we only have guid of this node, not site id directly. If we try to open the site we get expected error codes, but for this parent it is 404 if user has no permission. And it is expected,  because operation is based on "search" mechanism so since the user has no permission, a node cannot be found. The Search API works in a similar way, when user has no permission for a node, that node is not included in search results. This mechanism works for many years like that. That's why I had to add permission check on FE side.

AleksanderSklorz:
> > @nikita-web-ua just one question - I understand that you handle that case on frontend side? Wasn't there a possibility to handle that by backend because ideally backend should to check if user has permission to something and then return proper status code (403 ideally). Then only change for frontend should be handle that 403 code and display proper message.
> > It would be ideal solution because I can imagine situation that some user can have custom application and can have custom logic for libraries - in that case user will get the same problem with our backend as we had.
> 
> This was my thought as well, but Damian says it works as expected. For some reason there is a parent node for the site, and first we only have guid of this node, not site id directly. If we try to open the site we get expected error codes, but for this parent it is 404 if user has no permission. And it is expected, because operation is based on "search" mechanism so since the user has no permission, a node cannot be found. The Search API works in a similar way, when user has no permission for a node, that node is not included in search results. This mechanism works for many years like that. That's why I had to add permission check on FE side.

Ok, thank you for explanation. It's a bit strange that backend works like that to be honest because when I use some API and I don't have access then I expect something like 403 (not 404 because 404 is confusing in that case because 404 looks like "that library doesn't exist at all" but in reality it exists so user can be confused why his library "doesn't exist"). But ok, if for backend people it's fine to have those 404 then let's leave it and apply your solution.

projects/aca-content/src/lib/store/effects/library.effects.ts
nikita-web-ua: done, ty

projects/aca-content/src/lib/store/effects/library.effects.ts
nikita-web-ua: done

AleksanderSklorz (approved):

g-jaskowski (approved):

PR #4826, opened 2025-10-02
[ACS-10165] create all libraries component

g-jaskowski:
https://hyland.atlassian.net/browse/ACS-10165

### DIFF
--- a/projects/aca-content/src/lib/components/favorite-libraries/favorite-libraries.component.spec.ts
+++ b/projects/aca-content/src/lib/components/favorite-libraries/favorite-libraries.component.spec.ts
@@ -79,7 +79,7 @@ describe('FavoriteLibrariesComponent', () => {
   });
 
   it('should set data on error', () => {
-    spyOn(contentApiService, 'getFavoriteLibraries').and.returnValue(throwError('error'));
+    spyOn(contentApiService, 'getFavoriteLibraries').and.returnValue(throwError(() => 'error'));
     fixture.detectChanges();
 
     expect(component.list).toBe(null);
--- a/projects/aca-content/src/lib/components/libraries-base/libraries-base.component.ts
+++ b/projects/aca-content/src/lib/components/libraries-base/libraries-base.component.ts
@@ -86,11 +86,7 @@ export class LibrariesBaseComponent extends PageComponent {
 
   protected appHookService = inject(AppHookService);
 
-  constructor() {
-    super();
-  }
-
-  navigateTo(node: SiteEntry) {
+  private navigateTo(node: SiteEntry) {
     if (node?.entry?.guid) {
       this.store.dispatch(new NavigateLibraryAction(node.entry, this.navigateRoute));
     }
--- a/projects/aca-content/src/lib/components/libraries/libraries.component.ts
+++ b/projects/aca-content/src/lib/components/libraries/libraries.component.ts
@@ -33,10 +33,6 @@ import { LibrariesBaseComponent } from '../libraries-base/libraries-base.compone
   encapsulation: ViewEncapsulation.None
 })
 export class LibrariesComponent extends LibrariesBaseComponent implements OnInit {
-  constructor() {
-    super();
-  }
-
   ngOnInit() {
     super.ngOnInit();
 
--- a/projects/aca-content/src/lib/mock/libraries-mock.ts
+++ b/projects/aca-content/src/lib/mock/libraries-mock.ts
@@ -38,12 +38,12 @@ export const librariesMock: SitePaging = {
 export const favoriteLibrariesMock: FavoritePaging = {
   list: {
     entries: [
-      { entry: { id: '1', targetGuid: '1', title: 'Favorite Library 1', visibility: 'public' } },
-      { entry: { id: '2', targetGuid: '2', title: 'Favorite Library 2', visibility: 'private' } }
+      { entry: { targetGuid: '1', target: { id: '1', guid: '1', title: 'Favorite Library 1', visibility: 'public' } } },
+      { entry: { targetGuid: '2', target: { id: '2', guid: '2', title: 'Favorite Library 2', visibility: 'private' } } }
     ],
     pagination: { count: 25, skipCount: 0 }
   }
-} as any as FavoritePaging;
+};
 
 export const libraryColumnsPresetMock: DocumentListPresetRef[] = [
   {
--- a/projects/aca-content/src/lib/components/libraries-base/libraries-base.component.spec.ts
+++ b/projects/aca-content/src/lib/components/libraries-base/libraries-base.component.spec.ts
@@ -27,31 +27,31 @@ import { LibrariesBaseComponent } from './libraries-base.component';
 import { AppTestingModule } from '../../testing/app-testing.module';
 import { CustomEmptyContentTemplateDirective, PaginationComponent, UnitTestingUtils } from '@alfresco/adf-core';
 import { DocumentListComponent } from '@alfresco/adf-content-services';
-import { SiteEntry } from '@alfresco/js-api';
-import { Router } from '@angular/router';
 import { provideEffects } from '@ngrx/effects';
-import { RouterEffects } from '@alfresco/aca-shared/store';
+import { NavigateLibraryAction, RouterEffects } from '@alfresco/aca-shared/store';
 import { LibraryEffects } from '../../store/effects';
 import { getTitleElementText } from '../../testing/test-utils';
 import { librariesMock, libraryPaginationMock } from '../../mock/libraries-mock';
 import { DebugElement } from '@angular/core';
+import { MockStore, provideMockStore } from '@ngrx/store/testing';
+import { initialState } from '@alfresco/aca-shared';
 
 describe('LibrariesBaseComponent', () => {
   let component: LibrariesBaseComponent;
   let fixture: ComponentFixture<LibrariesBaseComponent>;
   let unitTestingUtils: UnitTestingUtils;
-  let router: Router;
+  let store: MockStore;
 
   beforeEach(async () => {
     await TestBed.configureTestingModule({
       imports: [AppTestingModule, LibrariesBaseComponent],
-      providers: [provideEffects([RouterEffects, LibraryEffects])]
+      providers: [provideEffects([RouterEffects, LibraryEffects]), provideMockStore({ initialState })]
     }).compileComponents();
     fixture = TestBed.createComponent(LibrariesBaseComponent);
     component = fixture.componentInstance;
 
-    router = TestBed.inject(Router);
     unitTestingUtils = new UnitTestingUtils(fixture.debugElement);
+    store = TestBed.inject(MockStore);
 
     component.list = librariesMock;
     fixture.detectChanges();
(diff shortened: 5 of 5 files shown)


projects/aca-content/assets/i18n/en.json
AleksanderSklorz, line 1: I think we would need to move translations for all languages from ADW because of 2 reasons: 
1. To avoid waiting for again translating them by translation team (so until that we would have lost languages). 
2. To avoid payment for the same translations again.

AleksanderSklorz:
@g-jaskowski I found issue probably. In ADW we could open sidebar for some libraries even when we are not member of them but now in ACA it's disabled. In ADW it was possible, the only thing is that we could not edit fields. 

ADW:
image

ACA: 
image

projects/aca-content/src/lib/components/library-list/library-list.component.html
```diff
@@ -0,0 +1,94 @@
...
+        [target]="documentList"
+        [pagination]="pagination"
+        (changePageSize)="onChangePageSize($event)"
+        (changePageNumber)="onChange($event)"
+        (nextPage)="onChange($event)"
+        (prevPage)="onChange($event)"
+      />
+    </div>
+
+    <div class="aca-sidebar" *ngIf="infoDrawerOpened$ | async">
```
AleksanderSklorz: This div is redundant - you can move class and *ngIf to set it on aca-info-drawer tag directly

projects/aca-content/src/lib/components/library-list/library-list.component.html
```diff
@@ -0,0 +1,94 @@
...
+          <adf-empty-content
+            icon="library_books"
+            [title]="'APP.BROWSE.LIBRARIES.EMPTY_STATE.ALL_LIBRARIES.TITLE'"
+            subtitle="APP.BROWSE.LIBRARIES.EMPTY_STATE.ALL_LIBRARIES.TEXT"
+          />
+        </adf-custom-empty-content-template>
+
+        <data-columns>
+          <ng-container *ngFor="let column of columns; trackBy: trackByColumnId">
+            <ng-container *ngIf="column.template && !(column.desktopOnly && isSmallScreen)">
```
AleksanderSklorz: For new code (and we can treat that as new code as we move it completely to ACA) we should to try to use new control flow: https://angular.dev/guide/templates/control-flow

Please apply for directives like *ngIf, *ngFor etc.

projects/aca-content/src/lib/components/library-list/library-list.component.ts
```diff
@@ -0,0 +1,142 @@
...
+    DataColumnListComponent,
+    DataColumnComponent,
+    DocumentListComponent,
+    CustomEmptyContentTemplateDirective
+  ],
+  templateUrl: './library-list.component.html',
+  encapsulation: ViewEncapsulation.None
+})
+export class LibraryListComponent extends PageComponent implements OnInit {
+  pagination: Pagination = new Pagination({
```
AleksanderSklorz: no need to set type when you initialize field inline

projects/aca-content/src/lib/components/library-list/library-list.component.ts
```diff
@@ -0,0 +1,142 @@
...
+    skipCount: 0,
+    maxItems: 25,
+    totalItems: 0
+  });
+  isLoading = false;
+  list: SitePaging;
+  columns: DocumentListPresetRef[] = [];
+
+  constructor(
+    private appHookService: AppHookService,
```
AleksanderSklorz: can we make those all parameters in constructor as readonly?

projects/aca-content/src/lib/components/library-list/library-list.component.ts
```diff
@@ -0,0 +1,142 @@
...
+  handleNodeClick(event: Event) {
+    this.navigateTo((event as CustomEvent).detail?.node);
+  }
+
+  onChangePageSize(pagination: Pagination) {
+    this.preferences.paginationSize = pagination.maxItems;
+    this.getList(pagination);
+  }
+
+  onChange(pagination: Pagination) {
```
AleksanderSklorz: we should not create function just to call other function. If you need to call that function in html then better to just make that inner function as public

projects/aca-content/src/lib/components/library-list/library-list.component.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    list: {
+      entries: [
+        { entry: { id: '1', guid: '1', title: 'Library 1', visibility: 'public' } },
+        { entry: { id: '2', guid: '2', title: 'Library 2', visibility: 'private' } }
+      ],
+      pagination: { count: 25, skipCount: 0 }
+    }
+  };
+
+  beforeEach(async () => {
```
AleksanderSklorz: async here seems redundant because I can't see any await in that block

projects/aca-content/src/lib/components/library-list/library-list.component.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+      userPreference.paginationSize = 1;
+      component.ngOnInit();
+      expect(sitesService.getSites).toHaveBeenCalledWith({ maxItems: userPreference.paginationSize });
+    });
+
+    it('should set data on error', () => {
+      getSitesSpy.and.returnValue(throwError(() => 'error'));
+      component.ngOnInit();
+
+      expect(component.list).toBe(null);
```
AleksanderSklorz: instead toBe(null) you can use toBeNull(); (for all places). There are also functions like toBeFalse() etc.

projects/aca-content/src/lib/components/library-list/library-list.component.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    });
+  });
+
+  describe('Node navigation', () => {
+    it('should not navigate when node is null or missing guid', () => {
+      spyOn(router, 'navigate').and.stub();
+      component.navigateTo(null);
+      expect(router.navigate).not.toHaveBeenCalled();
+
+      component.navigateTo({ entry: {} } as any);
```
AleksanderSklorz: instead as any please use as SiteEntry

AleksanderSklorz:
@g-jaskowski I believe we should to also move e2es related with moved code to ACA. If we won't move it now then ACA part of All Libraries won't be safe so somebody can merge some change which breaks it so better to move them as part of that PR to have it secured immediately

AleksanderSklorz:
@g-jaskowski I found important bug. 
1. Create Records Management lib in Share. 
2. Configure hruser to have access to that lib (example of configuration can be found in video). 
3. If you open that lib in ACA then you can see more available actions like for example Remove file from hold (but user has not capability for that added in user's role). 
4. Try the same in ADW - those options are hidden when user has not capability. They should be visible only when user has capabilities for that. 

You need to verify all actions on Records Management lib itself and everything what is inside like holds, files in holds etc. because we added multiple validation there which check roles. It was done as that MNT: https://hyland.atlassian.net/browse/MNT-24923 so we need that as it was MNT. 

Only problem which I see is that there may be a problem to check it properly because all validations are in ADW and I think ACA should not to know anything about holds, records etc. 


https://github.com/user-attachments/assets/acf5c5a8-7e22-46fb-b2d3-b55616e2aba5

AleksanderSklorz:
> @g-jaskowski I believe we should to also move e2es related with moved code to ACA. If we won't move it now then ACA part of All Libraries won't be safe so somebody can merge some change which breaks it so better to move them as part of that PR to have it secured immediately

@g-jaskowski Regarding to that comment about e2e - please check with Adam or Akash if you should to remove e2es in ADW which you move to ACA (and just leave e2es in ADW for extended part) OR if those e2es should be in both ACA and ADW

projects/aca-content/assets/i18n/en.json
g-jaskowski: All translations provided in ADW were copied to ACA. There was no tooltip translation.

g-jaskowski:
> @g-jaskowski I found issue probably. In ADW we could open sidebar for some libraries even when we are not member of them but now in ACA it's disabled. In ADW it was possible, the only thing is that we could not edit fields.
> 
> ADW: image
> 
> ACA: image

Issue fixed, details are accessible in info drawer



> @g-jaskowski I believe we should to also move e2es related with moved code to ACA. If we won't move it now then ACA part of All Libraries won't be safe so somebody can merge some change which breaks it so better to move them as part of that PR to have it secured immediately

All e2es using All Libraries page in ADW are related to ADW exclusive member management feature. There were no tests that could be moved, I've let Adam know that those are missing.



> @g-jaskowski I found important bug.
> 
> 1. Create Records Management lib in Share.
> 2. Configure hruser to have access to that lib (example of configuration can be found in video).
> 3. If you open that lib in ACA then you can see more available actions like for example Remove file from hold (but user has not capability for that added in user's role).
> 4. Try the same in ADW - those options are hidden when user has not capability. They should be visible only when user has capabilities for that.
> 
> You need to verify all actions on Records Management lib itself and everything what is inside like holds, files in holds etc. because we added multiple validation there which check roles. It was done as that MNT: https://hyland.atlassian.net/browse/MNT-24923 so we need that as it was MNT.
> 
> Only problem which I see is that there may be a problem to check it properly because all validations are in ADW and I think ACA should not to know anything about holds, records etc.
> 
>  Screen.Recording.2025-10-03.at.09.12.05.mov

I've adjusted the api call arguments so that only base template sites are fetched. Other available templated are related to Record Management.

AleksanderSklorz:
@g-jaskowski there is also component called libraries.component.ts - please check if we can reuse instead of adding new one

AleksanderSklorz:
> @g-jaskowski there is also component called libraries.component.ts - please check if we can reuse instead of adding new one

@g-jaskowski there left this one comment, please have a look

projects/aca-content/src/lib/components/favorite-libraries/favorite-libraries.component.spec.ts
```diff
@@ -23,189 +23,129 @@
...
+      maxItems: userPreference.paginationSize
     });
+  });
 
-    it('should set title based on selectedRowItemsCount', () => {
-      fixture.detectChanges();
-
-      expect(getTitleElementText(fixture)).toBe('APP.BROWSE.LIBRARIES.MENU.FAVORITE_LIBRARIES.TITLE');
+  it('should set data on error', () => {
+    spyOn(contentApiService, 'getFavoriteLibraries').and.returnValue(throwError('error'));
```
AleksanderSklorz: throwError is deprecated, please use something different 
image

projects/aca-content/src/lib/components/libraries-base/libraries-base.component.ts
```diff
@@ -0,0 +1,102 @@
...
+  @Input() navigateRoute: string;
+
+  @Output() changePageSize = new EventEmitter<Pagination>();
+  @Output() changePageNumber = new EventEmitter<Pagination>();
+  @Output() nextPage = new EventEmitter<Pagination>();
+  @Output() prevPage = new EventEmitter<Pagination>();
+
+  protected appHookService = inject(AppHookService);
+
+  constructor() {
```
AleksanderSklorz: I think no need for empty constructor with only super. It should be called automatically from parent class

projects/aca-content/src/lib/components/libraries-base/libraries-base.component.ts
```diff
@@ -0,0 +1,102 @@
...
+    super();
+  }
+
+  navigateTo(node: SiteEntry) {
```
AleksanderSklorz: I think that function is used only inside component ts file so should be private. It should not be public only to use it in unit tets. In unit tests it should be tested through public things so when handleNodeClick executes private navigateTo function then you test navigateTo logic by events actions from html which use handleNodeClick

projects/aca-content/src/lib/components/libraries/libraries.component.ts
```diff
@@ -22,56 +22,18 @@
...
   templateUrl: './libraries.component.html',
+  imports: [LibrariesBaseComponent],
   encapsulation: ViewEncapsulation.None
 })
-export class LibrariesComponent extends PageComponent implements OnInit {
-  columns: DocumentListPresetRef[] = [];
-
-  constructor(private appHookService: AppHookService) {
+export class LibrariesComponent extends LibrariesBaseComponent implements OnInit {
+  constructor() {
```
AleksanderSklorz: the same like previously - no need for empty constructor

projects/aca-content/src/lib/mock/libraries-mock.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export const favoriteLibrariesMock: FavoritePaging = {
+  list: {
+    entries: [
+      { entry: { id: '1', targetGuid: '1', title: 'Favorite Library 1', visibility: 'public' } },
+      { entry: { id: '2', targetGuid: '2', title: 'Favorite Library 2', visibility: 'private' } }
+    ],
+    pagination: { count: 25, skipCount: 0 }
+  }
+} as any as FavoritePaging;
```
AleksanderSklorz: is that possible to instead of using any - correct objects stored in that const? I think problem is that Favorite doesn't expect id field so can we just remove that field from mock?

projects/aca-content/src/lib/components/libraries-base/libraries-base.component.ts
g-jaskowski: removed, thanks!

projects/aca-content/src/lib/components/libraries-base/libraries-base.component.ts
g-jaskowski: function in now private and unit tests use events, thank you

projects/aca-content/src/lib/components/libraries/libraries.component.ts
g-jaskowski: removed as well, thanks!

projects/aca-content/src/lib/mock/libraries-mock.ts
g-jaskowski: it was a matter of moving library data to target field, thanks!

PR #4827, opened 2025-10-03
[ACS-10297] Add label and title to back button in about section

MichalKinas:
https://hyland.atlassian.net/browse/ACS-10297

### DIFF
--- a/projects/aca-content/about/src/about.component.html
+++ b/projects/aca-content/about/src/about.component.html
@@ -1,6 +1,10 @@
 <aca-page-layout>
   <div class="aca-page-layout-header">
-    <button mat-icon-button [routerLink]="landingPage">
+    <button
+      mat-icon-button
+      [attr.aria-label]="'APP.ABOUT.BACK_BUTTON_LABEL' | translate"
+      [attr.title]="'APP.ABOUT.BACK_BUTTON_LABEL' | translate"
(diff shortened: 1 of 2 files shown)


projects/aca-content/about/src/about.component.html
```diff
@@ -1,6 +1,10 @@
 <aca-page-layout>
   <div class="aca-page-layout-header">
-    <button mat-icon-button [routerLink]="landingPage">
+    <button
+      mat-icon-button
+      [attr.aria-label]="'APP.ABOUT.BACK_BUTTON_LABEL' | translate"
```
AleksanderSklorz, line 5: I think aria-label is redundant when you use title because title is read by screen reader too

MichalKinas: As far as I know some screen readers may not properly process the title so I think combining the title and aria-label is safer, in this approach we can be sure that element will have a tooltip and will be announced by all screen readers, adding: https://www.a11y-collective.com/blog/aria-label-vs-title/ for reference

AleksanderSklorz: ok, sounds good in that case

AleksanderSklorz (approved):

nikita-web-ua (approved):

PR #4834, opened 2025-10-07
[ACS-10286] Add label and aria expanded to the rule set header

MichalKinas:
https://hyland.atlassian.net/browse/ACS-10286

### DIFF
--- a/projects/aca-content/folder-rules/src/rule-list/rule-list/rule-list.ui-component.scss
+++ b/projects/aca-content/folder-rules/src/rule-list/rule-list/rule-list.ui-component.scss
@@ -1,26 +1,33 @@
+/* stylelint-disable selector-class-pattern */
+@use '@angular/material' as mat;
+
 .aca-rule-list {
-  display: flex;
-  flex-direction: column;
-  overflow-y: auto;
-  gap: 8px;
+  @include mat.expansion-overrides(
+    (
+      container-background-color: var(--theme-background-color),
+      header-text-color: var(--theme-text-color),
+      container-shape: 12px
+    )
+  );
 
   &__item {
-    display: flex;
-    flex-direction: column;
     border: 1px solid var(--theme-border-color);
     border-radius: 12px;
-    overflow: hidden;
+    /* stylelint-disable-next-line declaration-no-important */
+    box-shadow: none !important;
+
+    .mat-expansion-panel-body {
+      padding: 0;
+    }
 
     &__header {
-      display: flex;
-      flex-direction: row;
-      gap: 4px;
-      align-items: stretch;
-      cursor: pointer;
-      color: var(--theme-text-color);
-      user-select: none;
       font-size: 0.9em;
-      padding: 0.5em 1em;
+      padding: 0 16px;
+
+      &.mat-expanded {
+        border-bottom: 1px solid var(--theme-border-color);
+        border-radius: 0;
+      }
 
       & > * {
         display: flex;
@@ -38,12 +45,6 @@
         &__icon {
           transform: scale(0.8);
         }
-      }
-    }
-
-    &.aca-expanded {
-      .aca-rule-list__item__header {
-        border-bottom: 1px solid var(--theme-border-color);
       }
     }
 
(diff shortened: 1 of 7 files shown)


AleksanderSklorz:
@MichalKinas can you have a look and fix those sonar issues? 
image

MichalKinas:
> @MichalKinas can you have a look and fix those sonar issues? image

Yeah I saw them but I really wonder if we should fix them, I can change it to a button but then the looks will differ and I believe the author of the code used a div to mitigate potential issues of using a button here, what do you think?

AleksanderSklorz:
> > @MichalKinas can you have a look and fix those sonar issues? image
> 
> Yeah I saw them but I really wonder if we should fix them, I can change it to a button but then the looks will differ and I believe the author of the code used a div to mitigate potential issues of using a button here, what do you think?

Hmm, I see that button works like "expand/collapse". So we have also one more option instead using input or button we can use component from angular material - expansion panel. Why we have implemented own code here when we can use ready component from angular material? This should resolve accessibility + sonar issue together. What do you think?

MichalKinas:
> > > @MichalKinas can you have a look and fix those sonar issues? image
> > 
> > 
> > Yeah I saw them but I really wonder if we should fix them, I can change it to a button but then the looks will differ and I believe the author of the code used a div to mitigate potential issues of using a button here, what do you think?
> 
> Hmm, I see that button works like "expand/collapse". So we have also one more option instead using input or button we can use component from angular material - expansion panel. Why we have implemented own code here when we can use ready component from angular material? This should resolve accessibility + sonar issue together. What do you think?

It's done now

AleksanderSklorz:
@MichalKinas can we add little space between both panels? Actually there is no space but on ACA we had them separated: 
On branch: 
image

On ACA: 
image

AleksanderSklorz:
@MichalKinas before we had panels expanded by default, now they are collapsed. See video: 

https://github.com/user-attachments/assets/dd10c76b-f0af-422b-ba6a-4ac6dd6dd412

AleksanderSklorz:
@MichalKinas axe gives issue related with aria-hidden for those icons inside "Rules from linked folder". Can that be fixed? 
image

projects/aca-content/folder-rules/src/rule-list/rule-list/rule-list.ui-component.scss
```diff
@@ -1,26 +1,33 @@
...
+  );
 
   &__item {
-    display: flex;
-    flex-direction: column;
     border: 1px solid var(--theme-border-color);
     border-radius: 12px;
-    overflow: hidden;
+    /* stylelint-disable-next-line declaration-no-important */
+    box-shadow: none !important;
```
AleksanderSklorz, line 17: Do we need that important? We can try to have higher specificity

projects/aca-content/folder-rules/src/rule-list/rule-list/rule-list.ui-component.scss
```diff
@@ -1,26 +1,33 @@
...
+
+    .mat-expansion-panel-body {
```
AleksanderSklorz, line 19: Are we still using selectors from mat-selectors? If yes then can you replace hardcoded places with those selectors from that file (in for example this line and 27th line)?

MichalKinas:
> @MichalKinas can we add little space between both panels? Actually there is no space but on ACA we had them separated: On branch: image
> 
> On ACA: image

It won't look good with material accordion, it expects that all panels will be gathered at one when collapsed, additional space is there when they're expanded

MichalKinas:
> @MichalKinas before we had panels expanded by default, now they are collapsed. See video:
> 
>  Screen.Recording.2025-10-08.at.10.57.16.mov

Fixed now

MichalKinas:
> @MichalKinas axe gives issue related with aria-hidden for those icons inside "Rules from linked folder". Can that be fixed? image

Fixed too

projects/aca-content/folder-rules/src/rule-list/rule-list/rule-list.ui-component.scss
MichalKinas: We could but this would require couple Material selectors usage, this component is not extended anywhere so I think that using important here will be better than material selectors

projects/aca-content/folder-rules/src/rule-list/rule-list/rule-list.ui-component.scss
MichalKinas: I think they will be obsolote as soon as we will migrate to MD3, then most of them should be removed and replaced by MD3 tokens so I wouldn't add more to the selectors file for now, the ones I used are not there for now

PR #4911, opened 2025-11-27
[ACS-10607] A11y playwright test POC

akashrathod28:
https://hyland.atlassian.net/browse/ACS-10607


e2e/playwright/a11y-personal-files/src/tests/personal-files-core-a11y.e2e.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  test.afterAll(async () => {
+    await Utils.deleteNodesSitesEmptyTrashcan(nodesApi, trashcanApi, 'afterAll failed');
+  });
+
+  test('[XAT-19726] Page has no critical accessibility violations', async ({ page }) => {
+    try {
+      await checkA11y(page, undefined, { detailedReport: true });
+      expect(true).toBeTruthy();
+    } catch (error) {
```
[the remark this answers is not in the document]
akashrathod28: I have mark POC in comment to let know. this test will get modified once we start writing full test for a11y

projects/aca-playwright-shared/src/utils/a11y-validators.ts
```diff
@@ -0,0 +1,250 @@
...
+  const tagName = await element.evaluate((el) => el.tagName.toLowerCase());
+  const role = await element.getAttribute('role');
+  const tabindexAttr = await element.getAttribute('tabindex');
+  const tabindex = tabindexAttr === null ? null : parseInt(tabindexAttr, 10);
+  const hasFocusableTabIndex = tabindex !== null && !Number.isNaN(tabindex) && tabindex >= 0;
+
+  return INTERACTIVE_TAG_NAMES.includes(tagName) || INTERACTIVE_ROLES.includes(role || '') || hasFocusableTabIndex;
+}
+
+export async function hasRole(element: Locator, expectedRole: string): Promise<boolean> {
```
AleksanderSklorz: That file has multiple unused functions: 

Image

I understand some of them may be helpful in future but can we add them in moment of introducing usages for them? To avoid situation that some functions will leave unused

projects/aca-playwright-shared/src/utils/a11y-validators.ts
```diff
@@ -0,0 +1,250 @@
...
+
+export async function isValidDialog(element: Locator): Promise<boolean> {
+  const role = await element.getAttribute('role');
+  const hasLabel = await hasAccessibleAttribute(element);
+  const isDialog = role === 'dialog' || role === 'alertdialog';
+  return isDialog && hasLabel;
+}
+
+export async function getAccessibleName(element: Locator): Promise<string> {
+  const ariaLabel = await element.getAttribute('aria-label');
```
AleksanderSklorz: I believe that whole code can be shorter, just something like: 

```ts
  return (await element.getAttribute('aria-label')) || (await element.getAttribute('title')) || (await element.textContent()).trim() || '';

File: projects/aca-playwright-shared/src/utils/a11y-validators.ts [hunk 4]
```diff
@@ -0,0 +1,250 @@
...
+      }
+    }
+  }
+  return true;
+}
+
+export async function verifyRegionAccessibleNames(
+  page: Page,
+  regionSelector: string,
+  elementSelector: string = 'button, [role="button"], a'
```
AleksanderSklorz: no need for type when default value is assigned

e2e/playwright/a11y-personal-files/src/tests/personal-files-core-a11y.e2e.ts
```diff
@@ -0,0 +1,166 @@
...
+        const nodeDetails = v.nodes.map((n) => `[${n.target.join('>')}]`).join(' | ');
+        // eslint-disable-next-line no-console
+        console.log(`🔴 ${v.id} (${v.impact}) - ${v.nodes.length} node(s): ${nodeDetails}`);
+      });
+      expect(violations.length).toBe(1);
+    }
+    expect(violations).toBeDefined();
+  });
+
+  test('[XAT-19755] Create folder dialog has ARIA attributes', async ({ page, personalFiles }) => {
```
AleksanderSklorz, line 134: Isn't that detected by Axe automatically or Axe doesn't detect this kind of issues? (I thought it can detect missing arias as it's just html validation?)

projects/aca-playwright-shared/src/utils/a11y-validators.ts
AleksanderSklorz: BTW @akashrathod28 I'm also wondering for future if some of these functions won't be redundant because for example checking role can be probably done by Axe itself so for example hasRole will be redundant?

e2e/playwright/a11y-personal-files/src/tests/personal-files-core-a11y.e2e.ts
akashrathod28: you're right, axe validates ARIA that's present but won't flag a dialog that's missing [role="dialog"]altogether (that's not "invalid," just absent). It also can't test behavior like Escape-key closing. So this test covers two things axe can't: (1) the correct role/label were actually applied, and (2) keyboard interaction works. I've also adding a scoped [checkA11y()] on the dialog itself in next commit, since the page-level scan runs before the dialog opens and was missing it entirely.

AleksanderSklorz: ok, got it, thank you for explanation

projects/aca-playwright-shared/src/utils/a11y-validators.ts
akashrathod28: removed all unused functions: hasRole, hasRoleOrLabel, isFocusable, isValidDialog, getElementInfo, verifyFormInputsHaveLabels, verifyHeadingHierarchy, verifyLandmarks. Kept only the 7 that are actually called by the test suite. Agreed on your second point too — will add new helpers only when a real test needs them going forward, rather than pre-building a generic toolkit. And yes, hasRole was likely redundant with what axe already validates (aria-allowed-role, etc.), so good instinct there.

projects/aca-playwright-shared/src/utils/a11y-validators.ts
```diff
@@ -0,0 +1,125 @@
...
+  const tagName = await element.evaluate((el) => el.tagName.toLowerCase());
+  const role = await element.getAttribute('role');
+  const tabindexAttr = await element.getAttribute('tabindex');
+  const tabindex = tabindexAttr === null ? null : parseInt(tabindexAttr, 10);
+  const hasFocusableTabIndex = tabindex !== null && !Number.isNaN(tabindex) && tabindex >= 0;
+
+  return INTERACTIVE_TAG_NAMES.includes(tagName) || INTERACTIVE_ROLES.includes(role || '') || hasFocusableTabIndex;
+}
+
+export async function getAccessibleName(element: Locator): Promise<string> {
```
AleksanderSklorz: @akashrathod28 you forgot to remove this one. It's unused too now

AleksanderSklorz (approved):

PR #5050, opened 2026-02-17
[ACS-11158]: style update

rmnvch:
https://hyland.atlassian.net/browse/ACS-11158

relates to https://github.com/Alfresco/alfresco-ng2-components/pull/11659

### DIFF
--- a/projects/aca-content/src/lib/ui/theme.scss
+++ b/projects/aca-content/src/lib/ui/theme.scss
@@ -336,6 +336,13 @@ html {
       )
     );
   }
+
+  @include mat.tooltip-overrides(
+    (
+      container-color: var(--adf-accessibility-tooltip-background-color),
+      supporting-text-size: 14px
+    )
+  );
 }
 
 .adf-search-properties-form,


MichalKinas (approved):

projects/aca-content/src/lib/ui/theme.scss
```diff
@@ -30,6 +30,12 @@
       subscript-text-size: 12px
     )
   );
+  @include mat.tooltip-overrides(
```
AleksanderSklorz, line 340: We have already some styling in ADF for tooltips. Do we need both or can we close ACA's PR or maybe it would be better to move that change to ADF instead? 

https://github.com/Alfresco/alfresco-ng2-components/blob/develop/lib/core/src/lib/styles/_globals.scss#L91

projects/aca-content/src/lib/ui/theme.scss
```diff
@@ -30,6 +30,12 @@
...
+    (
+      container-color: var(--adf-accessibility-tooltip-background-color),
+      supporting-text-size: 14px
```
AleksanderSklorz, line 343: Earlier we used 
var(--theme-body-2-font-size)

Do we need to hardcode font size or can we use that variable again?

projects/aca-content/src/lib/ui/theme.scss
rmnvch: @AleksanderSklorz ,
I'm removing that in https://github.com/Alfresco/alfresco-ng2-components/pull/11659

projects/aca-content/src/lib/ui/theme.scss
rmnvch: @AleksanderSklorz , I thought it's not a good idea to tie tooltip font size to body font size because theorethically body size can be changed without tooltip's sizing. So it was either hardcoding or new var

AleksanderSklorz (approved):

PR #5089, opened 2026-03-11
[ACS-9765] Records Management lib and its content action restrictions

rmnvch:
https://hyland.atlassian.net/browse/ACS-9765

key changes:

- new app rule to hide/show move/copy action for rma system folders
- set of util functions to identify rma-related entities
- injection token for commenting service with multi flag to make possible adding on ADW side its own

### DIFF
--- a/projects/aca-content/src/lib/services/aca-node-permission-comments.service.ts
+++ b/projects/aca-content/src/lib/services/aca-node-permission-comments.service.ts
@@ -1,0 +1,47 @@
+/*!
+ * Copyright © 2005-2025 Hyland Software, Inc. and its affiliates. All rights reserved.
+ *
+ * Alfresco Example Content Application
+ *
+ * This file is part of the Alfresco Example Content Application.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail. Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * The Alfresco Example Content Application is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * The Alfresco Example Content Application is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * from Hyland Software. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { Node } from '@alfresco/js-api';
+import { ExternalNodePermissionCommentsTabService } from '@alfresco/aca-shared';
+
+type DenyRule = (node: Node) => boolean;
+
+export class AcaNodePermissionCommentsService implements ExternalNodePermissionCommentsTabService {
+  private readonly blacklistedTypes = new Set<string>(['rma:hold']);
+  private readonly blacklistedAspects = new Set<string>(['rma:frozen']);
+
+  private readonly denyRules: ReadonlyArray<DenyRule> = [(node) => this.hasBlacklistedAspect(node), (node) => this.hasBlacklistedType(node)];
+
+  canAddComments(node: Node): boolean {
+    return this.denyRules.every((rule) => !rule(node));
+  }
+
+  private hasBlacklistedAspect(node: Node): boolean {
+    return (node.aspectNames ?? []).some((aspect) => this.blacklistedAspects.has(aspect));
+  }
+
+  private hasBlacklistedType(node: Node): boolean {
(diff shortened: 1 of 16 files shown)


projects/aca-shared/src/lib/utils/node.utils.ts
```diff
@@ -43,3 +52,22 @@ export function isLibrary(node: { entry: Node | any }): boolean {
...
+const isRmaTransferContainer = hasNodeType(RMA_NODE_TYPES.transferContainer);
+const isRmaUnfiledRecordsContainer = hasNodeType(RMA_NODE_TYPES.unfiledRecordContainer);
+
+export const isRmaSystemFolder = (node: Node): boolean =>
+  isRmaHoldsContainer(node) || isRmaTransferContainer(node) || isRmaUnfiledRecordsContainer(node);
+
+export const isRmaRestrictedCreateFolder = (node: Node): boolean => isRmaHold(node) || isRmaTransferContainer(node) || isRmaHoldsContainer(node);
+
+export const isRmaContent = (node: Node): boolean => {
+  return node.nodeType.startsWith('rma:');
```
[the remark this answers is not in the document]
rmnvch: `Node` in our API contract requires `[nodeType: string] `
(@alfresco/js-api) so `isRmaContent(node: Node)` intentionally assumes a valid object and does not guard for missing `[nodeType]`

projects/aca-shared/src/lib/utils/node.utils.ts
```diff
@@ -43,3 +52,22 @@ export function isLibrary(node: { entry: Node | any }): boolean {
...
   }
 }
+
+const hasNodeType =
+  (type: RmaNodeType) =>
+  (node: Node): boolean =>
+    node.nodeType === type;
+
+const isRmaHoldsContainer = hasNodeType(RMA_NODE_TYPES.holdContainer);
+const isRmaHold = hasNodeType(RMA_NODE_TYPES.hold);
```
AleksanderSklorz, line 62: General comment but please apply to everything - ACA should not know anything about Records Managements, Holds etc. Those features are ADW related features which is paid. Please move all checks to ADW. It should be possible because ADW extends ACA. 

I know that in ACA you can find All Libraries and there is Records Management lib but you don't need to care about that because this lib won't be visible in production for customers using ACA. Only customers using ADW will see that lib because that lib is not available for customers using community version of ACS.

projects/aca-content/src/lib/services/aca-node-permission-comments.service.ts
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * from Hyland Software. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { Node } from '@alfresco/js-api';
+import { ExternalNodePermissionCommentsTabService } from '@alfresco/aca-shared';
+
+type DenyRule = (node: Node) => boolean;
+
+export class AcaNodePermissionCommentsService implements ExternalNodePermissionCommentsTabService {
```
AleksanderSklorz, line 30: As it's just service then can't we just use @Injectable?

projects/aca-shared/src/lib/utils/node.utils.ts
```diff
@@ -43,3 +52,22 @@ export function isLibrary(node: { entry: Node | any }): boolean {
...
+const isRmaTransferContainer = hasNodeType(RMA_NODE_TYPES.transferContainer);
+const isRmaUnfiledRecordsContainer = hasNodeType(RMA_NODE_TYPES.unfiledRecordContainer);
+
+export const isRmaSystemFolder = (node: Node): boolean =>
```
AleksanderSklorz, line 66: I see you are checking mainly type of node but please note that except that we have also some extra permissions which can be set in Share: 
If you go to Share -> Sites -> Records Management then you can for example: 
1. Set permissions for File Plan/Holds/Transfers/Unifiled Records using Manage Permissions option for each: 

Image

2. In RM Admin Tool you can create roles with different capabilities and then assign users/groups to them: 

Image

We need to ensure that some permission is not required to do some action because then we need to allow for specific action if user has required permission

projects/aca-shared/src/lib/utils/node.utils.ts
rmnvch: @AleksanderSklorz , that's a bit confusing: 
`in ACA you can find All Libraries and there is Records Management lib .....  this lib won't be visible in production for customers using ACA` because that was the reason I decided to apply that to ACA. I believe we should hide RM for ACA then

AleksanderSklorz: Ok let me explain it. 
We have two versions of our app - community (free) and enterprise (paid). Records Management is available only in paid version (enterprise) and to create that lib you need have paid ACS (backend). 
You can see that lib in ACA because when you create Records Management then you create it by paid API and then ACA loads all libs (including Records Management). But in real life if user has free license then it can't create Records Management (as it's only available in paid version) and user who has paid license is not using ACA because is using ADW. Hence you don't need to do anything related with RM (and ACA should not be aware about that) because in real life that case won't be possible. It's possible only for us

PR #5116, opened 2026-03-27
[ACS-10732] After switching the language and reloading, the app reverts many components back to English.

dominikiwanekhyland:
https://hyland.atlassian.net/browse/ACS-10732

### DIFF
--- a/projects/aca-content/src/lib/components/search/search-save/sidenav/save-search-sidenav.component.spec.ts
+++ b/projects/aca-content/src/lib/components/search/search-save/sidenav/save-search-sidenav.component.spec.ts
@@ -22,13 +22,14 @@
  * from Hyland Software. If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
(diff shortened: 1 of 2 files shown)


projects/aca-content/src/lib/components/search/search-save/sidenav/save-search-sidenav.component.spec.ts
```diff
@@ -94,7 +95,26 @@ describe('SaveSearchSidenavComponent', () => {
       url: 'search?q=abc',
       id: 'search1'
     });
-  }));
+  });
+
+  it('should update title when language changes', () => {
+    const langChangeSubject = new Subject<any>();
```
AleksanderSklorz: instead of any can we use LangChangeEvent to align type with original type?

Image

AleksanderSklorz:
@dominikiwanekhyland this copilot's comment should be fixed because it seems valid: https://github.com/Alfresco/alfresco-content-app/pull/5116#discussion_r3001027041. 
We have set "teardown: { destroyAfterEach: false }" in test.ts so components, services etc. are not automatically destroyed and because of that that mock can pass to other tests (and affect them) if won't be reset manually. Now it may be success but for different order of tests (or when we add more tests) it may fail.

AleksanderSklorz (approved):

PR #5134, opened 2026-04-03
[ACS-11446] Automated last test cases from ACA test execution & various refactoring

datguychen:
https://hyland.atlassian.net/browse/ACS-11446


projects/aca-playwright-shared/src/utils/logger.ts
```diff
@@ -24,65 +24,29 @@
...
 export interface LoggerLike {
-  info(...messages: string[]): void;
-  log(...messages: string[]): void;
-  warn(...messages: string[]): void;
-  error(...messages: string[]): void;
+  info(message: string): void;
+  log(message: string): void;
+  warn(message: string): void;
+  error(message: string): void;
+  table(message: object): void;
```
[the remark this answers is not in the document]
datguychen: I don't like using unknown, so I will skip that proposition

projects/aca-playwright-shared/src/api/tags-api.ts
```diff
@@ -38,59 +39,75 @@ export class TagsApi {
...
+        }
+        logger.info(`Tag created: "${created.entry.tag}" (id: ${created.entry.id})`);
+        results.push(created);
+      }
+      return results;
     } catch (error) {
-      console.error(error);
-      return null;
+      throw new Error(`Failed to create tags: ${error}`);
     }
```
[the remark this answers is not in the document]
datguychen: this is not needed in my opinon.

e2e/playwright/info-drawer/src/tests/edit-mode.e2e.ts
```diff
@@ -0,0 +1,197 @@
...
+
+        await fileActionsApi.waitForNodes(folder938, { expect: 1 });
+      } catch (error) {
+        console.error(`beforeAll failed : ${error}`);
+        throw error;
+      }
+    });
+
+    test.afterAll(async () => {
+      for (const tag of [tag938]) {
```
AleksanderSklorz: Couple comments to that. 

In multiple places I see you are iterating over array with only one tag instead just pass that tag to ```deleteTagsByTagName``` function. That function gets single string parameter so it can be just ```tagsApi.deleteTagsByTagName(tag938);``` .

Moreover if you are going to remove only one tag per function's call I would rename that function to deleteTagByTagName instead of deleteTagsByTagName because in reality that function removes only single tag, not many tags. If you wanted to implement that function to remove many tags by one call then this name can stay but please move array to inside of that function to avoid repeating array each time.

e2e/playwright/search/src/tests/search-filters-tags.e2e.ts
```diff
@@ -33,7 +32,7 @@ test.describe('Search - Filters - Tags', () => {
   let fileActionsApi: FileActionsApi;
   let file1Id: string;
   let file2Id: string;
-  let tagPaging: TagPaging;
+  let createdTags: { entry: { id: string; tag: string } }[];
```
AleksanderSklorz: Why you are defining new type instead of reuse existing type for tags? Type should be ```TagEntry[]```

projects/aca-playwright-shared/src/api/nodes-api.ts
```diff
@@ -119,9 +115,9 @@ export class NodesApi {
     imageProps: any = null,
     author: string = '',
     majorVersion: boolean = true,
-    aspectNames: string[] = null
-  ): Promise<NodeEntry | null> {
-    if (!aspectNames) {
+    aspectNames: string[] = []
+  ): Promise<NodeEntry> {
+    if (aspectNames.length === 0) {
       aspectNames = ['cm:versionable']; // workaround for REPO-4772
```
AleksanderSklorz: This case is some workaround. Can we check if that workaround is still needed?

projects/aca-playwright-shared/src/api/tags-api.ts
```diff
@@ -38,59 +39,75 @@ export class TagsApi {
...
     try {
-      return this.apiService.tagsApi.createTags(tags);
+      const results: TagEntry[] = [];
+      for (const tag of tagNames) {
+        const result = await this.apiService.tagsApi.createTags([{ tag }]);
+        let created: TagEntry;
+        if ('entry' in result) {
+          created = result as TagEntry;
+        } else if ('list' in result) {
+          const firstEntry = (result as TagPaging).list?.entries?.[0];
```
AleksanderSklorz: (result as TagPaging) type assertion is redundant. You already check if 'list' is inside result and 'list' exists only for TagPaging so typescript automatically knows that when you enter inside that 'if' block then it must be TagPaging. So it can be just:
```ts
const firstEntry = result.list?.entries?.[0];

File: projects/aca-playwright-shared/src/api/tags-api.ts [hunk 7]
```diff
@@ -38,59 +39,75 @@ export class TagsApi {
...
+      throw new Error(`Failed to list tags: ${error}`);
     }
   }
 
   async deleteTagsByTagName(tagName: string): Promise<void> {
     try {
       const response = await this.listTags({ tag: tagName, matching: true });
-      const tagIds = response.list.entries.map((entry) => entry.entry.id);
-      await this.deleteTags(tagIds);
+      const tags = response.list?.entries.map((entry) => ({ id: entry.entry.id, tag: entry.entry.tag })) || [];
```
AleksanderSklorz: That code creates new type for 'tags': 

Image

You see, ts detects is as something like ```{ id: string, tag: string }[]```. 

Also creating own object inside map by ```({ id: entry.entry.id, tag: entry.entry.tag })``` is redundant. If you check then 'entry' field contains what you need exactly. So that line can be changed to: 
```const tags = response.list?.entries.map((entry) => entry.entry) || [];```

After that you can see that code is shorter (no creating custom object) + type is better: 

Image

File: projects/aca-playwright-shared/src/api/tags-api.ts [hunk 8]
```diff
@@ -38,59 +39,75 @@ export class TagsApi {
...
       return this.apiService.tagsApi.assignTagToNode(nodeId, tag);
     } catch (error) {
-      console.error(error);
-      return null;
+      throw new Error(`Failed to assign tag to node: ${error}`);
     }
   }
 
-  async deleteTags(tagIds: string[]): Promise<void> {
+  async deleteTags(...tags: { id: string; tag?: string }[]): Promise<void> {
```
AleksanderSklorz: We have already type for this (Tag). No need to create new custom type. Just change to: 
```async deleteTags(...tags: Tag[]): Promise<void>```

File: projects/aca-playwright-shared/src/page-objects/components/edit-mode.component.ts [hunk 9]
```diff
@@ -22,22 +22,25 @@
...
 
-  async waitForReload(): Promise<void> {
-    try {
-      await this.getChild('').waitFor({ state: 'attached', timeout: 2000 });
-      await this.getChild('').waitFor({ state: 'detached', timeout: 2000 });
-    } catch (e) {
-      this.logger.info('Spinner was not present');
-    }
-  }
+  public tagsAccordion = this.page.locator('[data-automation-id="adf-content-metadata-tags-panel"]');
```
AleksanderSklorz, line 31: Please move fields above constructor

AleksanderSklorz (approved):

PR #5140, opened 2026-04-08
[ACS-10621] library-properties cleanup

datguychen:
https://hyland.atlassian.net/browse/ACS-10621

### DIFF
--- a/e2e/playwright/info-drawer/src/tests/library-properties.e2e.ts
+++ b/e2e/playwright/info-drawer/src/tests/library-properties.e2e.ts
@@ -222,7 +222,6 @@ test.describe('Non manager', () => {
   let sitesApi: SitesApi;
 
   const username = `user1-${Utils.random()}`;
-  const user2 = `user2-${Utils.random()}`;
   const user3 = `user3-${Utils.random()}`;
   const site = {
     name: `site1-${Utils.random()}`,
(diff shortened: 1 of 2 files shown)


AleksanderSklorz:
@datguychen this Copilot's comment looks not addressed, do we really want remove that test case?
https://github.com/Alfresco/alfresco-content-app/pull/5140#discussion_r3050803516

datguychen:
> @datguychen this Copilot's comment looks not addressed, do we really want remove that test case? [#5140 (comment)](https://github.com/Alfresco/alfresco-content-app/pull/5140#discussion_r3050803516)

@AleksanderSklorz Yes, this test case was outdated and effectively tested nothing. It was originally created based on this ticket: https://hyland.atlassian.net/browse/ACA-4335#icft=ACA-4335, but that ticket explicitly decided not to implement the functionality because of lack of time.

The test itself only checked whether something was hidden, which is an incorrect approach. It did not verify that the check was performed in the correct place or at the correct moment in the test flow. As a result, it always returned a false‑positive outcome without actually validating any behavior.

AleksanderSklorz:
> > @datguychen this Copilot's comment looks not addressed, do we really want remove that test case? [#5140 (comment)](https://github.com/Alfresco/alfresco-content-app/pull/5140#discussion_r3050803516)
> 
> @AleksanderSklorz Yes, this test case was outdated and effectively tested nothing. It was originally created based on this ticket: https://hyland.atlassian.net/browse/ACA-4335#icft=ACA-4335, but that ticket explicitly decided not to implement the functionality because of lack of time.
> 
> The test itself only checked whether something was hidden, which is an incorrect approach. It did not verify that the check was performed in the correct place or at the correct moment in the test flow. As a result, it always returned a false‑positive outcome without actually validating any behavior.

ok, thank you for explanation, sounds good to me

AleksanderSklorz (approved):

PR #5206, opened 2026-05-29
ACS-11861 Sharable link redirects to Login Page

rmnvch:
https://hyland.atlassian.net/browse/ACS-11861

### DIFF
--- a/projects/aca-shared/src/lib/services/app.service.ts
+++ b/projects/aca-shared/src/lib/services/app.service.ts
@@ -52,6 +52,7 @@ import { ShellAppService } from '@alfresco/adf-core/shell';
 import { AppSettingsService } from './app-settings.service';
 import { UserProfileService } from './user-profile.service';
 import { MatDialog } from '@angular/material/dialog';
+import { minimatch } from 'minimatch';
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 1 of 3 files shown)


projects/aca-shared/src/lib/services/app.service.ts
```diff
@@ -52,6 +52,7 @@ import { ShellAppService } from '@alfresco/adf-core/shell';
 import { AppSettingsService } from './app-settings.service';
 import { UserProfileService } from './user-profile.service';
 import { MatDialog } from '@angular/material/dialog';
+import { minimatch } from 'minimatch';
```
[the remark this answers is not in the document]
rmnvch: fixed

MichalKinas, line 55: Do we actually need minimatch here? I can recall some PR that removed this dependency

rmnvch: I can assume it might have been "mismatch-browser" lib, from ACA or ADW because it's not needed anymore.
but If it was the "minimatch" exactly - that makes sense too, because it wasn't used in ACA directly.
in this case we deal with glob patterns from config file, not plain string paths:
```
**/preview/s/*
**/settings
```
A plain regex would work as well, but it is usually less safe/maintainable here.

There is also another option - to bring dependency on 
`OidcAuthenticationService` from ADF, but that will create another coupling (avoiding minimatch dependency though) 
I believe current approach is better

MichalKinas (approved):

PR #5236, opened 2026-06-17
[ACS-11973] Fix: context menu disappears during bulk upload

nikita-web-ua:
https://hyland.atlassian.net/browse/ACS-11973

### DIFF
--- a/projects/aca-content/src/lib/components/favorites/favorites.component.spec.ts
+++ b/projects/aca-content/src/lib/components/favorites/favorites.component.spec.ts
@@ -31,7 +31,7 @@ import { AppTestingModule } from '../../testing/app-testing.module';
 import { AppService, ContentApiService } from '@alfresco/aca-shared';
 import { getTitleElementText } from '../../testing/test-utils';
 import { MatSnackBarModule } from '@angular/material/snack-bar';
-import { testHeader } from '../../testing/document-base-page-utils';
+import { testHeader, testUploadEvents } from '../../testing/document-base-page-utils';
 
 describe('FavoritesComponent', () => {
   let fixture: ComponentFixture<FavoritesComponent>;
@@ -143,4 +143,9 @@ describe('FavoritesComponent', () => {
   });
 
   testHeader(FavoritesComponent);
+
+  testUploadEvents(
+    () => component,
+    () => fixture
+  );
 });
--- a/projects/aca-content/src/lib/components/favorites/favorites.component.ts
+++ b/projects/aca-content/src/lib/components/favorites/favorites.component.ts
@@ -80,8 +80,8 @@ export class FavoritesComponent extends PageComponent implements OnInit {
     super.ngOnInit();
 
     this.subscriptions = this.subscriptions.concat([
-      this.uploadService.fileUploadComplete.pipe(debounceTime(300)).subscribe(() => this.reload()),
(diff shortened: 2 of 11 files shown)


MichalKinas (approved):

AleksanderSklorz:
@nikita-web-ua I found issue - when we have bulk uploading and we open context menu for file, then when folder appears it displays context menu but empty, see video 


https://github.com/user-attachments/assets/32f81b20-0dfd-42f9-96bd-e3602b00330a

nikita-web-ua:
> @nikita-web-ua I found issue - when we have bulk uploading and we open context menu for file, then when folder appears it displays context menu but empty, see video
> 
>  Screen.Recording.2026-06-18.at.13.10.11.mov

good catch, fixed

AleksanderSklorz:
@nikita-web-ua if you call View action to open file then viewer changes displayed file until everything is loaded. Should we fix this as well? 


https://github.com/user-attachments/assets/2f014d54-8bbf-4f0e-b029-360d8f6f5e86

nikita-web-ua:
> @nikita-web-ua if you call View action to open file then viewer changes displayed file until everything is loaded. Should we fix this as well?
> 
>  Screen.Recording.2026-06-19.at.09.53.16.mov

I can see the same behaviour on the acadev if you open a file during upload, so it is not related with a fix. Should we create a separate ticket for that?

AleksanderSklorz:
> > @nikita-web-ua if you call View action to open file then viewer changes displayed file until everything is loaded. Should we fix this as well?
> > Screen.Recording.2026-06-19.at.09.53.16.mov
> 
> I can see the same behaviour on the acadev if you open a file during upload, so it is not related with a fix. Should we create a separate ticket for that?

Ok then please create new task for that. It's strange behaviour because it should just show file for which action was triggered.  It may be caused because you can upload new version of file and it can't recognize if there is uploading new file or new version of file so probably we will need to add some condition which allows to recognize that.

AleksanderSklorz (approved):

nikita-web-ua:
> > > @nikita-web-ua if you call View action to open file then viewer changes displayed file until everything is loaded. Should we fix this as well?
> > > Screen.Recording.2026-06-19.at.09.53.16.mov
> > 
> > 
> > I can see the same behaviour on the acadev if you open a file during upload, so it is not related with a fix. Should we create a separate ticket for that?
> 
> Ok then please create new task for that. It's strange behaviour because it should just show file for which action was triggered. It may be caused because you can upload new version of file and it can't recognize if there is uploading new file or new version of file so probably we will need to add some condition which allows to recognize that.

created https://hyland.atlassian.net/browse/ACS-12047

AleksanderSklorz:
> > > > @nikita-web-ua if you call View action to open file then viewer changes displayed file until everything is loaded. Should we fix this as well?
> > > > Screen.Recording.2026-06-19.at.09.53.16.mov
> > > 
> > > 
> > > I can see the same behaviour on the acadev if you open a file during upload, so it is not related with a fix. Should we create a separate ticket for that?
> > 
> > 
> > Ok then please create new task for that. It's strange behaviour because it should just show file for which action was triggered. It may be caused because you can upload new version of file and it can't recognize if there is uploading new file or new version of file so probably we will need to add some condition which allows to recognize that.
> 
> created https://hyland.atlassian.net/browse/ACS-12047

Thank you

PR #5256, opened 2026-06-30
[ACS-12075] more multibrowser stabilization pt 3

adamswiderskipw:
https://hyland.atlassian.net/browse/ACS-12075


projects/aca-playwright-shared/src/base-config/playwright.config.ts
```diff
@@ -159,7 +159,7 @@ export const getGlobalConfig: PlaywrightTestConfig = {
   /* Retry on CI only */
   retries: env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: 3,
+  workers: Number(env.PLAYWRIGHT_WORKERS) || 3,
```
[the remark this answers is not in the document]
adamswiderskipw: non-integer or 0 will never be used

projects/aca-playwright-shared/src/api/search-api.ts
```diff
@@ -58,6 +58,115 @@ export class SearchApi {
     }
   }
 
+  async searchForNode(fileName: string, options?: { maxRetries?: number }): Promise<ResultSetPaging> {
+    const query = {
+      query: {
+        query: `cm:name:"${fileName}"`,
+        language: 'afts'
+      },
```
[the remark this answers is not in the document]
adamswiderskipw: these characters are not allowed for file names.

.github/actions/deploy-local-acs/action.yml
```diff
@@ -116,14 +120,49 @@ runs:
...
+          fi
+          echo "Retrying in $((attempt * 10))s..."
+          sleep $((attempt * 10))
+        done
+
+        helm install acs acs-deployment/helm/alfresco-content-services \
+          --set global.search.sharedSecret="$(openssl rand -hex 24)" \
+          --values acs-deployment/test/enterprise-integration-test-values.yaml \
+          --values .github/acs-deployment-values-override.yaml \
+          "${EXTRA_VALUES[@]}"
```
gionn: you probably want this before your local .github override so there is still a chance to override something

.github/actions/deploy-local-acs/action.yml
```diff
@@ -116,14 +120,49 @@ runs:
...
+          if helm dep build acs-deployment/helm/alfresco-content-services; then
+            break
+          fi
+          if [[ $attempt -eq $max_retries ]]; then
+            echo "helm dep build failed after ${max_retries} attempts"
+            exit 1
+          fi
+          echo "Retrying in $((attempt * 10))s..."
+          sleep $((attempt * 10))
+        done
```
gionn: not really sure it's worth having it, when github cdn is having issue there is an high chance that the build would fail later as well on any other interaction

.github/actions/deploy-local-acs/action.yml
```diff
@@ -18,6 +18,10 @@ inputs:
     description: 'ACS deployment version (defaults to latest)'
     required: false
     default: 'latest'
+  latest_images:
```
gionn: alpha_images or pre_release_images is probably even more intuitive

e2e/playwright/smoke-test/src/tests/viewer/viewer.e2e.ts
```diff
@@ -76,7 +75,7 @@ test.describe('viewer file', () => {
       }
     }
 
-    docLibId = await siteActionsAdmin.getDocLibId(siteAdmin);
+    await siteActionsAdmin.getDocLibId(siteAdmin);
```
AleksanderSklorz: Is calling that function still required? It looks strange - here is id loading but that id is never used

projects/aca-content/src/lib/store/effects/download.effects.spec.ts
```diff
@@ -112,4 +113,43 @@ describe('DownloadEffects', () => {
...
+    it('should use the version name (not the current node name) as the download file name', () => {
+      spyOn(contentUrlService, 'getVersionContentUrl').and.returnValue(of('http://external-host.example.com/versions/1.0/content?attachment=true'));
+
+      const anchor = document.createElement('a');
+      spyOn(anchor, 'click');
+      spyOn(document, 'createElement').and.returnValue(anchor);
+      spyOn(document.body, 'appendChild').and.callThrough();
+      spyOn(document.body, 'removeChild').and.callThrough();
+
+      spyOn(store, 'select').and.returnValues(
```
AleksanderSklorz: For normal code using spyOn and returnValues is correct approach but for Store from NgRx there are other solutions for mocking data which are better when working with store, see: 
https://ngrx.io/guide/store/testing#:~:text=)%3B%0A%20%20%7D)%3B%0A%7D)%3B-,Using%20Mock%20Selectors,-MockStore%20also%20provides

projects/aca-content/src/lib/store/effects/download.effects.ts
```diff
@@ -143,16 +143,44 @@ export class DownloadEffects {
...
+    } else {
+      this.triggerDownload(url, fileName);
+    }
+  }
 
-      link.style.display = 'none';
-      link.download = fileName;
-      link.href = url;
+  private downloadThroughBlob(url: string, fileName: string) {
+    fetch(url)
```
AleksanderSklorz: Could you please check if instead of fetch you can use HttpClient? 

Using HttpClient is more preferable in angular because HttpClient is Angular class to work with Http calls. Also fetch returns Promise, when HttpClient's functions returns Observable (which is also more preferable for Angular). 

You can inject HttpClient class using following: 
private readonly httpClient = inject(HttpClient);

projects/aca-content/src/lib/store/effects/download.effects.ts
```diff
@@ -143,16 +143,44 @@ export class DownloadEffects {
...
+      .then((response) => response.blob())
+      .then((blob) => {
+        const blobUrl = window.URL.createObjectURL(blob);
+        this.triggerDownload(blobUrl, fileName);
+        setTimeout(() => window.URL.revokeObjectURL(blobUrl), 0);
+      })
+      .catch(() => this.triggerDownload(url, fileName));
+  }
+
+  private triggerDownload(url: string, fileName: string) {
```
AleksanderSklorz: Can you record video with that case to explain why that whole approach with creating element manually is required? Basically creating elements using for example createElement are very rare situations because most html should be placed directly in component in html files instead of creating it manually like that so I'm wondering what case requires that.

package.json
```diff
@@ -98,7 +98,7 @@
     "@nx/angular": "22.7.6",
     "@nx/eslint-plugin": "22.5.4",
     "@nx/workspace": "22.5.4",
-    "@playwright/test": "^1.56.1",
+    "@playwright/test": "1.61.1",
```
AleksanderSklorz: earlier we had ^ before version so it's probably missing

package-lock.json
AleksanderSklorz, line 1: Do you have correct node and npm versions? When I run npm install it builds different package-lock.json file for me. Please check your versions and update file if needed

adamswiderskipw: Yeah, I wanted to update the playwright version manually but forgot about the ^

e2e/playwright/viewer/src/tests/viewer-version-actions.e2e.ts
```diff
@@ -54,14 +54,27 @@ test.describe('Version actions', () => {
...
+
+  async function viewFirstFileVersion(page: PageWithViewer) {
     await page.dataTable.selectItems(filenameAfterUpdate);
     await page.acaHeader.clickMoreActions();
     await page.matMenu.clickMenuItem('Manage Versions');
     await page.manageVersionsDialog.clickListActionButtonForVersion('1.0');
     await page.matMenu.clickMenuItem('View');
   }
 
+  async function waitForViewerDocumentToRender(page: PageWithViewer) {
```
akashrathod28: please add return type like -  `Promise<void>`

e2e/playwright/viewer/src/tests/viewer-version-actions.e2e.ts
```diff
@@ -54,14 +54,27 @@ test.describe('Version actions', () => {
...
+    await page.viewer.waitForViewerToOpen();
+    await page.viewer.waitForViewerContentToRender('document');
+  }
+
+  async function downloadFromViewer(page: PageWithViewer) {
```
akashrathod28: please add return type like - `Promise<Download>`

package-lock.json
AleksanderSklorz: @adamswiderskipw that package-lock.json looks still incorrectly. When I run npm install I can see a lot of differences: 
image

Please ensure you use required node and npm version and reinstall dependencies to have updated package-lock.json

AleksanderSklorz: See scrollbars how many differences there are highlighted

adamswiderskipw: Updated node and npm, try now :)

AleksanderSklorz (approved):

PR #5263, opened 2026-07-01
ACS-12055 Auto-detect latest ACS image tag from Quay.io in local deploy

akashrathod28:
https://hyland.atlassian.net/browse/ACS-12055


gionn:
we agreed to just pull and use https://github.com/Alfresco/acs-deployment/blob/master/helm/alfresco-content-services/pre-release_values.yaml instead of the default when you want to test against pre-releases

PR #5272, opened 2026-07-06
[ACS-11909] [ADW] Link to file or folder plan for Type filter and Delete operation

dominikiwanekhyland:
…

https://hyland.atlassian.net/browse/ACS-11909

### DIFF
--- a/app/src/app.config.json
+++ b/app/src/app.config.json
@@ -185,6 +185,10 @@
               {
                 "name": "SEARCH.SEARCH_HEADER.FILTERS.TYPE.DOCUMENT",
                 "value": "TYPE:'cm:content'"
+              },
+              {
+                "name": "SEARCH.SEARCH_HEADER.FILTERS.TYPE.LINK",
+                "value": "TYPE:'cm:link'"
               }
             ]
           }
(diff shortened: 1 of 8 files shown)


AleksanderSklorz (commented):
@dominikiwanekhyland when you have selected Link filter and then you create Link then you can't see that Link until filters are changed. It should be visible after creation as you have selected Link filter

https://github.com/user-attachments/assets/94b89267-5543-46e8-bed4-c835eb07dcdf

AleksanderSklorz:
@dominikiwanekhyland when I try to delete Link but then I cancel it then I have 2 issues
- there is infinite spinner which never finishes 
- why spinner is displayed at all? It has sense to display it when removing is confirmed but if we cancel it then there is nothing changed so we slow down user because of that as user needs to wait until data is loaded again/spinner is hidden but there is nothing changed really. We should not display spinner and data should not be reloaded when we cancel removing 


https://github.com/user-attachments/assets/e4bb4959-7404-4cf1-a3d5-3022c78df478

dominikiwanekhyland:
@AleksanderSklorz thanks, I think those issues are fixed now

AleksanderSklorz (approved):

PR #5290, opened 2026-07-15
[ACS-8468] Add copy-to-clipboard-action to content-metadata

Shivangi917:
https://hyland.atlassian.net/browse/ACS-8468

### DIFF
--- a/app/src/app.config.json
+++ b/app/src/app.config.json
@@ -125,7 +125,8 @@
     },
     "multi-value-pipe-separator": ", ",
     "multi-value-chips": true,
-    "copy-to-clipboard-action": true
+    "copy-to-clipboard-action": true,
+    "display-copy-to-clipboard-icon": false
   },
   "search-headers": {
     "filterWithContains": true,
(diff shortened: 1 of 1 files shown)


app/src/app.config.json
```diff
@@ -124,7 +124,8 @@
       ]
     },
     "multi-value-pipe-separator": ", ",
-    "multi-value-chips": true
+    "multi-value-chips": true,
+    "copy-to-clipboard-action": true
```
MichalKinas: So are we setting `displayCopyToClipboardIcon` to `true` somewhere?

Shivangi917: No, adf has this line `this.copyToClipboardAction = this.appConfig.get<boolean>('content-metadata.copy-to-clipboard-action');` for double click functionality that is why we needed this line in aca. For copy-icon we don't have this line we can change it to true or false in this file `alfresco-ng2-components/lib/core/src/lib/card-view/components/card-view/card-view.component.ts`. Will it be okay to add an entry to aca like there is for double click to copy functionality

MichalKinas: Yes please do that

Shivangi917: done

AleksanderSklorz (approved):

MichalKinas (approved):

PR #5291, opened 2026-07-16
[ACS-9106] Remove double checkbox

Shivangi917:
https://hyland.atlassian.net/browse/ACS-9106


projects/aca-content/src/lib/components/dl-custom-components/thumbnail-column/thumbnail-column.component.spec.ts
```diff
@@ -109,4 +109,11 @@ describe('ThumbnailColumnComponent', () => {
     const tooltip = component.getToolTip(context);
     expect(tooltip).toBe('');
   });
+
+  it('should return isSelected false when multiselect is false and row is selected', () => {
+    component.multiselect = false;
+    component.context = { row: { isSelected: true }, data: { getValue: () => '' }, col: {} };
+
+    expect(component.isSelected).toBeTrue();
+  });
```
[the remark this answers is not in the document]
Shivangi917: Yes, correct. Fixed that in my next commit

projects/aca-content/src/lib/components/dl-custom-components/thumbnail-column/thumbnail-column.component.ts
```diff
@@ -40,12 +40,15 @@ export class ThumbnailColumnComponent implements OnChanges {
   @Input()
   context: any;
 
+  @Input()
+  multiselect = false;
```
MichalKinas: This logic seems to be a bit off, in this case when multiselect is enabled row can't be selected, I think we should refine the param naming and treat this param as a selection override that should be connected with datatable multi selection value, please create a variable for it in serach results component, provide the same value for multi selection in data table and for this param and correct the naming of the param so that in the future it would be more explanatory

Shivangi917: i have changed its name and added it to search component too

projects/aca-content/src/lib/components/dl-custom-components/thumbnail-column/thumbnail-column.component.ts
AleksanderSklorz, line 1: I'm wondering - do we need that component now? It used only by search page and now we don't display that selected icon at all. This component is also used to display rest icons like node type on search page: 

Image

and after removing that component they are gone: 

Image

but if you check other page like personal files - those icons are visible normally, without that extra component: 

Image

Shivangi - please check if we can remove that thumbnail-column.component.ts at all and display icons normally like on personal files page without additional component.

Shivangi917: ok, i need to check that

projects/aca-content/src/lib/components/search/search-results/search-results.component.html
```diff
@@ -111,7 +111,10 @@
                 [sortable]="false"
               >
                 <ng-template let-context>
-                  <aca-custom-thumbnail-column [context]="context" />
+                  <mat-icon *ngIf="isMaterialIcon(context); else thumbImage">{{ getMatIconName(context) }}</mat-icon>
+                  <ng-template #thumbImage>
+                    <img [src]="getIconSource(context)" [alt]="getTooltip(context)" [title]="getTooltip(context)" />
```
AleksanderSklorz: It's not exactly what I meant in that comment: https://github.com/Alfresco/alfresco-content-app/pull/5291#discussion_r3602929151

You just moved that code from thumbnail-column.component.ts to this file and removed thumbnail component but it's still the same approach. If you read my comment again - I wrote there that Personal Files are working without that and it should be investigated if we really need that code or we can do it like in Personal Files and display those icons without that extra code. Please investigate how Personal Files are implemented and if that code is really needed or we can display those icons in other way. 

If we still need that code then removing that thumbnail component has no sense as we can just use that component

Shivangi917: okay

projects/aca-content/src/lib/components/search/search-results/search-results.component.html
```diff
@@ -24,6 +24,7 @@
             <div class="aca-content__advanced-filters">
               <div class="aca-content__advanced-filters--header">
                 <p>{{ 'APP.BROWSE.SEARCH.ADVANCED_FILTERS' | translate }}</p>
+              <aca-search-action-menu (sortingSelected)="onSearchSortingUpdate($event)" />
```
AleksanderSklorz: Why did you move that line to this place instead of keeping it in previous place? It breaks layout now, earlier 3 dots were next to header with checkbox: 

Image

but now location is next to Advance filters label 

Image

Shivangi917: i'm sorry while removing the component I didn't notice that

AleksanderSklorz (approved):

dominikiwanekhyland (approved):

MichalKinas (commented):
The UI looks broken after this change:

Image

projects/aca-content/src/lib/components/search/search-results/search-results.component.html
```diff
@@ -104,23 +104,19 @@
           >
             <data-columns>
               <data-column
+                class="adf-search-results__thumbnail-sort"
```
MichalKinas: Is the class used anywhere? Do we need it?

Shivangi917: No, initially it was to do the styling and I changed in scss file using this class in my previous commit but now we don't need it I'll remove it

MichalKinas (approved):

PR #5292, opened 2026-07-16
[ACS-12246] Sorting for some columns in Shared Files page are not working (Location, Size, Modified by, Shared by)

dominikiwanekhyland:
https://hyland.atlassian.net/browse/ACS-12246

### DIFF
--- a/projects/aca-content/assets/app.extensions.json
+++ b/projects/aca-content/assets/app.extensions.json
@@ -2646,7 +2646,6 @@
         {
           "id": "app.libraries.name",
           "key": "title",
-          "sortingKey": "name",
           "title": "APP.DOCUMENT_LIST.COLUMNS.NAME",
           "type": "text",
           "class": "adf-ellipsis-cell adf-expand-cell-3",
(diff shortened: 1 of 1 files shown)


Shivangi917 (approved):

projects/aca-content/assets/app.extensions.json
```diff
@@ -2646,7 +2646,6 @@
         {
           "id": "app.libraries.name",
           "key": "title",
-          "sortingKey": "name",
```
MichalKinas, line 2649: Why are we removing sortingKeys? Was there any other change related to that?

dominikiwanekhyland: @MichalKinas 

In november there were changes in ADF datatable and I think this line caused this issue 
       if ((column.sortingKey || column.key) === current?.key) {
                newDirection = current.direction?.toLowerCase() === 'asc' ? 'desc' : 'asc';
            }



For a column where sortingKey differs from key, that toggle check never matches

dominikiwanekhyland: @MichalKinas 

And that Shared Files is client sorted, so it doesn't need sortingKey, as it uses column.key for sorting

MichalKinas: Makes sense, thanks for the explanation, approved

MichalKinas (approved):

AleksanderSklorz (approved):

PR #5294, opened 2026-07-17
[ACS-9583] Strip details segment from url

Shivangi917:
https://hyland.atlassian.net/browse/ACS-9583

### DIFF
--- a/projects/aca-content/src/lib/store/effects/viewer.effects.ts
+++ b/projects/aca-content/src/lib/store/effects/viewer.effects.ts
@@ -217,7 +217,7 @@ export class ViewerEffects {
 
     const detailsIndex = urlSegments.findIndex((segment) => segment.path === 'details');
     if (detailsIndex > -1) {
-      urlSegments.splice(detailsIndex, 2);
+      urlSegments.splice(detailsIndex);
     }
 
     return urlSegments.reduce(function (acc, item) {
--- a/projects/aca-content/src/lib/store/effects/viewer.effects.ts
+++ b/projects/aca-content/src/lib/store/effects/viewer.effects.ts
@@ -215,6 +215,11 @@ export class ViewerEffects {
 
     const urlSegments: UrlSegment[] = urlSegmentGroup.segments;
 
+    const detailsIndex = urlSegments.findIndex((segment) => segment.path === 'details');
+    if (detailsIndex > -1) {
+      urlSegments.splice(detailsIndex, 2);
+    }
+
     return urlSegments.reduce(function (acc, item) {
       acc.push(item.path, item.parameters);
       return acc;
--- a/projects/aca-content/src/lib/store/effects/viewer.effects.spec.ts
+++ b/projects/aca-content/src/lib/store/effects/viewer.effects.spec.ts
@@ -89,5 +89,14 @@ describe('ViewerEffects', () => {
 
       expect(router.navigateByUrl['calls'].argsFor(0)[0].toString()).toEqual('/view/(viewer:nodeId)?path=absolute-path');
     }));
+
+    it('should navigate to the parent route and preserve the original location when opening the viewer', fakeAsync(() => {
+      store.dispatch(new ViewNodeAction('nodeId', { location: 'personal-files/details/abc123/permissions' }));
+      tick(100);
+
+      expect(router.navigateByUrl['calls'].argsFor(0)[0].toString()).toEqual(
+        '/personal-files/permissions/(viewer:view/nodeId)?location=personal-files%2Fdetails%2Fabc123%2Fpermissions'
+      );
+    }));
   });
 });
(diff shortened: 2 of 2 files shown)


projects/aca-content/src/lib/components/toolbar/view-node/view-node.component.ts
```diff
@@ -91,7 +91,9 @@ export class ViewNodeComponent {
 
   private navigateToViewer(id: string): void {
     this.activatedRoute.queryParams.pipe(take(1)).subscribe((params) => {
-      const location = params.location || this.router.url;
+      const detailsIndex = this.router.url.indexOf('/details/');
+      const baseUrl = detailsIndex > -1 ? this.router.url.substring(0, detailsIndex) : this.router.url;
+      const location = params.location || baseUrl;
```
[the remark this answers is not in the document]
Shivangi917: the old code was passing full url which caused navigation error. even if we keep query params urlSegments reads path segments not query params

projects/aca-content/src/lib/components/toolbar/view-node/view-node.component.ts
```diff
@@ -91,7 +91,9 @@ export class ViewNodeComponent {
 
   private navigateToViewer(id: string): void {
     this.activatedRoute.queryParams.pipe(take(1)).subscribe((params) => {
-      const location = params.location || this.router.url;
+      const detailsIndex = this.router.url.indexOf('/details/');
+      const baseUrl = detailsIndex > -1 ? this.router.url.substring(0, detailsIndex) : this.router.url;
```
AleksanderSklorz: This solution doesn't seem correct. By modifying url it breaks navigation to previous location when you click X icon to close viewer, see below video. If I'm on permissions page and click X for viewer then I go to personal files page instead of permissions page but when I'm on library page and click X for viewer then I go to libraries page. 

https://github.com/user-attachments/assets/da421e9a-2112-4c29-81a6-8eb3e2c58e98

AleksanderSklorz:
@Shivangi917 strange issue - when I open viewer in other place then metadata sidebar at right side is not opened automatically but when I open viewer through permissions page then it's opened automatically - is that expected? Looks like some inconsistency issue 

https://github.com/user-attachments/assets/5615c77d-29c6-477a-a2c0-4dccd94fbc8b

projects/aca-content/src/lib/components/toolbar/view-node/view-node.component.ts
AleksanderSklorz: Next reason why this solution seems not fully correct is that when you try to open viewer on permissions page in other way than click on eye icon, for example through Manage Version dialog, then again there is that old issue. Looks like fix should be in some common place for both ways of opening viewer and probably instead of modifying location url we should somehow properly handle that location 

https://github.com/user-attachments/assets/89ce8a93-7a26-4201-b210-75db80fe27c4

projects/aca-content/src/lib/components/toolbar/view-node/view-node.component.spec.ts
```diff
@@ -138,4 +138,29 @@ describe('ViewNodeComponent', () => {
     component.onClick();
     expect(mockStore.dispatch).toHaveBeenCalledWith(new ViewNodeAction(id, { location: 'other-location/1234' }));
   });
+
+  it('should strip the details segment from the current URL when location query param is absent', () => {
+    const linkNode = {
```
AleksanderSklorz: missing type

projects/aca-content/src/lib/components/toolbar/view-node/view-node.component.spec.ts
```diff
@@ -138,4 +138,29 @@ describe('ViewNodeComponent', () => {
...
+    component.data = {
+      iconButton: true
+    };
+    mockStore.dispatch.calls.reset();
+    route.queryParams = of({});
+    mockStore.select.and.returnValue(of(linkNode));
+    mockRouter.url = 'some-url/details/nodeId/permissions';
+    fixture.detectChanges();
+
+    component.onClick();
```
AleksanderSklorz: Let's keep correct flow - instead of calling onClick function you should start from beginning of flow which is clicking on button where onClick is assigned as it's start point of that flow

projects/aca-content/src/lib/store/effects/viewer.effects.ts
```diff
@@ -215,6 +215,11 @@ export class ViewerEffects {
 
     const urlSegments: UrlSegment[] = urlSegmentGroup.segments;
 
+    const detailsIndex = urlSegments.findIndex((segment) => segment.path === 'details');
+    if (detailsIndex > -1) {
+      urlSegments.splice(detailsIndex, 2);
```
AleksanderSklorz: You assume there are 2 left segments. It may be correct now but in future we may have more parts and it will break. Maybe you can skip second parameter (2) and just leave: 
urlSegments.splice(detailsIndex);

then it will remove all rest segments instead of only 2 so if number of segments will change it will be still working

AleksanderSklorz:
Sonar is failed/cancelled, please restart 

image


image

projects/aca-content/src/lib/store/effects/viewer.effects.ts
Shivangi917: done

AleksanderSklorz (approved):

MichalKinas (approved):

PR #5306, opened 2026-07-23
[ACS-12311] Show message for non versionable file

Shivangi917:
https://hyland.atlassian.net/browse/ACS-12311

### DIFF
--- a/projects/aca-content/src/lib/services/content-management.service.ts
+++ b/projects/aca-content/src/lib/services/content-management.service.ts
@@ -642,9 +642,18 @@ export class ContentManagementService {
     );
   }
 
+  private isNodeVersionable(node: any): boolean {
+    return (node?.aspectNames ?? []).includes('cm:versionable');
+  }
+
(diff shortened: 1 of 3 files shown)


projects/aca-content/src/lib/services/content-management.service.ts
```diff
@@ -642,9 +642,18 @@ export class ContentManagementService {
...
   }
 
+  private isNodeVersionable(node: any): boolean {
+    return (node?.aspectNames ?? []).includes('cm:versionable');
+  }
+
   private openVersionManagerDialog(node: any, focusedElementOnCloseSelector?: string) {
     // workaround Shared
     if (node.isFile || node.nodeId) {
+      if (!this.isNodeVersionable(node)) {
```
MichalKinas, line 652: That's not the usual approach we take in ACA, instead of showing the action that won't work for some nodes we need to modify the visibility of a rule for this action, please add necessary check for `MANAGE_VERSIONS` rules, also make sure that aspects are fetched search results page

MichalKinas (changes requested):

PR #5307, opened 2026-07-23
[ACS-12311] Remove manage versions dialog box for non versionable file

Shivangi917:
https://hyland.atlassian.net/browse/ACS-12311


projects/aca-shared/rules/src/app.rules.ts
```diff
@@ -529,6 +529,20 @@ export function createVersionRule(minimalVersion: string): (context: RuleContext
...
+ * JSON ref: `app.selection.file.isVersionable`
+ *
+ * @param context Rule execution context
+ */
+export const isVersionableFile = (context: RuleContext): boolean => {
+  const file = context?.selection?.file;
+  if (!file?.entry) {
+    return false;
+  }
+  return (file.entry.aspectNames ?? []).includes('cm:versionable');
```
AleksanderSklorz, line 543: To ensure - is that solution correct because when I compared it with remote env then it doesn't work now for cases when it was working previously, for example see video with comparison 

https://github.com/user-attachments/assets/480e5e70-2bb1-4869-aac6-60c687e1f9e1

AleksanderSklorz: cc: @MichalKinas ?

projects/aca-shared/rules/src/app.rules.ts
```diff
@@ -529,6 +529,20 @@ export function createVersionRule(minimalVersion: string): (context: RuleContext
   };
 }
 
+/**
+ * Checks if the selected file has the `cm:versionable` aspect.
+ * JSON ref: `app.selection.file.isVersionable`
+ *
+ * @param context Rule execution context
+ */
+export const isVersionableFile = (context: RuleContext): boolean => {
```
AleksanderSklorz, line 538: we have documentation for all rules in rules-list.md file, please add this rule there as well

projects/aca-shared/rules/src/app.rules.spec.ts
```diff
@@ -250,6 +250,38 @@ describe('app.evaluators', () => {
     });
   });
 
+  describe('isVersionableFile', () => {
+    it('should return [true] when the selected file has the cm:versionable aspect', () => {
```
AleksanderSklorz: Please replace all [true] or [false] from your tests titles with just true or false because it's confusing because it looks like array of true or false

projects/aca-shared/rules/src/app.rules.spec.ts
```diff
@@ -250,6 +250,38 @@ describe('app.evaluators', () => {
...
+      context.selection.file = { entry: { aspectNames: ['cm:versionable'] } } as any;
```
AleksanderSklorz: Instead of as any please use as NodeEntry. The same for rest your tests

projects/aca-shared/rules/src/app.rules.ts
Shivangi917: It should work now @AleksanderSklorz

MichalKinas (approved):

AleksanderSklorz (approved):

akashrathod28 (approved):

PR #5332, opened 2026-08-05
[ADW] Search page displays a loading bar when navigating back from the File Viewer

dominikiwanekhyland:
https://hyland.atlassian.net/browse/ACS-12389

### DIFF
--- a/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
+++ b/projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
@@ -390,7 +390,6 @@ describe('SearchComponent', () => {
   it('should NOT render the loading bar when navigating back with an unchanged query', async () => {
     spyOn(queryBuilder, 'execute');
     queryParams.next({ q: encodeQuery({ userQuery: 'cm:name:"test*"' }) });
-    tick();
 
     component.isLoading = false;
     fixture.detectChanges();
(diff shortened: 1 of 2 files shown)


projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
```diff
@@ -385,6 +385,40 @@ describe('SearchComponent', () => {
     expect(queryBuilder.execute).not.toHaveBeenCalled();
   }));
 
+  it('should NOT show the loading bar when navigating back with an unchanged query', fakeAsync(() => {
```
AleksanderSklorz: Can we test if mat-progress-bar is/isn't rendered in DOM instead of test isLoading flag? Purpose of that test is checking if loading bar is showed but test doesn't check it. Test just check state of isLoading flag which doesn't prove loading bar existence (for example it can have missing if condition in html and then test will pass but real case will be broken)

projects/aca-content/src/lib/components/search/search-results/search-results.component.spec.ts
```diff
@@ -385,6 +385,40 @@ describe('SearchComponent', () => {
...
+    queryParams.next({ q: encodeQuery({ userQuery: 'cm:name:"test*"' }) });
+
+    tick();
+
+    expect(component.isLoading).toBeFalse();
+
+    flush();
+  }));
+
+  it('should show the loading bar when navigating with a changed query', fakeAsync(() => {
```
AleksanderSklorz: Same like https://github.com/Alfresco/alfresco-content-app/pull/5332/changes#r3720930858

AleksanderSklorz (approved):

PR #5333, opened 2026-08-06
[ACS-12391] Clicking the reduce panel displays reduced details view

Shivangi917:
https://hyland.atlassian.net/browse/ACS-12391

This PR fixes the issue of clicking on reduce panel and seeing search bar instead of reduced details page.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -38,6 +38,5 @@
   },
   "cacheDirectory": "nxcache",
   "useInferencePlugins": false,
-  "useLegacyCache": true,
-  "analytics": false
+  "useLegacyCache": true
 }
--- a/nx.json
+++ b/nx.json
@@ -38,5 +38,6 @@
   },
   "cacheDirectory": "nxcache",
   "useInferencePlugins": false,
-  "useLegacyCache": true
+  "useLegacyCache": true,
+  "analytics": false
 }
--- a/projects/aca-content/src/lib/components/details/details.component.spec.ts
+++ b/projects/aca-content/src/lib/components/details/details.component.spec.ts
@@ -253,8 +253,8 @@ describe('DetailsComponent', () => {
   });
 
   describe('Reduce Panel', () => {
-    const getReducePanelButton = () => testingUtils.getByDataAutomationId('close-library').nativeElement;
-    const clickReducePanelButton = () => getReducePanelButton().click();
+    const getReducePanelButton = (): HTMLButtonElement => testingUtils.getByDataAutomationId('close-library').nativeElement;
+    const clickReducePanelButton = (): void => getReducePanelButton().click();
 
     it('should navigate to the file list page, keep the node selected and the info drawer open', () => {
       queryParamsSubject.next({ location: '/personal-files' });
--- a/projects/aca-content/src/lib/components/details/details.component.spec.ts
+++ b/projects/aca-content/src/lib/components/details/details.component.spec.ts
@@ -25,11 +25,11 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { AppTestingModule } from '../../testing/app-testing.module';
 import { DetailsComponent } from './details.component';
-import { ActivatedRoute } from '@angular/router';
+import { ActivatedRoute, NavigationEnd, Router } from '@angular/router';
 import { BehaviorSubject, of, Subject } from 'rxjs';
 import { Store } from '@ngrx/store';
 import { AppExtensionService, AppHookService, ContentApiService } from '@alfresco/aca-shared';
-import { NavigateToFolder, SetSelectedNodesAction } from '@alfresco/aca-shared/store';
+import { NavigateToFolder, NavigateToPreviousPage, SetInfoDrawerStateAction, SetSelectedNodesAction } from '@alfresco/aca-shared/store';
 import { Node, NodeEntry, PathElement } from '@alfresco/js-api';
 import { BreadcrumbComponent, ContentService, NodesApiService, SearchQueryBuilderService } from '@alfresco/adf-content-services';
 import { By } from '@angular/platform-browser';
@@ -45,7 +45,9 @@ describe('DetailsComponent', () => {
   let appHookService: AppHookService;
   let location: Location;
   let store: Store;
+  let router: Router;
   let node: NodeEntry;
+  let queryParamsSubject: BehaviorSubject<any>;
 
   const mockStream = new Subject();
   const storeMock = {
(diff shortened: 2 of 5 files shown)


projects/aca-content/src/lib/components/details/details.component.ts
```diff
@@ -124,8 +134,20 @@ export class DetailsComponent extends PageComponent implements OnInit, OnDestroy
     }
   }
 
+  saveRoute(route: string): void {
+    this._previousRoute = route;
+  }
```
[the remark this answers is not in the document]
Shivangi917: the value is saved only when location exists or not

projects/aca-content/src/lib/components/details/details.component.spec.ts
```diff
@@ -45,7 +45,9 @@ describe('DetailsComponent', () => {
   let appHookService: AppHookService;
   let location: Location;
   let store: Store;
+  let router: Router;
   let node: NodeEntry;
+  let queryParamsSubject: BehaviorSubject<any>;
```
AleksanderSklorz: If you check definition of ActivatedRoute, expected type is Params instead of any so please replace any type with Params: 

Image

projects/aca-content/src/lib/components/details/details.component.spec.ts
```diff
@@ -243,4 +248,35 @@ describe('DetailsComponent', () => {
 
     expect(locationSpy).toHaveBeenCalled();
   });
+
+  describe('Reduce Panel', () => {
+    it('should save the previous route from the location query param', () => {
+      queryParamsSubject.next({ location: '/personal-files' });
+      fixture.detectChanges();
+
+      expect(component.previousRoute).toBe('/personal-files');
```
AleksanderSklorz: As I mentioned in some other PR - please test final result instead of intermediate state - setting of previousRoute is not final because previousRoute is later used by 
```ts
if (this.previousRoute) {
      this.router.navigateByUrl(this.previousRoute);
    } else {
      this.store.dispatch(new NavigateToPreviousPage());
    }

```
And it's final result which you should test instead

projects/aca-content/src/lib/components/details/details.component.ts
```diff
@@ -77,6 +78,10 @@ export class DetailsComponent extends PageComponent implements OnInit, OnDestroy
   nodeIcon: string;
   canManagePermissions = true;
 
+  get previousRoute(): string {
```
AleksanderSklorz: Why getter? It's not used outside that component. You can keep it just private. If you make it as getter because of unit tests then as I mentioned in earlier comment - you should not test that property because it's intermediate state instead of final result. You should test code which I mentioned in other comment so then you don't need to access that property in tests

projects/aca-content/src/lib/components/details/details.component.ts
```diff
@@ -69,6 +69,7 @@ export class DetailsComponent extends PageComponent implements OnInit, OnDestroy
   private readonly nodesApiService = inject(NodesApiService);
   private readonly appHookService = inject(AppHookService);
   private readonly location = inject(Location);
+  private _previousRoute = '';
```
AleksanderSklorz: please keep private readonly and private separately, they should be grouped separately - private readonly one group and just private other group

projects/aca-content/src/lib/components/details/details.component.spec.ts
```diff
@@ -243,4 +248,35 @@ describe('DetailsComponent', () => {
...
+    });
+
+    it('should navigate to the previous route and re-open the info drawer', () => {
+      queryParamsSubject.next({ location: '/personal-files' });
+      fixture.detectChanges();
+      const navigateSpy = spyOn(router, 'navigateByUrl').and.stub();
+      Object.defineProperty(router, 'events', { value: of(new NavigationEnd(1, '', '')) });
+
+      component.goBack();
```
AleksanderSklorz: Typically user doesn't call goBack function and instead user clicks button so could you also click button (to keep the same user's flow) instead of calling function directly?

projects/aca-content/src/lib/components/details/details.component.spec.ts
```diff
@@ -243,4 +248,35 @@ describe('DetailsComponent', () => {
...
+      expect(navigateSpy).toHaveBeenCalledWith('/personal-files');
+      expect(store.dispatch).toHaveBeenCalledWith(jasmine.any(SetInfoDrawerStateAction));
+      expect(store.dispatch).toHaveBeenCalledWith(jasmine.objectContaining({ payload: true }));
+    });
+
+    it('should dispatch NavigateToPreviousPage when there is no previous route', () => {
+      fixture.detectChanges();
+      Object.defineProperty(router, 'events', { value: of(new NavigationEnd(1, '', '')) });
+
+      component.goBack();
```
AleksanderSklorz: Typically user doesn't call goBack function and instead user clicks button so could you also click button (to keep the same user's flow) instead of calling function directly?

AleksanderSklorz:
@Shivangi917 could you please check with @akashrathod28 or @MichalKinas that scenario from below video? When I click reduce icon then metadata panel at right side is not open so should not it be opened as we switch to reduced mode? Please note it works fine when you were on Viewer page - then that panel is opened after reducing 

https://github.com/user-attachments/assets/627950a8-1d07-4b0f-a0a6-963cc625c17b

projects/aca-content/src/lib/components/details/details.component.ts
AleksanderSklorz: If you remove getter as suggested in https://github.com/Alfresco/alfresco-content-app/pull/5333/changes#r3728232516 then you should remove _ prefix

projects/aca-content/src/lib/components/details/details.component.ts
Shivangi917: I've removed it. While working on this issue, I was referring to another file and noticed the same thing being used there. And in confusion, I defined it as getter

projects/aca-content/src/lib/components/details/details.component.ts
Shivangi917: done

nx.json
```diff
@@ -38,5 +38,6 @@
   },
   "cacheDirectory": "nxcache",
   "useInferencePlugins": false,
-  "useLegacyCache": true
+  "useLegacyCache": true,
+  "analytics": false
```
AleksanderSklorz: unexpected change, please revert it

projects/aca-content/src/lib/components/details/details.component.spec.ts
```diff
@@ -243,4 +251,46 @@ describe('DetailsComponent', () => {
 
     expect(locationSpy).toHaveBeenCalled();
   });
+
+  describe('Reduce Panel', () => {
+    const getReducePanelButton = () => testingUtils.getByDataAutomationId('close-library').nativeElement;
```
AleksanderSklorz: Could you please add return type to that function? Actually it's missing return type. In case when function returns nothing like below you can add void as return type

nx.json
Shivangi917: yes, actually it gets added automatically every time i do `npm run start -- -c adf` I forgot to remove it this time

AleksanderSklorz (approved):

dominikiwanekhyland (approved):

PR #5343, opened 2026-08-10
[ACS-12460] Fix sonarcloud issues

Shivangi917:
https://hyland.atlassian.net/browse/ACS-12460

This PR fixes the vulnerabilities of SonarCloud.

### DIFF
--- a/.github/actions/adf-linking/action.yml
+++ b/.github/actions/adf-linking/action.yml
@@ -21,7 +21,7 @@ runs:
           cd ../alfresco-ng2-components
           if [ -f "pnpm-lock.yaml" ]; then
             corepack enable pnpm
-            pnpm add @angular/material
+            pnpm add @angular/material --ignore-scripts
           else
             npm install @angular/material
           fi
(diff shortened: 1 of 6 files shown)


g-jaskowski (approved):

.github/actions/adf-linking/action.yml
```diff
@@ -21,7 +21,7 @@ runs:
           cd ../alfresco-ng2-components
           if [ -f "pnpm-lock.yaml" ]; then
             corepack enable pnpm
-            pnpm add @angular/material
+            pnpm add @angular/material --ignore-scripts
```
MichalKinas, line 24: I think we may want to skip those changes after pnpm migration is done right?

Shivangi917: yes, we won't need this PR if we merge migration PR

AleksanderSklorz:
@Shivangi917 please check copilot's comments

Shivangi917:
@AleksanderSklorz copilot's suggestion seems to only block `husky` installations but sonarcloud issues indicate this`Omitting "--ignore-scripts" allows lifecycle scripts to run during package installation.` and if we migrate to pnpm I don't think we would need this PR to be merged

PR #5360, opened 2026-08-31
[ACS-12333][ACS-12335][ACS-12334] Add Checkout / Check-in Support

nikita-web-ua:
https://hyland.atlassian.net/browse/ACS-12331

### DIFF
--- a/projects/aca-content/assets/app.extensions.json
+++ b/projects/aca-content/assets/app.extensions.json
@@ -537,8 +537,7 @@
               "visible": [
                 "app.selection.file",
                 "!app.navigation.isTrashcan",
-                "canToggleFileLock",
-                "!app.selection.isCheckedOut"
+                "canToggleFileLock"
               ]
             }
           },
@@ -570,7 +569,8 @@
                 "app.selection.notEmpty",
                 "!app.navigation.isTrashcan",
                 "!app.navigation.isLibraries",
-                "!isMultiSelection"
+                "!isMultiSelection",
+                "!app.selection.isWorkingCopy"
               ]
             }
           },
@@ -708,7 +708,9 @@
                 "app.selection.notEmpty",
                 "!app.navigation.isTrashcan",
                 "!app.navigation.isLibraries",
-                "!app.selection.isNodeLink"
+                "!app.selection.isNodeLink",
+                "!app.selection.isLockedOrWorkingCopy"
+
               ]
             }
           },
@@ -725,14 +727,20 @@
                 "!app.navigation.isTrashcan",
                 "!app.navigation.isLibraries",
                 "app.selection.notEmpty",
-                "app.selection.canDelete"
+                "app.selection.canDelete",
+                "!app.selection.isLockedOrWorkingCopy"
               ]
             }
           },
           {
+            "id": "app.create.separator.3",
+            "type": "separator",
+            "order": 900
+          },
+          {
             "id": "app.context.menu.node-info",
             "title": "APP.ACTIONS.NODE_INFO",
-            "order": 800,
+            "order": 950,
             "icon": "info",
             "actions": {
               "click": "NODE_INFORMATION"
@@ -748,11 +756,6 @@
             }
           },
           {
-            "id": "app.create.separator.3",
-            "type": "separator",
-            "order": 900
-          },
-          {
             "id": "app.toolbar.versions",
             "order": 1100,
             "title": "APP.ACTIONS.VERSIONS",
@@ -766,7 +769,8 @@
                 "app.selection.file.isVersionable",
                 "!app.navigation.isTrashcan",
                 "!app.selection.file.isLocked",
-                "!app.selection.isNodeLink"
+                "!app.selection.isNodeLink",
+                "!app.selection.isLockedOrWorkingCopy"
               ]
             }
           },
@@ -853,7 +857,8 @@
             "app.selection.notEmpty",
             "!app.navigation.isTrashcan",
             "!app.navigation.isLibraries",
-            "!isMultiSelection"
+            "!isMultiSelection",
+            "!app.selection.isLockedOrWorkingCopy"
           ]
         }
       },
@@ -900,7 +905,8 @@
         "rules": {
           "visible": [
             "app.selection.file",
-            "!app.navigation.isTrashcan"
+            "!app.navigation.isTrashcan",
+            "!app.selection.isLockedOrWorkingCopy"
           ]
         }
       },
@@ -918,8 +924,7 @@
           "visible": [
             "app.selection.file",
             "!app.navigation.isTrashcan",
-            "canToggleFileLock",
-            "!app.selection.isCheckedOut"
+            "canToggleFileLock"
           ]
         }
       },
@@ -1075,7 +1080,8 @@
             "app.selection.notEmpty",
             "!app.navigation.isTrashcan",
             "!app.navigation.isLibraries",
-            "!app.selection.isNodeLink"
+            "!app.selection.isNodeLink",
+            "!app.selection.isLockedOrWorkingCopy"
           ]
         }
       },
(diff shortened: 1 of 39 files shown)


projects/aca-content/src/lib/components/dl-custom-components/name-column/name-column.component.html
```diff
@@ -15,7 +15,7 @@
       {{ displayText$ | async }}
     </span>
 
-    <ng-container *ngIf="isFile && isFileWriteLocked">
+    <ng-container *ngIf="isFile && isFileWriteLocked || isWorkingCopy">
```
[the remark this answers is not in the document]
nikita-web-ua: done

projects/aca-content/src/lib/components/dl-custom-components/lock-icon/lock-icon.component.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Component, inject, Input, OnInit, ViewEncapsulation } from '@angular/core';
+import { TranslateService } from '@ngx-translate/core';
+import { NodeEntry } from '@alfresco/js-api';
+import { MatIconModule } from '@angular/material/icon';
+
+@Component({
+  selector: 'aca-lock-icon',
+  standalone: true,
+  template: `<mat-icon class="adf-datatable-cell-badge" [title]="tooltip" aria-hidden="true">lock</mat-icon>`,
+  imports: [MatIconModule],
```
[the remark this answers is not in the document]
nikita-web-ua: aria-hidden="true" set similar to other badges

projects/aca-content/assets/app.extensions.json
```diff
@@ -3193,6 +3201,19 @@
           }
         }
       ]
-    }
+    },
+    "badges": [
+      {
+        "id": "app.badge.lock",
+        "icon": "lock",
```
MichalKinas: Also if there is a custom component there icon can be skipped too I think

projects/aca-content/src/lib/components/dl-custom-components/lock-icon/lock-icon.component.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+import { Component, inject, Input, OnInit, ViewEncapsulation } from '@angular/core';
+import { TranslateService } from '@ngx-translate/core';
+import { NodeEntry } from '@alfresco/js-api';
+import { MatIconModule } from '@angular/material/icon';
+
+@Component({
+  selector: 'aca-lock-icon',
+  standalone: true,
```
AleksanderSklorz: Do we still need standalone flag (it's default already I guess)?

projects/aca-content/src/lib/components/dl-custom-components/name-column/name-column.component.spec.ts
```diff
@@ -142,4 +142,26 @@ describe('CustomNameColumnComponent', () => {
...
+          }
+        },
+        getValue: (key: string) => key
+      }
+    };
+
+    component.ngOnInit();
+    fixture.detectChanges();
+
+    expect(fixture.debugElement.nativeElement.querySelector('aca-locked-by')).not.toBe(null);
```
AleksanderSklorz: could you use UnitTestingUtils?

AleksanderSklorz: Instead of string selector, could you please use component's class (LockedByComponent)? It makes easier renaming when component's name changes

projects/aca-content/src/lib/directives/document-list.directive.spec.ts
```diff
@@ -352,4 +352,53 @@ describe('DocumentListDirective', () => {
...
     expect(documentListMock.preselectNodes).toEqual([]);
   }));
+
+  describe('rowFilter for cm:checkedOut nodes', () => {
+    beforeEach(() => {
+      documentListMock.rowFilter = undefined;
+    });
+
+    afterEach(() => {
+      documentListMock.rowFilter = undefined;
```
AleksanderSklorz: Just to ensure - do we really need to reset it twice (in beforeEach above and here in afterEach)?

projects/aca-shared/rules/src/app.rules.ts
```diff
@@ -320,14 +326,51 @@ export const isWriteLocked = (context: RuleContext): boolean =>
...
+ * Covers:
+ * - `entry.isLocked` — any server-reported lock
+ * - `cm:lockType` property — present when locked via the Lock action
+ * - `cm:workingcopy` aspect — the working copy created on checkout
+ * - `cm:checkedOut` aspect — the original file while its working copy exists
+ *
+ * @param context Rule execution context
+ * @param node Optional node to evaluate; defaults to context.selection.first
+ */
+export function isLockedOrWorkingCopy(context: RuleContext, node?: NodeEntry): boolean {
```
AleksanderSklorz, line 342: Please add documentation for all new rules to docs/extending/rules-list.md file

projects/aca-content/viewer/src/lib/components/viewer/viewer.component.html
```diff
@@ -5,6 +5,7 @@
     }"
     [maxRetries]="settings.viewerMaxRetries"
     [nodeId]="nodeId"
+    [readOnly]="node?.aspectNames?.includes('cm:checkedOut')"
```
AleksanderSklorz, line 8: could you please add unit test for that change to check if readOnly is set properly?

AleksanderSklorz:
I have couple questions which are present on below video: 
1. If I have working copy and modify it twice then it throws some error. Is that expected? (should modifying even be possible or should it be blocked)?
2. Except modifying images there are other options for modifying like metadata sidebar (changing name etc.). I see some options are disabled like aspects but we still can for example edit name (it's enabled). 
3. If we modify for example image like in first point then we lost information that file is working copy (on files list). Is that expected?


https://github.com/user-attachments/assets/08c06089-bd99-40bd-a49f-fbb918d73f4b

AleksanderSklorz:
@nikita-web-ua I just reminded - do we need some checks for -1 version to use old approach when somebody has older backend?

projects/aca-content/src/lib/components/dl-custom-components/lock-icon/lock-icon.component.ts
nikita-web-ua: yes, removed

projects/aca-content/src/lib/components/dl-custom-components/name-column/name-column.component.spec.ts
nikita-web-ua: done

nikita-web-ua: done

projects/aca-content/src/lib/directives/document-list.directive.spec.ts
nikita-web-ua: nope, removed

projects/aca-shared/rules/src/app.rules.ts
nikita-web-ua: added, thanks

projects/aca-content/viewer/src/lib/components/viewer/viewer.component.html
nikita-web-ua: done

nikita-web-ua:
> I have couple questions which are present on below video:
> 
> 1. If I have working copy and modify it twice then it throws some error. Is that expected? (should modifying even be possible or should it be blocked)?
> 2. Except modifying images there are other options for modifying like metadata sidebar (changing name etc.). I see some options are disabled like aspects but we still can for example edit name (it's enabled).
> 3. If we modify for example image like in first point then we lost information that file is working copy (on files list). Is that expected?
> 
>  Screen.Recording.2026-09-02.at.10.22.23.mov

1. Yes, image editing is enabled on working copy and disabled for checked out original on purpose as per AC. Although error is not expected, I've added redirection from deleted working copy to the new version of a node when action performed in viewer.
2. Yes, it is similar to Share, on working copy properties should be editable. (disabled ones are coming from BE as such)
3. When image is cropped and saved new version is uploaded which automatically removes a lock similar as by uploading new version on the locked file, I confirmed with Arabella this to be expected.

nikita-web-ua:
> @nikita-web-ua I just reminded - do we need some checks for -1 version to use old approach when somebody has older backend?

I've checked new API supports previous one, e.g. - node locked with previous /lock endpoint can be unlocked with new checkout API. So there should not be breaking changes

AleksanderSklorz (approved):

PR #5389, opened 2026-09-10
[MNT-25883] Fix: don't persist sorting key until document list load succeeds

nikita-web-ua:
https://hyland.atlassian.net/browse/MNT-25883

**Problem**: When a user sorts by an unsortable/invalid property (e.g. a custom property not indexed for ordering), the sort key is persisted to localStorage immediately on click, before the API request even completes. If that request then fails, the invalid key stays saved - every subsequent reload/refresh keeps re-sending the same broken sort, permanently breaking the view until the user manually clears local storage.

**Fix**: Defer persisting server-mode sort changes until the resulting document list load actually succeeds (ready), and discard the pending value if it fails (error), so an invalid sort key is never written to storage.


projects/aca-content/src/lib/directives/document-list.directive.spec.ts
```diff
@@ -270,6 +272,84 @@ describe('DocumentListDirective', () => {
     });
   });
 
+  describe('onSortingChanged - persisting sorting', () => {
+    const sortingChangedEvent = (detail: SortingChangedEventDetail) => new CustomEvent('sorting-changed', { detail }) as CustomEvent;
```
AleksanderSklorz: Please fix copilot's comment because it's valid + here you have missing return type for your sortingChangedEvent function

AleksanderSklorz (approved):

MichalKinas (approved):

